PR #12, opened 2025-01-30
Creating layup init cmdline and config file

Little-Ryugu:
Fixed bug with cmd_line functions
Created a layup init cmd_line function which copies the example layup configuration files to a user-specified location
Created a config file and a config file parser data class layupConfigs.
Unit tests for new functions were added.


.pre-commit-config.yaml
mschwamb, line 1: These files shouldn't keep changing every time there's a PR. where are they getting updated?

.pre-commit-config.yaml
```diff
@@ -47,7 +47,7 @@ repos:
         args: ["--verbose"]
     # Analyze the code style and report code that doesn't adhere.
   - repo: https://github.com/psf/black
-    rev: 23.7.0
+    rev: 24.10.0
```
mschwamb, line 50: Same for this - is there something that's changing this in the VS code or something like that?

src/layup/utilities/fileAccessUtils.py
```diff
@@ -0,0 +1,58 @@
...
+        The name of the argument being checked. Used for error message.
+
+    Returns
+    ----------
+    arg_fn : string
+        The filepath/name of the file to be checked.
+
+    """
+
+    logger = logging.getLogger(__name__)
```
mschwamb, line 25: Currrent  consensus is that we don't want logging, so everywhere there's logging should be printing to stderror or stdout

src/layup/utilities/fileAccessUtils.py
```diff
@@ -0,0 +1,58 @@
...
+
+    if os.path.exists(arg_fn):
+        return arg_fn
+    else:
```
mschwamb, line 29: Current consensus is that we don't want logging, so everywhere there's logging should be printing to stderror or stdout

src/layup/utilities/layupConfigs.py
```diff
@@ -0,0 +1,284 @@
...
+        """
+        validates the auxililary config attributes after initialisation.
+        """
+        for file in self.default_filenames:
+            if file != "meta_kernel":
+                if (
+                    self.default_filenames[file] == getattr(self, file)
+                    and getattr(self, file + "_url") != self.default_url[file]
+                ):
+                    logging.error(f"ERROR: url for {file} given but filename for {file} not given")
```
mschwamb, line 128: Current consensus is that we don't want logging, so everywhere there's logging should be printing to stderror or stdout

mschwamb (changes requested):
Some requested changes based on telecon discussions

mschwamb:
@Little-Ryugu  One other thing for when you're back from Spain is we want the configuration file for now to be an optional config file variable. If someone does not provide it on the command line then we want to go with the defaults for the files. It should also be if someone provides an empty config file too.

src/layup_cmdline/init.py
```diff
@@ -0,0 +1,54 @@
...
+    copy_location = os.path.abspath(args.path)
+    which_configs = "Default"
+    return copy_demo_configs(copy_location, which_configs, args.force)
+
+
+def main():
+    parser = LayupArgumentParser(
+        prog="layup init",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+        description="Initialize configuration files for a new simulation.",
```
mschwamb, line 30: there's no simulation

src/layup_cmdline/main.py
```diff
@@ -36,6 +36,7 @@ def main():
         "   orbitfit       fit orbits\n"
         "   predict        predict orbits\n"
         "   bootstrap      Download datafiles required to run layup\n"
+        "   init           Initialize a new simulation.\n"
```
mschwamb, line 39: it's not a simulation - generate default configuration file

src/layup_cmdline/orbitfit.py
```diff
@@ -13,6 +13,15 @@ def main():
...
     )
     optional = parser.add_argument_group("Optional arguments")
+    optional.add_argument(
+        "-c",
+        "--config",
+        help="Input configuration file name",
+        type=str,
+        dest="c",
+        required=False,
+    )
```
mschwamb, line 23: this also needs to be in boostrap.py

mschwamb (changes requested):
some minor changes requested  @Little-Ryugu

PR #41, opened 2025-03-05
Add CSV and HDF5 File Readers

### LINKED ISSUE #14: Build functions to read in six orbit types and populate data classes for layup convert, predict, and visualize
Six orbit types to read in - the same as sorcha - for these three components we're accepting either CSV or HDF5 versions of these files. Format/column names will be the same as Sorcha (see the[ inputs doc page](https://sorcha.readthedocs.io/en/latest/inputs.html) for a description/column headings).

drewoldag:
Fixes #14 .

Describe your changes.
- Adding `sorcha` as a dependency for later access to `convert` code.
- Copied over the `sorcha` CSV and HDF5 file reader code and tests.
- The file readers were updated to now return [Numpy structured arrays](https://numpy.org/doc/2.2/user/basics.rec.html) rather than Pandas dataframes. This will serve as the underlying data class for future calls by layup convert, predict, and visualize.
- For the CSV reader, we were able to completely remove the use of Pandas. We also made some simplifications since we now enforce that the CSV header row must be the first row (which aligned with Sorcha's documentation).
- For the HDF5 file reader, pandas is still the simplest API for initial ingestion with the chunking mechanisms we wanted, and the result is then converted to a structured array. 
- Several test data files have been added, including several from the sorcha repository.


tests/data/testcolour.txt
mschwamb, line 1: We're not going to have physical properties like this for layup - can we find a way to on use this file for testing?

tests/data/testcomet.txt
mschwamb, line 1: Same thing for this file. We're not going to be inputting something in this format into layup. So can we remove it/replace with something else for the tests?

src/layup/utilities/file_readers/CSVReader.py
```diff
@@ -0,0 +1,260 @@
...
+            Default = csv
+
+        **kwargs: dictionary, optional
+            Extra arguments
+        """
+        super().__init__(**kwargs)
+        self.filename = filename
+
+        if sep not in ["whitespace", "csv", "comma"]:
+            pplogger = logging.getLogger(__name__)
```
mschwamb, line 37: is it worth getting rid of the pplogger variable name in layup? It came from pp= postprocessing as being what sorcha did, so maybe it's worth retiring that here for layup.

mschwamb (changes requested):
I added some comments to consider and some comments about a subset of the test files

src/layup/utilities/file_readers/CSVReader.py
```diff
@@ -0,0 +1,260 @@
+import numpy as np
+import logging
+import sys
+
+from layup.utilities.file_readers.ObjectDataReader import ObjectDataReader
+
+# Characters we remove from column names.
+INVALID_COL_CHARS = "!#$%&‘()*+, ./:;<=>?@[\\]^{|}~"
+
```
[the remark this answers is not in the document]
wilsonbb: I'm fine with deferring to @drewoldag or @maxwest-uw here, but in my opinion it's fine for a constant like this which is used in multiple functions. I do think changing it to `_INVALID_COL_CHARS` would be the right convention though to indicate that it shouldn't be used beyond this module.

tests/data/testcomet.txt
wilsonbb: Removed since this was only for the comet-related test of this format that I copied over from sorcha. So just deleted that test here rather than replacing with a new file.

tests/data/testcolour.txt
wilsonbb: I created .txt file version of the existing CART.csv, and I'm now use that in our tests for comparing how the reader handles whitespace separated text files against normal CSV files.

src/layup/utilities/file_readers/CSVReader.py
wilsonbb: Ah that makes sense why it was the convention. Renamed to simply `logger` since it only has local scope but open to renaming if desired.

src/layup/utilities/file_readers/CSVReader.py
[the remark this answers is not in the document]
maxwest-uw: agree with wilson here

maxwest-uw (approved):
this looks good to me! I still feel like this would all be better as part of a shared `sorcha`<->`layup`ephemeris utility package but this all is fine as well

tests/data/testcolour.txt
drewoldag: Removed `testcolour.txt`.

src/layup/utilities/file_readers/CSVReader.py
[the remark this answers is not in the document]
drewoldag: Also, agreed, for this use case, having `_INVALID_COL_CHARS` as a global is fine.

mschwamb (approved):

PR #60, opened 2025-03-11
Initial Scaffolding of `convert`

### LINKED ISSUE #53: write main executing function for layup convert that hooks all the pieces together
- calls the functions to input the files
- calls the convert utilities 
- outputs the results by calling the output functions

From Pedro: This [notebook in Sorcha](https://sorcha.readthedocs.io/en/latest/notebooks/demo_CoordinateTransformations.html) does all the conversions. There's also the universal_keplerian methods that do the inverse transformation (from keplerian to cartesian elements). The code is 100% in numba, so processing a large number of inputs can be done as a numba loop and parallelized rather simply (I've done so in my own tests) for even faster runtimes (for impatient people like me :) )

drewoldag:
Fixes #53.

Initial scaffolding for `convert` verb.

Includes an implementation of `convert_cli` meant to be called by layup's CLI interface as well as a general parallelized `convert` method which can be called directly with numpy structured arrays.

The underlying convert method currently just returns the input data back unmodified, and a subsequent PR will add the logic for converting between the various formats. 


src/layup/convert.py
```diff
@@ -0,0 +1,92 @@
...
+from layup.file_io.CSVReader import CSVDataReader
+from layup.file_io.HDF5Reader import HDF5DataReader
+from layup.file_io.file_output import write_csv, write_hdf5
+
+
+def process_chunk(data):
+    pass
+
+
+def convert(
```
mschwamb, line 17: so I think we should this run_convert or something like that. I expect that orbit fitting will need converting to the ecliptic plane reference plane and also allowing convert to be the name of the function that does the orbit converting might be cleaner for those that need to use that if they have mixed types and directly using th python API

src/layup/utilities/file_io/ObjectDataReader.py
```diff
@@ -187,27 +198,43 @@ def _process_and_validate_input_table(self, input_table, **kwargs):
...
+                outstr = f"ERROR: While reading table {self.filename}. Multiple formats found."
+                logger.error(outstr)
+                sys.exit(outstr)
+
+            # Check that the expected columns are present in this chunk
+            for col in REQUIRED_COLUMN_NAMES[self._format_column_name]:
+                if col not in input_table.dtype.names:
+                    outstr = f"ERROR: While reading table {self.filename}. Required column {col} not found."
+                    logger.error(outstr)
+                    sys.exit(outstr)
```
mschwamb, line 231: I think REQUIRED_COLUMN_NAMES might need to be something sent in because for layup orbitfit the columns are going to be different for the CSV file/HDF file so does it make sense that required_column_names be set in the function that calls the reader fclass?

src/layup/convert.py
```diff
@@ -0,0 +1,92 @@
...
+    input_file = Path(input)
+    output_file = Path(f"{output_file_stem}.{file_format.lower()}")
+    output_directory = output_file.parent.resolve()
+
+    if num_workers < 0:
+        num_workers = os.cpu_count()
+
+    # Check that input file exists
+    if not input_file.exists():
+        raise FileNotFoundError(f"Input file {input_file} does not exist")
```
mschwamb, line 34: we have raise errors here but elsewhere we have logging that could go to standard error or a log file. So we probably want to be consistent

mschwamb: It might be okay to raise errors as this is for command line, but the bits that might be run in an more automated way like using the orbit fitting functions might be okay with the logging that could go stderror or logfile.

src/layup/convert.py
drewoldag: Would it make sense to call this `convert_orbit_format` or something similar? If `run_convert` makes sense, we can use that, but the name didn't seem immediately intuitive about how it would distinguish between the different converting cases.

mschwamb:
I'm sitting with @bernardinelli  - we've both looked at this - and we think this looks good and what will work for Rubin Solar System Processing and for using the CLI

src/layup/utilities/file_io/ObjectDataReader.py
drewoldag: Created https://github.com/Smithsonian/layup/issues/71 to track this work

src/layup/convert.py
drewoldag: I believe that this should be resolved as of the last commit from @wilsonbb

src/layup/convert.py
```diff
@@ -0,0 +1,92 @@
...
+    chunks = [(i, min(i + chunk_size, total_rows)) for i in range(0, total_rows, chunk_size)]
+
+    for chunk_start, chunk_end in chunks:
+        # Define the blocks of each chunk to send to the workersc
+        block_size = max(1, math.ceil(int(chunk_size / num_workers)))
+        blocks = [(i, min(i + block_size, chunk_end)) for i in range(chunk_start, chunk_end, block_size)]
+
+        with ProcessPoolExecutor(max_workers=num_workers) as executor:
+            futures = [
+                executor.submit(process_chunk, reader.read_rows(block_start=start, block_size=end - start))
```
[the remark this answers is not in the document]
drewoldag: This has been converted to it's own function. Currently it is kept in `convert.py`, but we can now easily move it to it's own module. The work to do so is tracked here: https://github.com/Smithsonian/layup/issues/70

src/layup/convert.py
```diff
@@ -0,0 +1,191 @@
...
+    block_size = max(1, int(len(data) / n_workers))
+    # Create a list of tuples of the form (start, end) where start is the starting index of the block
+    # and end is the last index of the block + 1.
+    blocks = [(i, min(i + block_size, len(data))) for i in range(0, len(data), block_size)]
+
+    with ProcessPoolExecutor(max_workers=n_workers) as executor:
+        # Create a future applying the function to each block of data
+        futures = [executor.submit(func, data[start:end], **kwargs) for start, end in blocks]
+        # Concatenate all processed blocks together as our final result
+        return np.concatenate([future.result() for future in futures])
```
mschwamb, line 48: since we're waiting for everything to finish anyway would it be worth using map() instead so by defaults things are in the same order as input? it's not a requirement that the output arrays be in the same order, but it is helpful as this is what most people would be expecting in terms of their output file?

I'm completely new to this kind of parallelization - I'm reading [this](https://superfastpython.com/processpoolexecutor-in-python/#Step_1_Create_the_Process_Pool) and it states

```
Firstly, map() returns an iterator over objects, not over Futures.

Secondly, map() returns results in the order that the tasks were submitted, not in the order that they are completed.
```

wilsonbb: Good question! The above does concatenate `future.result()` in the order the futures were submitted, which is the same order as the blocks we create. So the order of the input is maintained.

The tests in test_convert currently will fail if the order fails. (There is a separate failure at the moment in `test_file_output` which is unrelated btw)

wilsonbb (approved):

mschwamb (approved):

Issue #70, opened 2025-03-12
Isolate parallelization code


drewoldag:
Currently there is a parallelization code in `convert.py`, we should move this parallelization code out into it's own module. 

Likely, there will be a function that takes data and func to parallelize.
Perhaps something like this:
```
def parallelize(data, func, ...):
        block_size = max(1, math.ceil(int(chunk_size / num_workers)))
        blocks = [(i, min(i + block_size, chunk_end)) for i in range(chunk_start, chunk_end, block_size)]

        with ProcessPoolExecutor(max_workers=num_workers) as executor:
            futures = [
                executor.submit(process_chunk, reader.read_rows(block_start=start, block_size=end - start))
                for start, end in blocks
            ]
```

Another function in the same parallelize module that will accept a reader and send out chunks of data to parallelize function might be useful as well.

Issue #47, opened 2025-03-10
column checks/checks on input formats for layup convert/visualize/predict


mschwamb:
we need column checks/checks on input formats for layup convert/visualize/predict  & unit tests

mschwamb:
We might want to remove any additional columns that come with the inputs. For visualize I think we will want to allow an optional column for the color and line style for matplotlib

mschwamb:
We also need to check the file is all in the same format

PR #73, opened 2025-03-12
Implement `convert` logic.

### LINKED ISSUE #16: make layup convert
This utility reads in orbit file and converts to orbits to a user specified type and outputs to a file . We'll expand the capabilities after the March 10 kick off meeting, but the main goal is have:

The utility input and convert between the SAME 6 types of orbit definitions from [Sorcha](https://sorcha.readthedocs.io/en/latest/inputs.html) (note that read in is a separate ticket #14 ) 

From Pedro: This  [notebook in Sorcha](https://sorcha.readthedocs.io/en/latest/notebooks/demo_CoordinateTransformations.html) does all the conversions. There's also the universal_keplerian methods that do the inverse transformation (from keplerian to cartesian elements). The code is 100% in numba, so processing a large number of inputs can be done as a numba loop and parallelized rather simply (I've done so in my own tests) for even faster runtimes (for impatient people like me :) )

The question is how to package this - whether it's importing the relevant classes from Sorcha or including them separately in here. 

Output file formats are HDF5 and csv

drewoldag:
Fixes #16 and #47

Implementing convert logic for conversion between the different orbit definitions. 

Note that we currently only test BCOM and KEP as input formats, but successfully convert to all 6 expected output formats.

Some more minor changes in service of this:
* Fixing some small bugs/typos in the command line argument parser for convert
* updating the default number of CPU workers to 1 for the commandline interface since sharding out to multiple workers and combining the results has too high of an overhead cost for small input files. Previously we tried to use all CPUs available. We also only use a single worker in our unit tests to ensure that this is fast. (we can provide more tests for scaling here as part of  #70
* Moves specifying which columns are required from `ObjectDataReader` to `convert.py` as part of #47
* To make use of sorcha's orbit conversion, we made use of https://github.com/dirac-institute/sorcha/pull/1138

- [X] Have you written a unit test for any new functions?
- [X] Does Layup run successfully on a test set of input files/databases?
- [X] Have you used black on the files you have updated to confirm python programming style guide enforcement?


mschwamb:
Back seat driving here so ignore me if this is not useful/is annoying, the pyproject.toml needs to be updated to pull sorcha from the github version for now I think as it is CI is pulling from the conda-forge version which we haven't pushed the latest sorcha changes to

mschwamb (approved):

Issue #76, opened 2025-03-13
calculate 1/ao for comets


mschwamb:
This would likely be a new utility layup comet or something with that verbage. 

@carrholt  has provided slides on her wishlist for this https://docs.google.com/presentation/d/1k-xUlniZMsfslLGDsiXkQ4kNvo7EahIw9GQW6UFTvoY/edit?usp=sharing

From @matthewholman  Regarding integrations for 1/a_0 calculations, DE441 covers -13,200 years to +17,191 years (https://ssd.jpl.nasa.gov/doc/de440_de441.html).   That should be long enough, i.e. we could do the integrations with ASSIST and fits with layup without a problem.  Longer integrations will need a transition between ASSIST and REBOUND.

@hannorein would you be able to spend a small about of scoping out how challenging/time insensitive it would be for taking an assist run that had to stop at -13,200 years and starting a full rebound simulation? It may be completely out of scope but to go beyond ASSIST integrations into pure REBOUND, but it would be worth knowing how much effort would be needed/a possible pathway forward even for a future version 2.0

hannorein:
Switching from ASSIST to pure REBOUND would be very easy. A few lines of code. It's not as accurate of course because we we miss any additional forces but those don't matter much far out (If the comets are really really far away, one might need to include forces from galactic tides. I'm not sure if people typically do that when they calculate 1/ao). Performance wise, this depends on what the setup looks like. If these comets don't come close to the inner planets, then their motion can be neglected and it should be quite fast. If they do have close encounters, then all bets are off and it can very slow.

mschwamb:
Cool - if someone really wants this I imagine they'll live with the slowness. 

@hannorein Would you be interested in the next month working on the first part for layup comet (I don't know if layup ao would be a better name) integrating out to what ASSIST can do  with DE441 and having it set up in a way for adding in the switch to PURE REBOUND later?

hannorein:
In summary: give me some input data and I'll code up something that returns 1/a0 at r_h=250AU!

carrholt:
Here is an orbit file for long-period comets [COM_LPCs.csv](https://github.com/user-attachments/files/19234223/COM_LPCs.csv)

hannorein:
I started looking into this. See https://github.com/Smithsonian/layup/blob/cef1ec396b69ea58aa0f798640880fba0d039899/examples/Comets.ipynb

Questions:
[...]

PR #80, opened 2025-03-13
WIP - Comet example

hannorein:
Starting point for #76. Do not merge yet.


mschwamb:
I think we can lose this because there's a fork of the repo we can look at for the notebook.

PR #99, opened 2025-03-19
Move parallelization logic in convert into its own module

### LINKED ISSUE #70: Isolate parallelization code

wilsonbb:
Fixes #70 .

Moves the parallelization code called by layup convert into its own utilities file for shared functions for processing numpy structured arrays. This allows the code to be re-used by other layup verbs like orbitfit.

### DIFF
--- a/src/layup/convert.py
+++ b/src/layup/convert.py
@@ -5,13 +5,12 @@ from typing import Literal
 
 import numpy as np
 
-from concurrent.futures import ProcessPoolExecutor
-
 from sorcha.ephemeris.simulation_geometry import equatorial_to_ecliptic
 from sorcha.ephemeris.simulation_setup import _create_assist_ephemeris
 from sorcha.ephemeris.simulation_parsing import parse_orbit_row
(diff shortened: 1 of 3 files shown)


drewoldag (approved):
This looks good to me. Thanks for extracting this parallelization code!

PR #102, opened 2025-03-20
Main logic for `orbitfit` verb

### LINKED ISSUE #46: read in CSV/HDF5 inputs to layup orbitfit
Read in the same contents as the ADES format files (see #44), but instead of PSV for XML the file will be in CSV or HDF5

### LINKED ISSUE #106: Converting UTC to TDB
By convention, the times of astrometric observations (obs80 or ADES) are in UTC.  These need to be converted to TDB for the C++ code.  Examples of this conversion, using spiceypy routines, can be found in process_ob80.py.

drewoldag:
Fixes #46 .
Fixes #106

The core logic for the `orbitfit` verb. This is heavily influenced by the `convert` verb. 

However a different approach to parallelism was taken. In `convert`, a chunk at a time could be distributed across any number of processors. Here we need to take a chunk of object ids each with multiple observations and distribute them across processors such that each processor receives all the observations for any given object. i.e. the observations for an object cannot be spread across multiple processors. Each worker receives all of the observations for a single object, calls the pybind method for the C++ orbit_fit code and returns a structured array with the output. These structured arrays are then concatenated for our final result.

`orbit_fit.cpp` is given a simple stub method returning a fake result, and we provide a simple test that confirms that the output CSV file has the same number of rows as the number of unique provIDs in the input file. Additional testing will be added once actual orbit fitting is performed.


src/layup/orbitfit.py
```diff
@@ -0,0 +1,140 @@
...
+    # Check that the file format is valid
+    if output_file_format.lower() not in ["csv", "hdf5"]:
+        logger.error("File format must be 'csv' or 'hdf5'")
+
+    reader_class = INPUT_FORMAT_READERS[input_file_format]
+    if reader_class is None:
+        logger.error(f"File format {input_file_format} is not supported")
+
+    #! Don't know if this is the correct way to instantiate the reader
+    reader = reader_class(input_file, format_column_name="FORMAT")
```
wilsonbb, line 121: ADES data will not have a FORMAT column. Ex:

https://drive.google.com/file/d/17DA4pKCJCce5NgE3rGFS6upbxWF10cT1/view?pli=1

The interesting thing about ADES is that here the ObjID will be in the provid column. However the provid column is optional while the permid column is required. So a given object seems to all have the same permid but some rows may additionally have the provid.

drewoldag: I've removed the reference to `format_column_name` here, and the code seems to run as expected. But to your point, we need to figure out the appropriate way to hand permid and provid to create an ObjID column. 

And more generally, I think it would be good to modify our reader classes such that we can programmatically define the string "ObjID", so that if a user prefers "obj_id", they would be able to do specify that.

src/layup/orbitfit.py
```diff
@@ -0,0 +1,254 @@
...
+        The directory where the required orbital files are stored
+    """
+    # Define a structured dtype to match the OrbfitResult fields
+    result_dtype = np.dtype(
+        [
+            ("provID", "O"),  # Object ID
+            ("csq", "f8"),  # Chi-square value
+            ("ndof", "i4"),  # Number of degrees of freedom
+            ("state", "O"),  # State vector (6 elements)
+            ("epoch", "f8"),  # Epoch
```
mschwamb, line 47: can we make sure epoch has a unit attached with it so MJD_TBD so that we don't in later years get confused about what time we're in?

src/layup/orbitfit.py
```diff
@@ -0,0 +1,254 @@
...
+        The directory where the required orbital files are stored
+    num_workers : int
+        The number of workers to use for parallel processing. Default is 1
+    primary_id_column_name : str
+        The name of the primary identifier column for the objects. Default is "provID".
+    """
+
+    layup_observatory = LayupObservatory()
+
+    et_col = np.array([spice.str2et(row["obstime"]) for row in data], dtype="<f8")
```
mschwamb, line 101: My read of the documentation this morning is str2et gives time from J2000, might get to get JD we need to add the seconds since j2000, so there needs to be an additional step of 

jd_tdb = spice.j2000() + et_col / (24 * 60 * 60)

Is this correct @matthewholman

src/layup/orbitfit.py
mschwamb: Other verbs expect epochMJD_TDB so it would be good if the output column from orbitfit is named this.

src/lib/orbit_fit/orbit_fit.cpp
```diff
@@ -1202,7 +1202,24 @@ int main(int argc, char *argv[]) {
 
 #ifdef Py_PYTHON_H
 static void orbit_fit_bindings(py::module& m) {
-	m.def("orbit_fit", &orbit_fit::orbit_fit, R"pbdoc(Main function)pbdoc"); 
+	m.def("orbit_fit", [](const std::vector<Observation>& observations) -> OrbfitResult {
```
maxwest-uw, line 1205: the actual function that we'll be calling to access `orbit_fit` won't be `orbit_fit`, but instead what is currently called `run_from_files`(will probably be renamed to just `run_orbit_fit`) (confusing, sorry...)

maxwest-uw: the input will be the str to the cache dir and a list of `Observations` for a given object

maxwest-uw: sorry the recommendation here is probably that we'll not have `orbit_fit` be bound at all and instead just have the run function and `gauss`

maxwest-uw (commented):
all the pybind stuff looks good to me !

src/layup/orbitfit.py
wilsonbb: Per discussion with Max, wait on the renaming to avoid merge conflicts with their PR

src/layup/orbitfit.py
wilsonbb: 1. Confirmed with @matthewholman that this is the correct calculation. and renamed to jd_tdb for clarity.

wilsonbb: 2. Right now we're just using the column names on the example files, and aren't doing any conversion/sanitation. We probably want to lock in what that should look like (for example the current code will fail if the user gives a column of "Obstime" rather than "obstime"). But would like to isolate changes there to its own PR if we can

src/lib/orbit_fit/orbit_fit.cpp
wilsonbb: Ack, moved my fake orbit fit code to `run_from_files` so we can still run the test. So a small merge conflict with your PR but should be easy to fix.

maxwest-uw: yeah I can just rebase!

maxwest-uw (approved):

Issue #55, opened 2025-03-10
Convert Rust code for querying JPL for special case observatories


trruch:

matthewholman:
## Scope / design note

**TL;DR — Python-only, ingestion-side; no C++ changes.** For the handful of MPC codes that are *spacecraft* (HST, JWST, etc.), layup can't compute a position from parallax constants and currently raises `ValueError` unless the user supplies ADES `pos1/2/3`. #55 adds a fallback that queries **JPL Horizons** for the observer state and feeds it into the existing moving-observer path (the machinery #147 already built). Reference implementation is the Rust `from_horizons()` in `spacerocks` (`~/Dropbox/roman-footprint/spacerocks`).

### Decisions (confirmed)
1. **Use the JPL Horizons HTTP API**, not local SPK kernels — space-based observations are a rare case, so a live query is acceptable; avoids managing per-spacecraft kernels. (SPK-backed resolution can be a future enhancement if reproducibility/offline becomes a concern.)
2. **Query geocentric** (`center=@earth`) so the result drops straight into the existing "geocentric observer + add Earth's barycentric state" path from #147 — no new downstream code.
3. **Caching**: in-run batched cache only for v1.0 (one request per spacecraft, not per epoch); cross-run disk caching deferred to a follow-up.

### The mapping (MPC obscode → JPL NAIF id), ported from `spacerocks`
| MPC | NAIF | Observatory | | MPC | NAIF | Observatory |
|-----|------|-------------|-|-----|------|-------------|
| `250` | -48 | Hubble | | `C51` | -163 | WISE |
| `274` | -170 | JWST | | `C58` | -33 | NEO Surveyor |
| `245` | -79 | Spitzer | | `C49` | -234 | STEREO-A |
| `C55` | -227 | Kepler | | `C50` | -235 | STEREO-B |
| `C54` | -98 | New Horizons | | | | |

Roving observers (`247`/`270`) and occultation (`275`) are **out of scope** — they remain user-supplied via ADES position columns.

### Horizons request
`GET https://ssd.jpl.nasa.gov/api/horizons.api` with:
- `COMMAND='<naif id>'`, `EPHEM_TYPE='VECTORS'`, `VEC_CORR='NONE'` (geometric position at the epoch — layup does its own light-time in the integrator)
- `CENTER='@earth'`, `REF_SYSTEM='J2000'`, `REF_PLANE='FRAME'` (ICRF equatorial)
[...]

PR #104, opened 2025-03-26
Add LayupObservatory class for computing observatory positions/velocities

### LINKED ISSUE #49: compute observatory positions from input obscode
this is in the [observatory class](https://github.com/dirac-institute/sorcha/blob/3cbe74435a12f2e05c9f08da2579e5417f77317e/src/sorcha/ephemeris/simulation_parsing.py#L278) in Sorcha does this - Matt has the same code in [this example ](https://drive.google.com/file/d/14Cu11miKqRswKHuriIsNdkNAwEN6u0Qt/view?usp=sharing)

wilsonbb:
Fixes #49

Introduces a subclass of [Sorcha's Observatory](https://github.com/dirac-institute/sorcha/blob/575a72d198f5fe9570f17236aa8c1fb2b5ce60a6/src/sorcha/ephemeris/simulation_parsing.py#L278) class, `LayupObservatory` which handles layup specific configurations and provides a convenience method for obtaining barycentric observatory positions and velocities from [MPC Observatory codes](https://en.wikipedia.org/wiki/List_of_observatory_codes) when operating on our layup structured array representation of the input data. Note that the `LayupObservatory` assumes its input structured arrays has a column 'et' which has the "ephemeris time" of the observation in TDB.

Note that in the included test data, several space telescopes are included which do have a known geocentric position within our Observatory ephemeris. Currently for this case we return NaNs for the position and velocity vectors as the default behavior. We instead fail if the caller uses `fail_on_missing=True`. This case should eventually be handled by issue #55
 


drewoldag (approved):
This looks great, approved!

PR #119, opened 2025-04-17
Predict verb cmd_line args draft

### LINKED ISSUE #89: define layup predict CLI command line arguments
define layup predict CLI command line arguments

Little-Ryugu:
Fixes #89  .

Added command-line arguments for the predict verb.
Fixed bug in find_directory_or_exit and its unit tests.

### DIFF
--- a/src/layup_cmdline/predict.py
+++ b/src/layup_cmdline/predict.py
@@ -2,6 +2,8 @@
 # The `layup predict` subcommand implementation
 #
 import argparse
+import os
+from pathlib import Path
 from layup_cmdline.layupargumentparser import LayupArgumentParser
 from astropy.time import Time
 import astropy.units as u
@@ -28,6 +30,8 @@ UNIT_DICT = {
     "days": u.day,
     "d": u.day,
 }  # used to convert unit_str to an astropy.unit
+
+SEC_PER_DAY = 24 * 60 * 60
 
 
 def main():
@@ -143,6 +147,16 @@ def main():
     )
 
     optional.add_argument(
+        "-st",
+        "--station",
+        type=str,
+        dest="station",
+        default="X05",  # Rubin observatory
+        help="Station code for the observatory. Default is X05 (Rubin observatory). See https://www.minorplanetcenter.net/iau/lists/ObsCodes.html",
+        required=False,
+    )
+
+    optional.add_argument(
         "-t",
         "--timestep",
         help="Timestep for predict. Must be string consisting of float followed by the unit (d=day, h=hour, m=minute, s=second) e.g. 1.3h or 30m",
@@ -157,7 +171,7 @@ def main():
     return execute(args)
 
 
-def convert_input_to_JD_TDB(input_str: str) -> float:
+def convert_input_to_JD_TDB(input_str: str, cache_path: Path) -> float:
     """
     Convert a string to a JD_TDB date. The string can be in the format of a
     Julian date TDB for or a date string in the format YYYY-mm-dd.
@@ -166,6 +180,9 @@ def convert_input_to_JD_TDB(input_str: str) -> float:
     ----------
     input_str : str
         The input string to convert.
+    cache_path : Path, optional
+        The path to the cache directory. If not provided, the default cache
+        directory will be used.
 
     Raises
     ------
@@ -177,21 +194,35 @@ def convert_input_to_JD_TDB(input_str: str) -> float:
     float
         The converted JD_TDB date.
     """
+    import spiceypy as spice
+
+    spice_kernel_loaded = False
+    if os.path.exists(cache_path / "naif0012.tls"):
+        # If the file exists, load it
+        spice.furnsh(str(cache_path / "naif0012.tls"))
+        spice_kernel_loaded = True
+
     try:
         # Assume that input is a JD_TDB float that was converted to string. Attempt
         # to convert it back to a float.
         date_JD_TDB = float(input_str)
     except ValueError:
-        # If conversion to float fails, assume that the input was a date string
-        # with a format like YYYY-mm-dd. Convert that to a datetime and then to
+        # If conversion to float fails, assume that the input was a date string.
+        # If the spice kernel was loaded, we'll attempt to parse the date using
+        # spiceypy. Otherwise, we'll use assume that the input is a date string
+        # formatted as YYYY-mm-dd. We convert that to a datetime and then to
         # a float representation of a JD_TDB date.
-        try:
-            date_JD_TDB = Time(
-                datetime.strptime(input_str, "%Y-%m-%d"), format="datetime", scale="tdb"
-            ).tdb.jd
-        except ValueError:
-            # If the date string is not in the expected format, raise an error.
-            raise ValueError(f"Invalid date string format: {input_str}. Expected format is YYYY-mm-dd.")
+        if spice_kernel_loaded:
+            et = spice.str2et(input_str)
+            date_JD_TDB = spice.j2000() + et / SEC_PER_DAY
+        else:
+            try:
+                date_JD_TDB = Time(
+                    datetime.strptime(input_str, "%Y-%m-%d"), format="datetime", scale="tdb"
+                ).tdb.jd
+            except ValueError:
+                # If the date string is not in the expected format, raise an error.
+                raise ValueError(f"Invalid date string format: {input_str}. Expected format is YYYY-mm-dd.")
 
     return date_JD_TDB
 
@@ -202,16 +233,13 @@ def execute(args):
     from layup.utilities.cli_utilities import warn_or_remove_file
     from layup.utilities.file_access_utils import find_file_or_exit, find_directory_or_exit
     import sys
-
-    start_date = convert_input_to_JD_TDB(args.s)
-
-    end_date = convert_input_to_JD_TDB(args.e) if args.e else start_date + args.days
+    import pooch
 
     # check input exists
     find_file_or_exit(args.input, "input")
 
     # Check that output directory exists
-    find_directory_or_exit(args.o, "-o, --")
+    find_directory_or_exit(args.o, "-o, --output")
 
     # check for overwriting output file
     warn_or_remove_file(str(args.o), args.force, logger)
@@ -219,6 +247,13 @@ def execute(args):
     # check ar directory exists if specified
     if args.ar_data_file_path:
         find_directory_or_exit(args.ar_data_file_path, "-ar, --ar_data_path")
+        cache_dir = Path(args.ar_data_file_path)
+    else:
+        cache_dir = Path(pooch.os_cache("layup"))
+
+    start_date = convert_input_to_JD_TDB(args.s, cache_dir)
+
+    end_date = convert_input_to_JD_TDB(args.e, cache_dir) if args.e else start_date + args.days
 
     # check format of input file
     if args.i.lower() == "csv":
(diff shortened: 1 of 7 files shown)


src/layup_cmdline/predict.py
```diff
@@ -26,11 +137,105 @@ def main():
...
+
+    # Check that output directory exists
+    find_directory_or_exit(args.o, "-o, --")
+
+    # check for overwriting output file
+    warn_or_remove_file(str(args.o), args.force, logger)
+
+    # check that days is greater than zero if being used
+
+    if (args.e == None) and args.days <= 0:
```
mschwamb, line 202: Because of the additions, this could be changed to checking that end_time is > start_time and the print statement to just an error message.

src/layup_cmdline/predict.py
```diff
@@ -26,11 +137,105 @@ def main():
...
+        # Assume that input is a jd_TDB float that was converted to string. Attempt
+        # to convert it back to a float.
+        start_time = float(input_str)
+        start_time = spice.j2000() + start_time / (24 * 60 * 60)
+    except ValueError:
+        # If conversion to float fails, assume that the input was a date string
+        # with a format like YYYY-mm-dd. Convert that to a datetime and then to
+        # a float representation of a jd_TDB date.
+        try:
+            start_time = Time(datetime.strptime(input_str, "%Y-%m-%d"), format="datetime", scale="utc").tdb.jd
```
mschwamb, line 172: change scale to "tdb" and then this will work as intended so that 

 layup predict  main.py -s 2000-01-01
 gives  
 
 Start time: 2451544.5
End time: 2451558.5

mschwamb:
@drewoldag  I didn't want to make changes without seeing where you're at - I did some testing and spotting a few places where minor changes are needed to get the correct value for the start time and end time. 

One additional thing that would be good to do is ensure the start_time and end_time variable is named with a JD_TDB so it's clear what format it is in.  

Once these small changes are in this PR is ready to go (yay!)

src/layup_cmdline/predict.py
```diff
@@ -26,11 +137,105 @@ def main():
...
+
+    # check input exists
+    find_file_or_exit(args.input, "input")
+
+    # Check that output directory exists
+    find_directory_or_exit(args.o, "-o, --")
+
+    # check for overwriting output file
+    warn_or_remove_file(str(args.o), args.force, logger)
+
```
mschwamb, line 199: 197 and 198 with 

    ```
# check ar directory exists if specified
    if args.ar_data_file_path:
        find_directory_or_exit(args.ar_data_file_path, "-ar, --ar_data_path")

    # check format of input file
    if args.i.lower() == "csv":
        output_file = args.o + ".csv"
    elif args.i.lower() == "hdf5":
        output_file = args.o + ".h5"
    else:
        sys.exit("ERROR: File format must be 'csv' or 'hdf5'")

    # check for overwriting output file
    warn_or_remove_file(str(output_file), args.force, logger)
```
Ryan and I forgot to check some things and put the output file in the right format 


    ```
# check ar directory exists if specified
    if args.ar_data_file_path:
        find_directory_or_exit(args.ar_data_file_path, "-ar, --ar_data_path")

    # check input exists
    find_file_or_exit(args.input, "input")

    # Check that output directory exists
    find_directory_or_exit(args.o, "-o, --")
    # check format of input file
    if args.i.lower() == "csv":
        output_file = args.o + ".csv"
    elif args.i.lower() == "hdf5":
        output_file = args.o + ".h5"
    else:
        sys.exit("ERROR: File format must be 'csv' or 'hdf5'")

    # check for overwriting output file
    warn_or_remove_file(str(output_file), args.force, logger)
```

src/layup_cmdline/predict.py
drewoldag: Done, swapped this out for the comparison you mentioned.

src/layup_cmdline/predict.py
drewoldag: Added this logic.

src/layup_cmdline/predict.py
```diff
@@ -11,13 +17,118 @@ def main():
...
+    )
+
+    optional.add_argument(
+        "-s",
+        "--start-date",
+        help="start date as jd_TDB or date string YYYY-mm-dd that will be converted to jd_TDB. If not specified default wil be current date",
+        dest="s",
+        type=str,
+        default=str(
+            Time(datetime.now(timezone.utc), format="datetime", scale="utc").tdb.jd
```
drewoldag, line 120: @mschwamb Should the scale here also be "tdb"?

mschwamb: because timezone is utc it's okay to be scale utc because then it's converted to tdb - but we could just go straight to tdb

mschwamb: but then I think the timezone has to change as well if scale = tdb

drewoldag:
I believe that this is ready for review. I've created a separate issue for adding unit tests for the input value parsing (specifically the conversion to JD_TDB) But it would be good to cover the timestep_str conversion that happens as well. https://github.com/Smithsonian/layup/issues/142

I'll start in on the python verb implementation tomorrow as well.

mschwamb:
Since I had some hand in this PR probably best for someone else to review. Maybe @matthewholman  or @bernardinelli

matthewholman:
This looks good, but I have a couple of suggestions.


First, I suggest changing convert_input_to_JD_TDB as below.  I *believe* that spice.str2et gives the user a lot more flexibility in specifying the input time without forcing us to handle a huge number of cases.   datetime.strptime() may also be flexible, but it looks like we have to specify a format.

To get str2et to work, we only need to load the naif0012.tls kernel.  That's in the cache_dir, so the location would need to be changed from what it is below.


import spiceypy as spice

# Load leap second kernel
spice.furnsh('naif0012.tls')

def convert_input_to_JD_TDB(input_str: str) -> float:
    """
    Convert a string to a JD_TDB date. The string can be in the format of a
    Julian date TDB for or a date string in the format YYYY-mm-dd.

    Parameters
    ----------
    input_str : str
        The input string to convert.

    Raises
    ------
    ValueError
        If the input string is not in the expected format, YYYY-mm-dd

    Returns
    -------
    float
        The converted JD_TDB date.
    """

    try:
        # Assume that input is a JD_TDB float that was converted to string. Attempt
        # to convert it back to a float.
        date_JD_TDB = float(input_str)
    except ValueError:
        # If conversion to float fails, assume that the input was a date string
        # with a format like YYYY-mm-dd. Convert that to a datetime and then to
        # a float representation of a JD_TDB date.
        try:
            #date_JD_TDB = Time(
            #    datetime.strptime(input_str, "%Y-%m-%d"), format="datetime", scale="tdb"
            #).tdb.jd
            et = spice.str2et(input_str)
            date_JD_TDB = spice.j2000() + et/(24*60*60)
        except ValueError:
            # If the date string is not in the expected format, raise an error.
            raise ValueError(f"Invalid date string format: {input_str}. Expected format is YYYY-mm-dd.")

    return date_JD_TDB

matthewholman:
The second comment is that predict will need to supply the observatory location at the times we are predicting.  @wilsonbb has put together those routines for orbitfit verb.

mschwamb:
So for the command line interface, another optional argument is needed then a -s --station for the user obscode at least for now - later functionality can be built for the user to provide observatory location - the variable can default to Rubin:  X05

drewoldag:
Notes:

I added `--station` and am using spiceypy.str2et as a first attempt to convert strings to JD_TDB with a fall back to the datetime(…) approach if the kernels are not found. Though, if the kernels aren’t found there will be serious problems down the line...

I did notice a slight difference in the float representation of the JD_TDB between the two different approaches, and want to make sure that’s expected/understood.

When I run the command: layup predict main.py -s 2000-01-01 the output is:
- 2451544.5007428695 - from spiceypy
- 2451544.50 - from Time(datetime(…))

This difference is about 1 minute. Is that ok, for some definition of “ok”?

mschwamb:
It's a problem - it's because https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/FORTRAN/spicelib/str2et.html assumes the time the datestring is UT - we're staying it should be a TDB date string or JD.

mschwamb:
We could say the datestring is UT and the JD is in TDB but is that confusing?

matthewholman:
Copying this from slack:
The difference between the times is that datetime is treating the input date as TDB, while str2et defaults to UTC.  If you try
layup predict main.py -s "2000-01-01 TDB"
you should get the same value.

drewoldag:
@mschwamb, based on your first response to my statement, "we're saying it should be a TDB date string or JD", I can code to that. The resulting JD_TDB times match when I pass the date string as `2000-01-01TDB`* or `"2000-01-01 TDB"`** , as @matthewholman suggested. 

Given all of that combined with:
- As near as I can tell, I can't make python's datetime.strptime properly parse a date string like `2000-01-01 TDB` without going through preprocessing first to strip off the "TDB" part.
- The spice kernels are required when using Layup

It seems to make sense that we should just use the spice.str2et approach, and remove the Time(datetime.strptime(...)) fall back, which would be replaced with some reasonable logic telling what went wrong.


*Note there is no space between "2000-01-01" and "TDB".
**Note the quotes

drewoldag:
OK, the latest commit uses the following logic: 

Expected input is JD or TDB datestring.
We check to see if naif0012.tls exists. If it doesn't, we error out and tell the user to run `layup bootstrap`. 
If the input cannot be parsed as a float, we'll attempt to use spice.str2et to parse
If an exception is raised by spice.str2et, we'll raise an `ValueError` and prompt the user to try the format "YYYY-mm-ddTDB".

matthewholman (approved):
This looks good to me.

src/layup_cmdline/predict.py
```diff
@@ -26,11 +171,118 @@ def main():
...
+
+    try:
+        # Assume that input is a JD_TDB float that was converted to string. Attempt
+        # to convert it back to a float.
+        date_JD_TDB = float(input_str)
+    except ValueError:
+        try:
+            # If conversion to float fails, assume that the input was a date string
+            # in the format YYYY-mm-ddTDB.
+            et = spice.str2et(input_str)
```
drewoldag, line 215: spice.str2et(input_str).et2utc()
astropy.Time(...) # convert from UTC to TDB. 

Can pull this out later to use this across Layup and potentially Sorcha as well.

drewoldag: Some round trip tests (i.e. input UTC, outputs the same UTC time...) Pedro may have some thoughts on this too.

matthewholman: # This code snippet verifies that et2str and str2et accurately roundtrip
# over the span of times of concern and beyond.

import numpy as np
import spiceypy as spice

dir_path = "/Users/mholman/Library/Caches/layup/"
spice.furnsh(dir_path + "meta_kernel.txt")

et_start = spice.str2et('1950-Jan-1')
et_end = spice.str2et('1990-Jan-1')
ets = np.arange(et_start, et_end, 1e5)

det = []
utcs = []
for et in ets:
    utc = spice.et2utc(et, 'ISOC', prec=10)
    utcs.append(utc)
    et_out = spice.str2et(utc)
    det.append(et_out-et)
    
det = np.array(det)

print(det.min(), det.max())

for utc in utcs:
    et = spice.str2et(utc)
    utc_out = spice.et2utc(et, 'ISOC', prec=10)
    if utc != utc_out:
        print(utc, utc_out)

PR #123, opened 2025-04-19
Input corrections

matthewholman:
I've fixed a few input issues.

This command now gets pretty far:
layup orbitfit -n 1 /Users/mholman/layup/tests/data/100_random_mpc_ADES_provIDs_no_sats.csv ADES_csv

- [ x] Does pip install still work?
- [ x] Does Layup run successfully on a test set of input files/databases?


.github/workflows/smoke-test.yml
```diff
@@ -35,6 +35,7 @@ jobs:
         git submodule update --init
         pip install -e .[dev]
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+        layup bootstrap
```
mschwamb, line 38: We shouldn't need to add this - we should have layup automatically install these files if they're not already installed when it first trys to run these commands.

mschwamb: Maybe this should be a ticket @drewoldag ?

src/layup/orbitfit.py
```diff
@@ -55,16 +56,23 @@ def _orbitfit(data, cache_dir: str):
     # Convert the astrometry data to a list of Observations
     observations = [
         Observation.from_astrometry(
-            d["ra"],
-            d["dec"],
+            d["ra"]*np.pi/180.,
```
mschwamb, line 59: For long term maintainability, I think ra and dec variables like time variables need to specify units.  We can make this a separate ticket for later, but it should be done in the cleanup stage I think

src/layup/orbitfit.py
```diff
@@ -182,16 +193,41 @@ def orbitfit_cli(
 
     chunks = _create_chunks(reader, chunk_size)
 
-    cache_dir = cli_args.ar_data_file_path
+    if cli_args is not None:
+        cache_dir = cli_args.ar_data_file_path
+    else:
```
mschwamb, line 198: Now on line 67-73 is setting kernels_loc but cache_dir is being change or set here but then how is that even being used if there's kernels_loc?

src/layup/utilities/data_processing_utilities.py
```diff
@@ -178,6 +184,8 @@ def obscodes_to_barycentric(self, data, fail_on_missing=False):
                 ("vz", "<f8"),
             ]
             res.append(np.array((x, y, z, vx, vy, vz), dtype=output_dtype))
+            #res.append([x, y, z, vx, vy, vz])
```
mschwamb, line 187: if this isn't correct, can be delete the line rather than comment out.

src/layup/utilities/data_processing_utilities.py
```diff
@@ -178,6 +184,8 @@ def obscodes_to_barycentric(self, data, fail_on_missing=False):
...
 
         # Combine all of our results into a single structured array
         return np.squeeze(np.array(res)) if len(res) > 1 else res[0]
+        #return np.array(res, dtype=float) if len(res) > 1 else np.array(res[0], dtype=float)
```
mschwamb, line 191: if this isn't correct can be delete rather than comment out?

src/layup/utilities/data_processing_utilities.py
matthewholman: That's a landmark to remind me of a place where I think the code should be revisited, but I will add a more meaningful comment.

src/layup/utilities/data_processing_utilities.py
matthewholman: Same thing.  That's a landmark to remind me of a place where I think the code should be revisited, but I will add a more meaningful comment.

src/layup/orbitfit.py
matthewholman: I have added a comment to that effect.

tests/data/1_random_mpc_ADES_provIDs_micro.csv
maxwest-uw, line 1: this needs to be deleted

tests/data/2_random_mpc_ADES_provIDs.csv
maxwest-uw, line 1: this needs to be deleted

maxwest-uw (commented):
let's please freeze this PR for the time being, it needs to be rebased with main to make sure we're not adding in stuff that was removed. I think this was merged with the `orbit_fit_integration` branch half way through me and drew's debugging of the CI issues so there's a lot of stuff in here that shouldn't go to main. plus I can't really tell what the new code actually is 😅

matthewholman:
> let's please freeze this PR for the time being, it needs to be rebased with main to make sure we're not adding in stuff that was removed. I think this was merged with the `orbit_fit_integration` branch half way through me and drew's debugging of the CI issues so there's a lot of stuff in here that shouldn't go to main. plus I can't really tell what the new code actually is 😅

Of course.  It would be fairly straightforward for me to spot the new code.  Is there a way for me to help?

maxwest-uw:
I can try merging it, since most of these commits are already in main it should hopefully resolve properly with a rebase. I'll let you know if I run into any problem!

PR #141, opened 2025-04-24
Issue/132

### LINKED ISSUE #132: re-generalize orbit fit code for more than just AstrometryObservation
for getting the orbit fit basic loop to work properly, we had to make the assumption that we can grab the `rho_hat` and other vectors from the optional `observation_type`, and as such we had to cast that optional as an `AstrometryObservation` (see [here](https://github.com/Smithsonian/layup/blob/main/src/lib/orbit_fit/orbit_fit.cpp#L240C1-L242C88)). `rho_hat`, `a_vec`, and `d_vec` should be accessible from the parent `Observation` class to make sure the code can run generally with all observation types.

from
```
Eigen::Vector3d Av = std::get<AstrometryObservation>(this_det.observation_type).a_vec;
Eigen::Vector3d Dv = std::get<AstrometryObservation>(this_det.observation_type).d_vec;
```

to
```
Eigen::Vector3d Av = this_det.a_vec;
Eigen::Vector3d Dv = this_det.d_vec;
```

### LINKED ISSUE #139: Refactor the observation class to help with stack and shift/rates observations
Refactor the observation class to help with stack and shift/rates observations

maxwest-uw:
Resolves #132 , resolves #139 

Moved `rho_hat`, `a_vec`, and `d_vec` out of `AstrometryObservation` and `StreakObvservation` and into their own methods that are now called from `Observation`.

### DIFF
--- a/src/lib/detection.cpp
+++ b/src/lib/detection.cpp
@@ -33,86 +33,46 @@ namespace py = pybind11;
 
 namespace orbit_fit
 {
+    Eigen::Vector3d rho_hat_from_ra_dec(double ra, double dec)
+    {
+        Eigen::Vector3d rho_hat;
+        rho_hat.x() = std::cos(dec) * std::cos(ra);
+        rho_hat.y() = std::cos(dec) * std::sin(ra);
+        rho_hat.z() = std::sin(dec);
+        return rho_hat;
+    }
 
+    Eigen::Vector3d a_vec_from_rho_hat(const Eigen::Vector3d &rho_hat)
+    {
+        Eigen::Vector3d a_vec;
+        a_vec.x() = rho_hat.z();
+        a_vec.y() = 0.0;
+        a_vec.z() = -rho_hat.x();
+        return a_vec;
+    }
+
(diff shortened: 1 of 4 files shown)


matthewholman:
This looks really good to me, Max!  I have a comment and a question.

First, if you have not already, I think we should verify that the output from the test file, the one that you and Wilson used yesterday or the one that Meg posted, is the same as that from the current main branch.

Second, I really like how you've abstracted the calculation of the A and D vectors.  However, we may need to change the formulas for calculating the A and D vectors on rare occasions, depending upon circumstances.   The formulas would still depend upon rho_hat.  Is there a clean way to allow for different versions of a_vec_from_rho_hat and d_vec_from_rho_hat?

maxwest-uw:
both the 100_random_mpc file and the holman_data_working files run and have the same output as main!

and for the a_vec and d_vec computations, we can have different flags for different situations, or if they're drastically different have separate functions that we can use conditionally

matthewholman:
Thanks for checking that the output is the same.  You are ahead of the game, as usual!

Being able to pass a flag to the a_vec/d_vec functions would work well.

matthewholman (approved):
This looks great to me!

PR #169, opened 2025-04-29
generalizes predict function

### LINKED ISSUE #157: generalize predict function
- make a `PredictResult` struct
- make it have proper input and return values instead of changing memory in place
- create an outer loop that can run from a vector of times to be called from the python level

maxwest-uw:
Fixes #157  .

- creates the `PredictResult` struct that can contain the output of a predict run
- creates an outer loop function for predict to run across many epochs from a single initial condition
- breaks out the predict code into its own file


matthewholman (approved):
This looks great to me.

PR #171, opened 2025-04-29
Implement orbit fit commandline overwrite

wilsonbb:
This wires in the --force argument from the orbit fit commandline. 

If force is not true, if the output file from orbit fit already exists, we raise an error. Otherwise, we delete and overwrite the existing file.


drewoldag (approved):
Looks great to me

PR #178, opened 2025-04-29
adds the jacobian for the transformations

### LINKED ISSUE #33: routines for converting covariance matrix of one orbit format to another format
building routines for converting covariance matrix of one orbit format to another format

bernardinelli:
fixes #33 
No unit tests yet, those will come eventually 

Adds jax and scipy as a dependency

### DIFF
--- a/src/layup/routines/orbit_conversion_jax.py
+++ b/src/layup/routines/orbit_conversion_jax.py
@@ -624,28 +624,28 @@ jac_xyz_cometary = jax.jacobian(universal_cartesian, argnums=(1, 2, 3, 4, 5, 6))
 
 def covariance_cometary_xyz(mu, x, y, z, vx, vy, vz, epochMJD_TDB, covariance):
     r = np.array([x, y, z])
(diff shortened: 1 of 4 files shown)


bernardinelli:
@matthewholman : in particular, can you check the function `covariance_xyz_keplerian`? I wrote the cometary to keplerian jacobian off the top of my head, so it'd be good for someone to double check that :)

src/layup/routines/orbit_conversion_jax.py
drewoldag, line 1: I would advocate for moving this to `.../src/layup/utilities/orbit_conversion.py` (Maybe drop the "jax" from the module name as well???

matthewholman:
@bernardinelli, covariance_xyz_keplerian looks correct to me.  To be really sure, would it be easy to wrap the whole function in jax?

drewoldag (approved):
I won't be so bold as to claim to validate the math, but mechanically this looks fine to me!

matthewholman (approved):
This looks good to me.

PR #186, opened 2025-04-30
Separate orbitfit output into two files

### LINKED ISSUE #124: optionally split layup orbitfit output into successful fits and fit failures
using an optional command line argument the user should be able to split the output of layup orbitfit into two files one that contains all the information from successful fits and one that outputs information related to unsuccessful fits

drewoldag:
Introduce a `split-output` flag that will prompt orbitfit to send output to two separate files. Defaults to `False`

Fixes #124  #192 .

Describe your changes.
Added a CLI flag, `-so, --split-output` that accepts True/False values. Defaults to False. If set to true, a second output file that uses the same stem as the output file name flag with `_flagged` appended will be created. 

The recarrays returned from `_orbitfit` will be filtered into two sets, those with flag==0 and those without.  Those with flag==0 go into the "good" file, those non-zero flags go to the "flagged" file. 


src/layup/orbitfit.py
```diff
@@ -236,14 +257,37 @@ def orbitfit_cli(
             if overwrite:
```
drewoldag, line 257: On the line above this make the logic:
```
if first_write and (os.path.exists(output_file) or os.path.exists(output_file_flagged)):
# if the output_file_flagged thing exists...
```

mschwamb (approved):

wilsonbb (approved):
Approved though I would prefer a flag named something other than split_output 

Maybe split_flagged or split_errors?

drewoldag:
Changed the cli flag name to `--separate-flagged`.

PR #195, opened 2025-04-30
when orbitfit fails the orbit output parameters will be populated with nans

### LINKED ISSUE #180: if fit was not a success populate the orbit output parameters as Nans
Currently the covariance matrix is set to Nans if the orbitfit is not successful - all values should be sent to Nans expect for the objectid column. Relevant lines [here](https://github.com/Smithsonian/layup/blob/97bdfe1ea32e74b726ee5d8ba08f935d2833931f/src/layup/orbitfit.py#L91C1-L111C6).

Little-Ryugu:
Fixes #180  .

changed so that when orbitfit fails the orbit output parameters will be populated with nans

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -95,22 +95,21 @@ def _orbitfit(data, cache_dir: str):
         [
             (
                 data["provID"][0],
-                res.csq,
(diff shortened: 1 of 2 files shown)


drewoldag (approved):
Approved! Thanks :)

PR #198, opened 2025-05-01
add optional sorting by obstime before sending off to C++ fitter

### LINKED ISSUE #197: sort observations to ensure they are in correct time order
if this is not the case then gauss method will struggle to find initial guess in some cases

mschwamb:
add optional sorting by obstime before sending off to C++ fitter

Fixes #197

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -64,14 +64,12 @@ def _orbitfit(data, cache_dir: str, sort_array=True):
     """
     if len(data) == 0:
         return np.array([], dtype=_RESULT_DTYPES)
-    
-    
+
     # sort the observations by the obstime if specified by the user
-    
-    if (sort_array):
-        data=np.sort(data, order="obstime")
-  
-    
+
+    if sort_array:
+        data = np.sort(data, order="obstime")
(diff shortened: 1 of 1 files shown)


src/layup/orbitfit.py
```diff
@@ -65,6 +65,11 @@ def _orbitfit(data, cache_dir: str):
     if len(data) == 0:
         return np.array([], dtype=_RESULT_DTYPES)
 
+    # sort the observations by the obstime if specified by the user
+
+    if sort_array:
+        data = np.sort(data, order="obstime")
```
drewoldag, line 71: We could modify the sorting algorithm to be `heapsort` which _should_ have a faster run time, and since stable sort isn't required, it might be better. Though, we may discover that the average size of the data set is small enough that it doesn't really matter. 

```suggestion
        data = np.sort(data, order="obstime", kind="mergesort")
```

drewoldag (approved):
This looks fine to me. One small comment/suggestion that may not be worth the commit. The only thing to keep in mind for the future is whether `obstime` is going to be specified as the required name for the column representing the time of observation.

PR #202, opened 2025-05-01
Prechecks before orbitfit

### LINKED ISSUE #181: write function to do checks on observations and identify if there's a problem with the input astrometry before sending it on the C++ code for fitting
We want a function that would look at d structure before observation structure is populated  in https://github.com/Smithsonian/layup/blob/97bdfe1ea32e74b726ee5d8ba08f935d2833931f/src/layup/orbitfit.py#L49

the function should return a bool value true if orbitfit should proceed or not actually call the orbitfit a fail and skip sending on to the C++ code and set the success flag to be a failure so the output is all nans in the orbit parameters data structure other than objectID (related to #180)

Conditions when we should not sent on to the C++ orbit fitting code:

- if data["ra'] or data["dec"] is not numeric (aka Nans or nones)
-  len of data is less than 3 
-  data[et] is less than zero 
-   any of these values data["x"], data["y"], data["z"], data["vx"], data["vy"], data["vz"]

Little-Ryugu:
Fixes #181 .

Added prechecks to data in python before sentind data to c++. 

if prechecks failed the output for the object is filled with nans and flag = -1

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -65,7 +65,6 @@ def _orbitfit(data, cache_dir: str, sort_array=True):
 
     if _is_valid_data(data):  # checks data being supplied to c ++ code is valid
 
-
         # sort the observations by the obstime if specified by the user
         if sort_array:
             data = np.sort(data, order="obstime", kind="mergesort")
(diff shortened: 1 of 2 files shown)


matthewholman (approved):
This looks good to me.

PR #211, opened 2025-05-02
Issue/159/initial guess

### LINKED ISSUE #159: make `run_from_vector` work off of `initial_guess` function
we have a lot of code in parallel at the moment between `run_from_vector` and the initial guess version. `run_from_vector` is going to be the "default" routine for orbit fit running, and will therefore build its own initial guess than pass it down to the `initial_guess` function, while a user who wants to do their own custom initial guess can call the `initial_guess` function directly.

matthewholman:
Fixes #159

This makes run_vector call run_vector_with_initial_condition once an initial guess or the solution of a previous iteration is available.


src/lib/orbit_fit/orbit_fit.cpp
```diff
@@ -1063,6 +1050,173 @@ namespace orbit_fit
...
+	    }
+
+	    FitResult ffr = first_fit_res.value();
+
+	    // Should check if num is already larger than the full set of detections.
+            num = detections_full.size();
+
+            first_d = detections_full.begin() + detections_full.size() - num;
+            last_d = detections_full.end();
+            std::vector<Observation> detections2(first_d, last_d);
```
maxwest-uw, line 1153: is the creation of `detections2` actually necessary for the algorithm? It seems like because `num` is equivalent to `detections_full.size()` and is then subtracted from that value, that `detections2` is always equal to detections_full `

matthewholman: It is not presently necessary, but I'm still tinkering with that.

maxwest-uw (approved):

Issue #85, opened 2025-03-14
Be aware of ecliptic vs equatorial inputs


drewoldag:
We need to declare all orbit inputs need to be in *ecliptic* coordinates. 

There may be changes that can be made in the code that will remove this requirement. But if nothing else we should create documentation that makes this very clear. 

The root of this is because we use Sorcha's `parse_orbit_row` in the `convert` verb. `parse_orbit_row` expects ecliptic and returns equatorial. In `convert` we convert the output of `parse_orbit_row` back to ecliptic because we assume that is the coordinate system that the user provided.

mschwamb:
We have a plan forward which is a new orbit type BCARTEQ - barycentric cartesian equatorial - this is actually what all the C++ is expecting orbit information to be in

Issue #219, opened 2025-05-08
read in optional initial guess for orbit fit  intput orbits and convert if needed to BCARTEQ before sending to C++


mschwamb:
read in optional initial guess for orbit fit  intput orbits and convert if needed to BCARTEQ before sending to C++

mschwamb:
related to #85

wilsonbb:
Closed by https://github.com/Smithsonian/layup/pull/241

Issue #218, opened 2025-05-08
read in predict input orbits and convert if needed to BCARTEQ before sending to C++


mschwamb:
read in predict orbit intput orbits and convert if needed to BCARTEQ before sending to C++

wilsonbb:
From the meeting I came away with the impression that we just want to rename our current BCART to BCARTEQ, but I want to check in if we should add that as a 7th format. So we would still have BCART as barycentric cartesian ecliptic (same as sorcha), and then we would just add in BCARTEQ

mschwamb:
BCART = barycentric cartesian ecliptic (same as sorcha). We would just add in BCARTEQ

mschwamb:
But our current BCART in many places like the orbitfit output right now is really BCARTEQ. So that's just a swap of name  for here https://github.com/Smithsonian/layup/blob/5b6b84ca7766588bd1fe93a9de15a2d3d2f16e40/src/layup_cmdline/orbitfit.py#L136 would go to BCARTEQ

Convert would need to hand converting into/out of  BCARTEQ  and BCART. @bernardinelli can comment what changes needed to be made in convert

PR #224, opened 2025-05-08
Add Explicit BCART_EQ Format

wilsonbb:
Fixes https://github.com/Smithsonian/layup/issues/85

Pre-req for https://github.com/Smithsonian/layup/issues/218 and https://github.com/Smithsonian/layup/issues/219

Adds in an explicit BCART_EQ format which is equatorial barycentric cartesian. We add the ability to read in this format across the various verbs, and we add the ability to convert between BCART_EQ and the other formats in layup convert.

Previously we assumed that BCART was always equatorial, but after this change we assume it is ecliptic to match up with sorcha: https://sorcha.readthedocs.io/en/latest/inputs.html

The current default output of orbit fit is BCART_EQ, and the various test files from orbit fit output have had their format column values changed from BCART to BCART_EQ

Describe your changes.


src/layup/utilities/orbit_conversion.py
```diff
@@ -704,9 +706,13 @@ def parse_covariance_row_to_CART(row, gm_total, gm_sun):
...
-        # It is simply a translation of the covariance matrix between,
-        # the two formats so return it as is.
+    if init_format == "BCART_EQ":
+        # We are in equatorial barycentric cartesian coordinates
+        return cov
+    elif init_format in ["CART", "BCART"]:
+        # TODO we should rotate from equatorial to ecliptic coordinates
+        # Since this is a translation we do not need to do anything
+        # differently for CART vs BCART
         return cov
```
wilsonbb, line 716: @bernardinelli 

So this function currently returns the covariance matrix to cartesian equatorial regardless of initial format.

For consistency, I believe I should be converting the covariance matrix to equatorial here, while leaving BCART_EQ unchanged above

bernardinelli: Yes, I agree! So we should do the following:
```
jj_rotation = np.array(block_diag(ECL_TO_EQ_ROTATION_MATRIX.T, ECL_TO_EQ_ROTATION_MATRIX.T))
return jj_rotation @ cov @ jj_rotation.T
```

wilsonbb: Done, thank!

PR #230, opened 2025-05-13
Editing the covariance matrix column names and fixing some bugs.

drewoldag:
The test data file that was being used for testing `convert` was using the `cov_[00-36]` column naming scheme for the covariance columns, when what we actually wanted was `cov_[0-5]_[0-5]`.  

The byproduct of that was that a check for expected covariance columns returned `False` and sent the logic down a code path that did not exercise the new @jax.jit-decorated covariance conversion functions. When the test data file was corrected, the covariance conversion functions _were_ executed, and resulted in jax-related failures. 

Commenting out the `@jax.jit` decorator allows the unit tests to pass, but doesn't identify the source of the jax-related test failure. 


src/layup/utilities/orbit_conversion.py
```diff
@@ -624,7 +624,7 @@ def universal_keplerian(mu, x, y, z, vx, vy, vz, epochMJD_TDB):
 jac_keplerian_xyz = jax.jacobian(universal_keplerian, argnums=(1, 2, 3, 4, 5, 6))
 
 
-@jax.jit
+# @jax.jit
 def covariance_ecl_to_eq_jax(covariance):
```
drewoldag, line 628: Is there a particular reason why this was rename to `..._jax` when all the other jax.jit decorated functions don't have the `_jax` suffix?

wilsonbb: Hmm I don't recall my particular reason for when I was writing the function. I don't think there's a reason to preserve it

wilsonbb (approved):
Thanks for resolving this!

PR #259, opened 2025-05-15
Update prediction time to UTC

### LINKED ISSUE #242: layup predict should output the correct time &  times as human readable date & JD
so epoch doesn't seem to changing in the predict output - I think it might be something about the reference for the cartesian, but it's not the time of those positions. 

We'll want the time outputted in two ways one as a human readable  UT TDB  and the other as JD TDB.

mschwamb:
Fixes #242

shifts time of prediction to front of structure, and renames to "epoch_UTC", providing a human-readable UTC string in the column.

Note that in order to furnish the spiceypy kernels, I broke that method outside of the LayupObservatory class.


tests/layup/test_predict.py
```diff
@@ -72,6 +72,12 @@ def __init__(self, g=None):
     assert np.all(output_data["ra_deg"] <= 360.0) and np.all(output_data["ra_deg"] >= 0.0)
     assert np.all(output_data["dec_deg"] <= 90.0) and np.all(output_data["dec_deg"] >= -90.0)
 
+    # Ensure that the epoch_utc column is present and in the correct format
+    assert all(isinstance(epoch, str) for epoch in output_data["epoch_UTC"])
+    assert all(len(epoch) == 20 for epoch in output_data["epoch_UTC"])
+    # All of our start and end dates for our predictions are in the year 2026
+    assert all(epoch.startswith("2026") for epoch in output_data["epoch_UTC"])
+
```
drewoldag, line 80: Would it be worth adding a single `assert output_data["epoch_UTC"][0] == "2026-MM-dd:HH:mm:ss"`?

drewoldag (approved):
Cool, this seems reasonable. Only one small comment, but not blocking.

tests/layup/test_predict.py
wilsonbb: Done.

PR #263, opened 2025-05-16
Support multiple IODs in `orbitfit`

### LINKED ISSUE #220: orbitfit python code set up to call another IOD based on flag
We have a command line argument in the CLI for this - so we just need a switch in the python to call a dummy function that for outputs any orbit - for now the orbit of asteroid Holman as the initial guess

drewoldag:
Fixes #220 

Describe your changes.

- Pass the iod CLI around the orbit fit code. 
- Created a stub do_other_fit for potential future complicated IODs that require different data massaging
- Extracted gauss IOD from `do_fit`, and added the beginning of a if/else ladder there, again for future IODs that don't require as much data massaging as those that might end up in the `do_other_fit`-like functions.

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -306,14 +306,39 @@ def create_empty_result(id, dtypes):
     )
 
 
-def do_fit(observations, seq, cache_dir):
+def do_gauss_iod(observations, seq):
+    """Calculate an initial orbit estimate using Gauss's method.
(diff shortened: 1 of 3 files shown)


mschwamb:
This looks good to me @matthewholman  what do you think?

matthewholman (approved):
Yes, this looks good to me!

PR #264, opened 2025-05-16
makes csq and ndof non-required fields for predict

### LINKED ISSUE #248: skip reading in chisquare column as input for predict
I think the chisquare isn't needed for predict. It would be good if this column wasn't read in as i don't see how the C code needs it. This would allow someone to use their own independent orbit fit they want to

maxwest-uw:
Resolves #248 

uses the `orbit_colm_flag` in `parse_fit_result` to ensure we don't need the `csq` and `ndof` fields for predict. Also adds a unit test that makes sure that predict doesn't use or care about those values. 


drewoldag (approved):
This all looks good to me.

PR #274, opened 2025-05-20
First implementation of predict covariance matrix

### LINKED ISSUE #271: add RA/Dec uncertainty or error ellipse calculation to predict
currently we have cartesian covariance but no uncertainties or error ellipse - it would be good to get on-sky uncertainties as an enhancement to predict

bernardinelli:
Fixes #271 

As @matthewholman said in #271, we should still test this a bit, but this should be mathematically correct 

One detail for the PR reviewer: see if I understood the `obs_covi` entries correctly

### DIFF
--- a/src/layup/predict.py
+++ b/src/layup/predict.py
@@ -35,10 +35,12 @@ def _get_result_dtypes(primary_id_column_name: str):
             ("rho_x", "f8"),  # The first of the 3 rho unit vector
             ("rho_y", "f8"),
             ("rho_z", "f8"),
-            ("obs_cov0", "f8"),  # The first of 4 of the observer covariance
-            ("obs_cov1", "f8"),
-            ("obs_cov2", "f8"),
-            ("obs_cov3", "f8"),
+            ("obs_cov_xx", "f8"),  # The first of 4 of the observer covariance
+            ("obs_cov_yy", "f8"),
+            ("obs_cov_xy", "f8"),
+            ("a_arcsec", "f8"),
+            ("b_arcsec", "f8"),
+            ("PA_deg", "f8"),
         ]
     )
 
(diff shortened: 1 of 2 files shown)


matthewholman (approved):
This looks correct to me.  I agree with your interpretation of the obs_cov entries.

bernardinelli:
@maxwest-uw Do I need to declare the extra columns I'm creating in the results structured array in `_get_result_dtypes` too? (I see that's where the unit test failed!)

src/layup/predict.py
```diff
@@ -107,6 +108,9 @@ def _predict(data, obs_pos_vel, times, cache_dir, primary_id_column_name):
 
     results = np.array(predict_results, dtype=_get_result_dtypes(primary_id_column_name))
     results["ra_deg"], results["dec_deg"] = vec2ra_dec([results["rho_x"], results["rho_y"], results["rho_z"]])
+    results["a_arcsec"], results["b_arcsec"], results["PA_deg"] = skyplane_cov_to_radec_cov(
+        results["ra_deg"], results["dec_deg"], results["obs_cov0"], results["obs_cov3"], results["obs_cov1"]
```
maxwest-uw, line 112: why don't we use the yx covariance (`obs_cov2`)? is it the same as the xy one?

maxwest-uw (approved):
lgtm, have a question out of pure curiosity 😅

src/layup/predict.py
matthewholman: Yes, it's symmetric.  Good question!

bernardinelli: Which also makes me think we don't need to save the four covariance entries :)

matthewholman: What was the issue with the unit tests?

bernardinelli: The columns for the uncertainty ellipse were not being created in the structure we're returning. I ended up not having time to fix it last week, but I just did. Let me push it!

matthewholman: Thanks!

Issue #144, opened 2025-04-25
populate observatory position class instance based on user provide positional information if available


mschwamb:
needed for handling satellites and mobile observatories where providing the obscode is not enough to create the observatory location class. Issue #143 handles reading this optional information in for ADES CSV/HDF5 input - This is being handled in reading in #52 for 80 column format.  

The 80 column work may cover this, but in case not this ticket is for taking the variables with the position information from ADES and creating the proper observatory  location class instance. 

Here's what the ADES manual states the important columns are:

[observatorypos.pdf](https://github.com/user-attachments/files/19914269/observatorypos.pdf)

mschwamb:
ADES format does not allow an observatory location group for radar observation . So no velocities are sent to the MPC in ADES format for new radar observations, and it looks like 80 column doesn't provide velocity info either. I made a separate ticket for properly reading in and handling radar/doppler observations in issue #146 

The ADES manual states that velocity information is optional for space-based observations and not permitted for terrestrial observers. Given there is no example observation with velocity values in the MPC, I propose that we develop layup right now such that any required spacecraft velocities get calculated from Horizons or spice kernels (which is already in issue #55)  and it be a enhancement to include user calculated spacecraft velocities (issue #149)

Issue #143, opened 2025-04-25
expand HDF5 reader and CSV reader to read in position of observatory (if provided) 


mschwamb:
Now that there's a full loop for orbitfit, we should the expand HDF5 reader and CSV reader to read in positional of observatory (if provided) , apparent mag/filter, and occultation info - this ensures occultation observations and satellite observations can be processed 

I've tried to go through the [ADES manual](https://github.com/IAU-ADES/ADES-Master/blob/master/ADES_Description.pdf) and find the relevant terms and update  these CSV files:

[holman_data_with_sats_occ.csv](https://github.com/user-attachments/files/19913793/holman_data_with_sats_occ.csv)

[100_random_mpc_ADES_latest_provid_only_with_sats_occ.csv](https://github.com/user-attachments/files/19913796/100_random_mpc_ADES_latest_provid_only_with_sats_occ.csv)

[bernstein_2004_kbos_data_with_sats_occ.csv](https://github.com/user-attachments/files/19913797/bernstein_2004_kbos_data_with_sats_occ.csv)

(the modified notebook that makes that files can be found here)

By default layup orbitfit uses the obscode to calculate the observatory position, but if the position is provided such as in the case of mobile or satellites than it should come from what's inputted using the sys, ctr, pos1,pos2, pos3 columns 

---

*The following was originally part of this ticket, but is now is being tracked in issue #204*
If RA and Dec are not populated then starRA, starDec, deltar, deltadec  should be because it's an occultation measurement and then the regular ra and dec variables can be filled via RA = starra+deltra/cos(stardec) Dec=stardec+deltadec

There's also the catalog column to read in the astcat column which will be needed for #51 

Matt mentioned earlier having C++ variables for apparent magnitude so mag, rmsmag, and band should be read in to capture that information if available

@matthewholman and @kjnapier, please comment/revise/double check

mschwamb:
[holman_data_with_sats_occ.csv](https://github.com/user-attachments/files/19915342/holman_data_with_sats_occ.csv)

Update asteroid Holman info without the times that before the 1960s that causes issues 

[holman_data_with_sats_occ.csv](https://github.com/user-attachments/files/19915375/holman_data_with_sats_occ.csv)

drewoldag:
A separate issue has been created for tracking expanding HDF5 reader and CSV reader to read in apparent mag/filter, catalog, and occultation info - issue #204

PR #277, opened 2025-05-22
Allow LayupObservatory to Handle User-Provided Observatory Positions

### LINKED ISSUE #268: ensure that stn/station is read in as a string
if not read in as a string can cause issues with getting the correct Observatory position

wilsonbb:
Works towards #144, by allowing the LayupObservatory class to check for the geocentric observatory positions that are provided by MPC Obs 80 Col format. This is the same format and units used by Sorcha's Observatory class from which LayupObservatory inherits. 

Note that this does not complete #143 which would need to convert the user-provided observatory info in ADES to this format and populate the 'obs_geo_x', 'obs_geo_y', and 'obs_geo_z' columns.

LayupObservatory maintains a map of obscode to position, and this change uses that same structure with a key of the form {obscode}_{et} for moving observatories.

To help testing for these changes, a couple of related changes were made in this PR:
* Removing the flag to not raise errors on missing observatories (AKA satellites)
* Enforcing that the observatory 'stn' column in ADES is always read in as a string. Fixes #268 
* Adding some additional type checking that the observatory code is a string.

Describe your changes.


src/layup/utilities/file_io/CSVReader.py
```diff
@@ -57,7 +57,7 @@ def __init__(self, filename, sep="csv", **kwargs):
         super().__init__(**kwargs)
         self.filename = filename
 
-        if sep not in VALID_FILE_FORMATS.keys():
+        if sep not in VALID_FILE_FORMATS:
```
drewoldag, line 60: Thank you!

src/layup/utilities/data_processing_utilities.py
```diff
@@ -316,7 +316,83 @@ def __init__(self, cache_dir=None):
...
+
+            if obscode_cache_key not in self.ObservatoryXYZ:
+                # Store the coordinates in the ObservatoryXYZ dictionary to be read by barycentricObservatoryRates
+                self.ObservatoryXYZ[obscode_cache_key] = coords
+            else:
+                # If the coordinates are not the same, raise an error
+                if not np.allclose(self.ObservatoryXYZ[obscode_cache_key], coords):
+                    raise ValueError(
+                        f"Observatory {obscode} has different coordinates at different epochs. "
+                        f"Coordinates at epoch {et} are {coords}, but at epoch {self.ObservatoryXYZ[obscode_cache_key]}."
```
drewoldag, line 389: I'm sorry, having a little trouble parsing this f-string. What will the final part print out? "but at epoch ???"

drewoldag (approved):
Overall this looks ok to me. Nice work!

src/layup/utilities/data_processing_utilities.py
wilsonbb: Ah this was very confused, thanks for catching. Corrected to 

```
                # If the coordinates are not the same, raise an error
                if not np.allclose(self.ObservatoryXYZ[obscode_cache_key], coords):
                    raise ValueError(
                        f"Observatory {obscode} has different coordinates reported at the same epoch."
                        f"Coordinates at epoch {et} previously were {self.ObservatoryXYZ[obscode_cache_key]}, but are now {coords}."
                    )
```

This error should catch when the user has been reporting different positions for the same observatory and epoch, outputting the obscode and epoch as well as the previously reported coordinates and the current coordinates.

PR #279, opened 2025-05-22
Defined minimum required column names in `predict` and `orbit`

### LINKED ISSUE #234: layup orbitfit check if minimum required columns exists
obstime wasn't in the input file I was using and it faulted out trying to covert the time

drewoldag:
Fixes #234 

Describe your changes.
- Adding required column lists to `predict` and `orbitfit`
- Fixed a bug in `ObjectReader`
- Added tests to prevent regression

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -35,6 +35,20 @@ from layup.utilities.file_io.file_output import write_csv, write_hdf5
 
 logger = logging.getLogger(__name__)
 
+# The list of required input column names for the provided observations to be fit.
+# Note: This should not include the primary id column name.
+REQUIRED_INPUT_OBSERVATIONS_COLUMN_NAMES = [
+    "ra",
+    "dec",
+    "obstime",
+]
+
+# The list of required column names for the guessed orbits (if provided).
+# Note: This should now include the primary id column name.
+REQUIRED_INPUT_GUESS_COLUMN_NAMES = [
+    "epochMJD_TDB",
+]
+
 INPUT_FORMAT_READERS = {
     "MPC80col": (Obs80DataReader, None),
     "ADES_csv": (CSVDataReader, "csv"),
@@ -765,7 +779,12 @@ def orbitfit_cli(
     if reader_class is None:
         logger.error(f"File format {input_file_format} is not supported")
 
-    reader = reader_class(input_file, primary_id_column_name=_primary_id_column_name, sep=separator)
+    reader = reader_class(
+        input_file,
+        primary_id_column_name=_primary_id_column_name,
+        sep=separator,
+        required_columns=REQUIRED_INPUT_OBSERVATIONS_COLUMN_NAMES,
+    )
     if guess_file is not None:
         # Check that the guess file exists
         if not guess_file.exists():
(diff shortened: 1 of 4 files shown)


src/layup/orbitfit.py
```diff
@@ -35,6 +35,20 @@
 
 logger = logging.getLogger(__name__)
 
+# The list of required input column names for the provided observations to be fit.
+# Note: This should not include the primary id column name.
+REQUIRED_INPUT_OBSERVATIONS_COLUMN_NAMES = [
+    "ra",
+    "dec",
+    "obstime",
+]
```
drewoldag, line 44: @mschwamb Please let me know if there are columns here that I've missed.

src/layup/orbitfit.py
```diff
@@ -35,6 +35,20 @@
...
+
+# The list of required column names for the guessed orbits (if provided).
+# Note: This should now include the primary id column name.
+REQUIRED_INPUT_GUESS_COLUMN_NAMES = [
+    "epochMJD_TDB",
+]
```
drewoldag, line 50: @mschwamb Please let me know if there are columns here that I've missed.

src/layup/predict.py
```diff
@@ -20,6 +20,12 @@
 from layup.utilities.file_io import CSVDataReader
 from layup.utilities.file_io.file_output import write_csv
 
+# The list of required input column names. Note: This should not include the
+# primary id column name.
+REQUIRED_INPUT_COLUMN_NAMES = [
+    "epochMJD_TDB",
+]
```
drewoldag, line 27: @mschwamb Please let me know if there are columns here that I've missed.

src/layup/orbitfit.py
wilsonbb: It feels like "stn" maybe as well? But we do check for it in other ways

mschwamb: stn is definitely required - it's possible someone could only have stack and shift and then would only have the raRate/decRate - (case is changing) not sure if we can do an or - if not at least checking the obstime and stn would be good

src/layup/orbitfit.py
mschwamb: We need FORMAT in there as well - ideally the other columns depend on what the input format is - not sure that's doable - otherwise at least checking these two is important allows doing finer checking later

drewoldag: Added `FORMAT`. The guess data is run through `convert` which has substantial column checking logic built in. So we should be ok here to just require epochMJD_TDB and FORMAT.

src/layup/orbitfit.py
drewoldag: Added 'stn' and also added support for defining a tuple of possible column names. So that you can do something like this: 
```
REQUIRED_COLUMNS = [
    'obstime',
    ('ra', 'raRate'),
    ('dec', 'decRate')
]
```

I think that this gets at the sprit of what you're asking for, but the logic is light. For instance if an input file had the columns `['obstime',` 'ra', 'decRate']` the check would pass. 

The thought of maintaining the index of the column name that worked crossed my mind, but that is really challenging for a maintainer to remember, and difficult to implement if there are non-related sets of column names. 

For instance something like this:
```
REQ_COLS =  
    'obstime',
    ('ra', 'raRate'),
    ('unrelated_to_ra', 'other_unrelated_to_ra')
]
```

It wouldn't help to track which index "worked" for the ra column names and try to apply that to the "unrelated_to_ra" columns.

Long story short - be careful with this newly added power.

PR #284, opened 2025-05-23
Enforce ingesting "stn" column as a str in CSVReader.read_objects

wilsonbb:
The previous fix to enforce reading the `stn` column as a string for the CSV reader was only being applied for read_rows and not read_objs. This is important since other unit tests used read_rows rather than read_objs (which is actually what is used by orbitfit).

Here we fix that and add a test file which has only MPC obscodes that are numeric (which is the case that triggers this bug).


drewoldag (approved):
Good catch!

PR #285, opened 2025-05-23
Setting up logging

### LINKED ISSUE #270: setup logging
we have logging in places but we're not actually outputting those log messages somewhere - if from the command line then let's send it to STDERR for now

drewoldag:
First pass to get logging set up. 

Fixes #270 

Describe your changes.
- Create a central logger configuration module under utilities.
- Did a first pass at replacing `print` statements with `logger.info`
- First pass as augmenting sys.exit statements with `logger.error` 
- Each command line verb will now instantiate a LayupLogger object like so: 
```
def execute():
    from layup.utilities.layup_logging import LayupLogger
    layup_logger = LayupLogger() # Can optionally pass `log_directory` i.e. from a CLI arg. LayupLogger(log_directory=<dir>)
    logger = layup_logger.get_logger("layup.<verb_name>") # The 'layup.' part will place this logger under the root logger
    logger.info("Log a message")
    # the rest of the logic
```

- When running from a notebook, can use a context manager to log like so:
```
from layup.utilities.layup_logging import LayupLogger

with LayupLogger() as layupLogger:
    logger = layupLogger.get_logger("layup.<some_name>")
    # do something
```


drewoldag:
FYI - I reverted the change in location of the command line verbs. They are now back outside the `.../src/layup` directory for the reasons that Meg was calling out.

src/layup/log.py
```diff
@@ -0,0 +1,52 @@
...
+    samples = 25
+    num_workers = os.cpu_count()
+
+    result = log(samples, num_workers)
+    logger.debug(f"Processed data with this length: {len(result)}")
+
+    result = log_by_id(samples, num_workers)
+    logger.debug(f"Processed data with this length: {len(result)}")
+
+    logger.error("I think you're great.")
```
wilsonbb, line 52: I am still amused by this

src/layup/log.py
```diff
@@ -0,0 +1,52 @@
+import logging
+import os
+
+import numpy as np
+
+from layup.utilities.data_processing_utilities import process_data, process_data_by_id
+
+logger = logging.getLogger(__name__)
+
+
```
wilsonbb, line 10: Can we add some comments to this file noting that the functions in this file for the layup `log` verb which is for demo purposes?

src/layup/predict.py
```diff
@@ -20,6 +21,8 @@
 from layup.utilities.file_io import CSVDataReader
 from layup.utilities.file_io.file_output import write_csv
 
+logger = logging.getLogger(__name__)
```
wilsonbb, line 24: To verify my understanding, here we are deliberately not calling and importing LayupLogger since we only want that to be used when calling the corresponding verb from the commandline? For an API caller they can provide their own logger, use LayupLogger, etc.

If this is the expected pattern, would it make sense to have some comments about this in LayupLogger or call it LayupCliLogger (I personally hate that name, but maybe communicates to future maintainers how it should be used)

wilsonbb (approved):
Overall looks good to me and worked to produce log files for me when I ran

`layup orbitfit /Users/wilsonbb/lincc/layup/tests/data/100_random_mpc_ADES_provIDs_no_sats.csv ADES_csv -n 4 -f --chunksize 4000`

Just a few questions about comments

src/layup/log.py
drewoldag: Added doc strings and comments throughout.

mschwamb (approved):

PR #296, opened 2025-10-28
Sexagesimal Output

### LINKED ISSUE #269: ability to output predict ra and dec in sexagesimal format with RA in hours and Dec in degrees
ability to output predict ra and dec in sexagesimal format with RA in hours and Dec in degrees - probably should be a command line flag with default in degrees for now

awilson110:
Fixes #269

Added function and command line flag to give users the option to have their output .csv file contain the coords in hours, minutes, seconds for RA and degrees, minutes, seconds for Dec. These are called ra_str_hms and dec_str_dms respectively. Before writing the .csv file the columns ra_str_hms and dec_str_dms are brought further to the front using pandas. Added known_sexagesimal.csv for unit testing that the conversion works correctly and handles edge cases well, and added holman_expected_predict_sg.csv to unit test that it outputs the columns correctly. 

### DIFF
--- a/src/layup_cmdline/predict.py
+++ b/src/layup_cmdline/predict.py
@@ -179,7 +179,7 @@ def main():
         "-sg",
         "--sexagesimal",
         action="store_true",
-        help="Option for output units of RA and Dec to be given in sexagesimal.",
+        help="Flag to add RA and Dec in sexagesimal format to the output.",
     )
 
     args = parser.parse_args()


src/layup/predict.py
```diff
@@ -55,6 +55,40 @@ def _get_result_dtypes(primary_id_column_name: str):
...
+
+    Returns
+    -------
+    input array with ra and dec in sexagesimal appended, called ra_str_hms and dec_str_dms respectively.
+    """
+    ra_deg = (data["ra_deg"] / 15) % 24.0  # Ensuring ra is within 24 hours/360 degrees
+    dec_deg = data["dec_deg"]
+    ra_decimal = ra_deg % 1.0  # Take decimal portion to find arcminutes
+    dec_decimal = np.abs(dec_deg) % 1
+    ra_m = ra_decimal * 60
```
mschwamb: make these ints give they are ints in the string

src/layup/predict.py
```diff
@@ -55,6 +55,40 @@ def _get_result_dtypes(primary_id_column_name: str):
...
+    dec_m = dec_decimal * 60
+    ra_decimal = ra_m % 1  # Take decimal portion again for arcseconds
+    dec_decimal = dec_m % 1
+    ra = np.empty(len(ra_deg), dtype="<U16")
+    dec = np.empty(len(ra_deg), dtype="<U16")
+    for i in range(len(ra_deg)):
```
mschwamb: add an empty new line for read ability before line 80

src/layup/predict.py
```diff
@@ -55,6 +55,40 @@ def _get_result_dtypes(primary_id_column_name: str):
...
+    """
+    ra_deg = (data["ra_deg"] / 15) % 24.0  # Ensuring ra is within 24 hours/360 degrees
+    dec_deg = data["dec_deg"]
+    ra_decimal = ra_deg % 1.0  # Take decimal portion to find arcminutes
+    dec_decimal = np.abs(dec_deg) % 1
+    ra_m = ra_decimal * 60
+    dec_m = dec_decimal * 60
+    ra_decimal = ra_m % 1  # Take decimal portion again for arcseconds
+    dec_decimal = dec_m % 1
+    ra = np.empty(len(ra_deg), dtype="<U16")
```
mschwamb: add a blank line for readability before line 78

src/layup_cmdline/predict.py
```diff
@@ -175,6 +175,16 @@ def main():
         required=False,
     )
 
+    optional.add_argument(
+        "-us",
+        "--units-sexagesimal",
+        help="Option for output units of RA and Dec to be given in sexagesimal. False will use degrees.",
+        dest="units",
```
mschwamb: because you have dest units - it's looking for the user to put some after the -us on the command line. There isn't anything for the user to specify. You want this to be like force flag. The reason why is it it defaults to false I don't need to raise the flag unless I want it to be true

src/layup/utilities/file_io/file_output.py
```diff
@@ -14,8 +14,19 @@ def write_csv(data, filepath):
         The data to write to the file.
     filepath : str
         The path to the file to write.
+    move_columns : dict, optional
+        Dict of any column names that need moved, paired with their new position.
     """
     df = pd.DataFrame(data)
+
+    if move_columns != {}:
+        column_names = list(df.columns.values)
```
mschwamb: what happens if the user sent a column name that doesn't exist in the data array? Need to check and error on that. you might be able to do a try and except

https://www.w3schools.com/python/python_try_except.asp

mschwamb (changes requested):
Getting closer

src/layup/utilities/file_io/file_output.py
```diff
@@ -5,7 +5,7 @@
 import pandas as pd
 
 
-def write_csv(data, filepath):
+def write_csv(data, filepath, move_columns={}):
```
mschwamb: Also I think it's better to not have an empty array as the default value. Better to use None instead of {} https://www.codiga.io/blog/python-avoid-empty-array-parameter/

src/layup/utilities/file_io/file_output.py
```diff
@@ -14,8 +14,22 @@ def write_csv(data, filepath):
...
     df = pd.DataFrame(data)
+
+    if move_columns != {}:
+        column_names = list(df.columns.values)
+        for col in move_columns.keys():
+            try:
+                column_names.pop(column_names.index((col)))
+                column_names.insert(move_columns[col], col)
+            except:
+                raise ValueError(f"column {col} not found in df.columns.values.")
```
mschwamb: Have you tested that this works by sending the code the incorrect thing through? It would be good to add this to the unit tests. https://stackoverflow.com/questions/129507/how-do-you-test-that-a-python-function-throws-an-exception

mschwamb (changes requested):
I think this should be the last set of requested changes.

src/layup/utilities/file_io/file_output.py
awilson110: I added a new unit test to test the column moving of write_csv, and one thing it tests is if returns an error for nonexistent columns

PR #321, opened 2026-04-15
Fixdocs

HoodedCrow:
This allows Layup to be built as an installable wheel. This is a step after PR320 trying to fix documentation build on CI. PR320 has been successfully tested on Mac in addition to Ubuntu.

-    Removes assist and rebound as git submodules and adds them to dependencies instead
-    Removes no longer needed create_lib_links.sh
-    Adds explicit wheel.packages so that layup_cmdline ends up in the wheel
-    Adds a tool.pdm.scripts section for tests.

Note1: with this change all .so end up in site-packages, so you only need to add that to LD_LIBRARY_PATH, no need for manual lib links
Note2: the change should be platform-independent (CMake asks Python for correct extension and paths), but I only have Ubuntu to test on

Checks:

-    pdm test runs locally
-    black only complains about a file I didn't touch
-    pdm build make a wheel
-    using "layup @ [git+git@github.com](mailto:git+git@github.com):HoodedCrow/layup.git" in dependencies of another project works

- [NA] Have you written a unit test for any new functions?
- [x ] Do all the units tests run successfully?
- [ x] Have you used black on the files you have updated to confirm python programming style guide enforcement?

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -41,5 +41,4 @@ jobs:
         pip list
     - name: Run unit tests with pytest
       run: |
-        export LD_LIBRARY_PATH=$(pwd):$LD_LIBRARY_PATH
         python -m pytest -n auto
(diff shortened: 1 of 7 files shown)


HoodedCrow:
Folded changes into PR320 after someone confirmed it worked on Mac as well.

PR #324, opened 2026-05-12
Pluggable IOD layer, multi-root picker, residual prefilter, and IAS15 min-dt floor

matthewholman:
Refactors `do_fit`'s root-handling and LM machinery into a coherent pipeline:

This was developed iteratively with Claude code.

  1. **Pluggable IOD layer** (`src/layup/iod.py`)
     - Registry-based: `register_iod(name, callable)`, `get_iod(name)`,
       `iod_methods()`. An IOD is any
       `(observations, seq) -> list[FitResult] | None`.
     - Gauss ships as the default. Future IOD methods (Lambert,
       motion-rate prior, etc.) drop in without modifying `do_fit`.
     - `do_fit(..., iod=...)` accepts either a registered name or a
       callable directly.

  2. **Multi-root picker** — LM from every IOD candidate at
     `screen_iter_max=80`, picks smallest-χ² converged root with
     r > 0.3 AU. Falls back to `full_iter_max=100` if nothing converges
     at the cheap tier. Closes the historical pitfall where `do_fit`
     committed to `solns[0]` (largest r) even when that was a phantom
     outer-SS solution for an inner-SS target.

  3. **Residual prefilter** — drops candidates whose predicted
     positions miss the observations by more than `threshold_sigma=1000σ`
     *before* running LM. Candidates whose inertial trajectory passes
     within 0.1 AU of the observer are passed through unfiltered (real
     NEO close approaches need their own integrator-budget solution,
     addressed by #4 below).

  4. **IAS15 min-dt floor** — `sim->ri_ias15.min_dt` is floored at
     1e-3 days during the picker LM loop. Caps integrator wallclock on
     phantom roots whose trajectories pass close to Earth (1000× speedup
     on diagnostic/scan, bit-identical fits).

  ## Validation (diagnostic/scan, 98 cases)

  Cartesian-LM via picker is **400-1000× faster** on close-encounter cases
  and **strictly more accurate** on cases where Gauss-IOD's `solns[0]` was wrong:

  | case | pre-picker dr | picker dr | change |
  |---|---:|---:|---:|
  | `classical_42AU_arc_002.00d` | 1.57e10 km | 7.22e8 km | **−95%** |
  | `centaur_25AU_arc_002.00d`   | 2.38e9 km  | 3.68e8 km | **−85%** |

  Test suite went from ~9 min to ~3 s wallclock as a side effect.

  ## Dependencies
  - Stacks on `feat/pybind11-eigen-includes` (PR #N). Once that lands,
    this PR's diff collapses to its 4 new commits.

  - [x] Have you written a unit test for any new functions? — `tests/layup/test_iod_picker.py` (8 tests)
  - [x] Does Layup run successfully on a test set of input files/databases? — full `diagnostic/scan`

### DIFF
--- a/src/layup/iod.py
+++ b/src/layup/iod.py
@@ -32,6 +32,7 @@ implementation through a class hierarchy.
 from __future__ import annotations
 
 import logging
+import math
 from typing import Callable, Optional, Sequence
 
 from layup.routines import FitResult, Observation, gauss
@@ -41,6 +42,24 @@ logger = logging.getLogger(__name__)
 GMtotal = 0.0002963092748799319
 AU_M = 149597870700
 SPEED_OF_LIGHT = 2.99792458e8 * 86400.0 / AU_M
+
+# Default bounds for the cheap physical-feasibility filter. We
+# deliberately *don't* check bound-orbit energy here: Gauss's velocity
+# component can be wildly wrong (5-10× circular) for a seed whose
+# *position* is correct, and LM walks those into convergence reliably.
+# Throwing them out would silently kill the right root in cases like
+# the 41 AU classical KBO where Gauss returns a hyperbolic-looking
+# velocity but the right geometry.
+_MIN_R_AU = 0.05      # interior to Mercury — almost certainly unphysical
+_MAX_R_AU = 1000.0    # well past the Kuiper-belt range we typically care about
+
+# Geocentric distance below which we treat a candidate as a close-Earth-
+# approach risk: full ASSIST integration will get stuck on the close
+# encounter (and a 2-body workaround would silently mishandle the real
+# physics of NEO close passes — those exist and need their own solution
+# eventually). For now the prefilter just passes such candidates through
+# to LM unchanged; LM may also be slow on them, which is a known issue.
+_CLOSE_EARTH_AU = 0.1
 
 
 # An IOD method takes (observations, seq) and returns either a list of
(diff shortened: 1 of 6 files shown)


hannorein (commented):
Setting the min_dt value is fine. I don't see any problem with that. But it might also help to use the new IAS15 timestepping algorithm by setting `adaptive_mode=2`. Currently ASSIST [still uses the old one](https://github.com/matthewholman/assist/blob/7c367749cf5776847dc355a60ef2b750da20e7ee/src/assist.c#L446). I'd suggest trying if this alone already helps.

matthewholman:
That's a very good point, @hannorein.  I think we have set the rebound requirements to be an earlier version.  Is the new time step algorithm in that version?

hannorein:
Yes, it's in that version.

matthewholman:
@hannorein, you were right about adaptive_mode=2.  It's faster in all the tested cases and is as accurate as the fixed minimum time step.

hannorein:
@matthewholman Glad to hear! By the way, I did a test to see if I can easily speed up IAS15 by removing compensated summation. But it only gives a 20% speedup. And for a standard solar system integration, the errors are orders of magnitude worse, so I don't think it's worth it.

matthewholman:
That's an interesting result!  I would have thought removing the compensated summation would yield a big speed up.  It's definitely worth keeping it in.

hannorein (approved):

Issue #302, opened 2026-01-07
os.fork() was called. os.fork() is incompatible with multithreaded code,


mschwamb:
Showing up in the linux runners as a warning:

  /opt/hostedtoolcache/Python/3.11.14/x64/lib/python3.11/multiprocessing/popen_fork.py:66: RuntimeWarning: os.fork() was called. os.fork() is incompatible with multithreaded code, and JAX is multithreaded, so this will likely lead to a deadlock.
    self.pid = os.fork()

matthewholman:
Fixed by #336, which pins the ProcessPoolExecutor workers to the spawn start method (same root cause as #256). No
more os.fork() in the multithreaded/JAX path. Closing.

PR #328, opened 2026-05-17
Add BK 5-parameter linear IOD (run_bk_iod) + complementary-IOD sweep

matthewholman:
implementation I had started.

Adds a **universal-BK 5-parameter linear initial orbit determination**
(`run_bk_iod`) as the layup-side analog of liborbfit's `prelim_fit`,
plus an engine-sweep tool that quantifies how BK-IOD compares to
Gauss on the diagnostic-scan dataset.

## What

A single closed-form weighted least-squares solve over `(α, β, γ,
α̇, β̇)` with `γ̇` pinned to 0.  Per-observation model (small-angle
no-gravity, gnomonic tangent-plane projection, light-time corrected):

```
t_i    = (jd_i − (r_obs_i · n̂₀) / c) − epoch
X_i    = r_obs_i · â,    Y_i    = r_obs_i · b̂
x_obs  = (ρ̂_i · â) / (ρ̂_i · n̂₀)
y_obs  = (ρ̂_i · b̂) / (ρ̂_i · n̂₀)

x_obs = α + α̇·t_i − γ·X_i
y_obs = β + β̇·t_i − γ·Y_i
```

`γ̇`'s covariance is filled from the bound-orbit energy prior
(`σ_γ̇²` from `bk_basis::sigma_gdot_sq`).  6×6 BK covariance is
transformed to Cartesian via the analytic `dcart_dbk` chain rule.
Returned as a standard `FitResult` with `method="bk_iod"`.

## Headline result: BK-IOD and Gauss are complementary

Running `tools/bk_iod_sweep.py` on the diagnostic-scan dataset (98
cases, σ_obs = 0.1″, BK LM applied after each IOD as the seed):

| Outcome | Count |
|---|---|
| Gauss→LM converges, BK→LM also converges | 64 |
| Only BK→LM converges | 8 |
| Only Gauss→LM converges | 18 |
| Both fail | 8 (all 0.04-day arcs, 3 obs — data limit) |

So a pipeline that tries Gauss and falls back to BK covers
**90/98 cases**, vs Gauss alone (82) or BK alone (72).

Across the 64 cases where both pipelines converge:
- Drift ratio (BK_LM / Gauss_LM): **median 0.26, mean 0.56** (BK
  seeds give ~4× tighter LM convergence on the typical case)
- Iter ratio (BK_LM / Gauss_LM): **median 0.68, mean 0.85** (BK
  seeds need ~32% fewer LM iterations)

Per-population breakdown (BK win = both converge and BK closer to truth):

```
Population              n   G->LM ok  B->LM ok  BK win  G win  BK only  G only
centaur_15AU           14         12        12      11      1        0       0
centaur_25AU           14         12        12      10      1        1       1
classical_42AU         14         11        13       8      3        2       0
mainbelt_2.5AU         14         12         6       4      1        1       7
mainbelt_3.5AU         14         13         5       4      1        0       8
scattered_70AU         14         10        12       7      2        3       1
sednoid_80AU           14         12        12      10      1        1       1
TOTAL                  98         82        72      54     10        8      18
```

Pattern: **BK-IOD dominates Gauss for distant objects** (centaur,
classical, scattered, sednoid: 46 BK wins vs 8 Gauss wins).  Gauss
wins for mainbelt where BK-IOD's small-`γ·ze` assumption breaks.

## Documented regime of validity

`bk_iod.cpp` carries an explicit "REGIME OF VALIDITY" block in its
docstring with an empirical sweet-spot table across arc lengths.
BK-IOD is sub-percent on heliocentric distance for TNOs at well-
chosen arcs (5-30 days for 40-80 AU objects), but fails outside its
regime — mainbelt at >1 day gives catastrophic answers.  The model
omits the perspective factor `1/(1 − γ·ze)`; that's a feature of
linear IODs (same as `prelim_fit`), not a bug, and is intended use
is to **seed the BK LM**, not produce a final orbit.

## Tests

`tests/layup/test_bk_iod.py` covers (10 tests, all passing):

  * Smoke: empty / few-obs guards return without crashing.
  * `test_bk_iod_distant_objects`: BK-IOD recovers truth to within
    a few percent of `r_helio` on 4 representative distant cases
    (sweet-spot tolerances per the regime table).
  * `test_bk_iod_seeds_lm_to_truth`: the actual intended use —
    BK-IOD's output, fed into `run_bk_native_fit`, converges to
    within 1% of truth on all 4 cases, even when the IOD itself is
    several percent off.  LM convergence basins are wider than
    BK-IOD's bias.

## New files

```
src/lib/orbit_fit/bk_iod.cpp   ~190 lines, inlined into namespace orbit_fit
src/lib/orbit_fit/orbit_fit.cpp  one-line include
src/main.cpp                  bk_iod_bindings registration
tests/layup/test_bk_iod.py    10 tests
tools/bk_iod_sweep.py         engine-sweep harness, produces the table above
```

## Dependencies

Stacks on `feat/bk-everywhere` (PR #326 — provides the bk_basis
primitives, the BK LM driver that BK-IOD seeds, and the dispatch
infrastructure).  Once #326 lands, this PR collapses to its 3 new
commits.

- [x] Have you written a unit test for any new functions? — `tests/layup/test_bk_iod.py`, 10 tests
- [x] Do all the units tests run successfully? — full suite green
- [x] Does Layup run successfully on a test set of input files/databases? — full diagnostic-scan sweep ran cleanly


src/lib/orbit_fit/bk_basis.cpp
```diff
@@ -0,0 +1,294 @@
...
+#include <pybind11/stl.h>
+
+namespace orbit_fit
+{
+
+    namespace
+    {
+        // Internal cached quantities at the BK position (alpha, beta).
+        //
+        //   p = n0 + alpha*a + beta*b
```
kjnapier: Notation can be cleaned up in this comment

src/lib/orbit_fit/bk_basis.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+    Eigen::Matrix<double, 6, 1> bk_to_cartesian(
+        const BKState &bk, const BKFiducial &fid)
+    {
+        const RhoFrame f = compute_rho_frame(bk.alpha, bk.beta, fid);
+        const double inv_g = 1.0 / bk.gamma;
+        const double inv_g2 = inv_g * inv_g;
+
+        const Eigen::Vector3d r = inv_g * f.rho_hat;
+        const Eigen::Vector3d v = inv_g * (bk.adot * f.rho_hat_alpha + bk.bdot * f.rho_hat_beta)
```
kjnapier, line 119: Double check the math. The units seem funky, but I think it's okay

kjnapier (changes requested):
I restored the comments. New comment is that this is a lot of data files to store in the repo

matthewholman:
📦 Part of the **BK orbit-fit stack** — full overview & suggested review order are in #326's description.

This PR **stacks on #326** (its diff vs `main` includes #326's commits), and #331 stacks on this one. Review order: **#326 → #328 → #331**, reviewing each PR's delta.

matthewholman:
Heads up on CI for the latest commit (`da02505`): 5 of 6 checks are green — including **`build (ubuntu-latest, 3.13)`** and both macOS legs — but **`build (ubuntu-latest, 3.14)`** has been stuck `in_progress` for ~30+ minutes without finishing.

This looks like the known ubuntu-runner hang rather than anything in this PR's content:
- #256 — ubuntu GitHub runner hangs on `test_orbit_fit_cli` (`num_workers > 1`)
- #302 — `os.fork()` + JAX multithreading `RuntimeWarning` on the Linux runners

The 3.13 ubuntu leg passing on the same commit is good evidence the code itself is fine; the 3.14 leg appears to be hitting the runner hang. Suggest cancelling/re-running that single leg (or relying on the 3.13 result) rather than treating it as a failure of this PR.

matthewholman:
Trimmed the in-repo test data (commit `c7cb159`).

The consolidated `bk_scan_truth.json` carried all 98 diagnostic-scan cases (135 KB), but the pytest suite only references **9** of them by name. The full 98-case sweep lives in `sweep_cases_from_diagnostic()`, which is explicitly *not* invoked by pytest collection, and `tools/bk_iod_sweep.py` reads the external scan directory (`--scan-dir`) rather than this fixture — so nothing in CI needs the other 89 cases.

Kept exactly the 9 referenced cases (mainbelt / classical / scattered / sednoid across the arc lengths the tests assert on): **135 KB → 11 KB**, with no change in coverage (the 53 BK tests still pass, and load a bit faster). The previous round already collapsed the original 98 separate files into this single minified file; this cut removes the unused cases from it.

Note on the stack: this trim is on `feat/bk-iod`. The same fixture is inherited from `feat/bk-everywhere` (#326), so when the BK stack is rebased it'll carry the trimmed version up; I kept it as a direct commit here (rather than rebasing now) to avoid colliding with the `test_bk_basis.py` velocity-units test just added to #326.

matthewholman:
@kjnapier, the data files have been trimmed and comments have been added.

matthewholman:
@kjnapier rebased onto current `main`. The data footprint is now just the trimmed `bk_scan_truth.json` fixture, which already lives on `main` via #326 — no other data files. The `bk_basis.cpp` comments (notation + velocity units) were resolved in #326 (notation documented, units verified by the FD test), so that file is no longer in this diff. Could you take another look?

kjnapier (approved):
Okay, looks good to me

PR #329, opened 2026-05-17
Fix weight_data=True: NameError + arcsec/radian unit bug

matthewholman:
implementation I had started.

Fixes two distinct bugs that have made `orbitfit(weight_data=True)`
non-functional since the feature was added.  Either bug alone is
sufficient to break the path; both must be fixed for it to do
anything sensible.

## Bug 1: `NameError: name 'astcat_column_present' is not defined`

`_orbitfit` in `src/layup/orbitfit.py` defined the flag as

```python
g_column_present = "astCat" in column_names
```

but then referenced `astcat_column_present` two lines later inside
the debias and weighting branches.  As soon as `weight_data=True`
(or debias) was passed, this raised `NameError` before any fitting
could happen.

**Fix**: rename the definition to match the usage sites
(`astcat_column_present`).  One-line change.

## Bug 2: arcsecond vs radian unit mismatch

`data_weight_Veres2017` returns the astrometric uncertainty in
**arcseconds** (per its docstring).  The call site assigned the
return value directly to `Observation.ra_unc` and `dec_unc`, which
are stored in **radians**.  Each observation's weight was therefore
~(180·3600/π)² ≈ 4×10¹⁰× too loose -- inflating the fit covariance
to absurd values (sigma_pos ~25 million km on a 3 AU object) and
driving chi-square/dof to zero (residuals dwarfed by the assumed
noise floor).

**Fix**: convert arcsec → radian at the assignment, with a comment
explaining why.

## Effect

Layup's reported position uncertainty for three real MPC objects
(mainbelt asteroids at ~2-3 AU, from the existing test fixture):

| Object   | `weight_data=False` | `weight_data=True` (after fix) |
|---:|---:|---:|
| 119839   | 269 km              | **123 km** |
| 742428   | 340 km              | **160 km** |
| 609631   | 306 km              |  **80 km** |

With the fix:
- sigma_pos drops by ~half (closer to JPL's published per-element
  sigmas for these objects)
- chi-square/dof moves from artificially low (0.05-0.23, suggesting
  the default 1-arcsec uncertainty over-weights the data) to a more
  honest 0.27-0.53

## Tests

`tests/layup/test_weight_data.py` (2 tests):

  - `test_orbitfit_weight_data_true_runs_without_crashing`:
    guards against the NameError-class regression -- exercises the
    `astcat_column_present` branch by calling
    `orbitfit(..., weight_data=True)` on real MPC data with no
    `astCat` column.
  - `test_orbitfit_weight_data_true_gives_sensible_uncertainty`:
    asserts `sigma_pos < 10000 km` for the 27-year mainbelt arc.
    Well above the actual ~123 km, but trivially below the
    25-million-km failure mode of the units bug.

## Dependencies

Independent of all other open PRs -- branches off `main`, touches
only `src/layup/orbitfit.py` and adds one test file.

- [x] Have you written a unit test for any new functions? — `tests/layup/test_weight_data.py`
- [x] Do all the units tests run successfully? — both new tests + existing suite
- [x] Does Layup run successfully on a test set of input files/databases? — real-MPC fixture (`4_random_mpc_ADES_provIDs_no_sats.csv`)

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -516,7 +516,7 @@ def _orbitfit(
 
         # Check if certain columns are present in the data
         column_names = data.dtype.names
-        g_column_present = "astCat" in column_names
+        astcat_column_present = "astCat" in column_names
         program_column_present = "program" in column_names
         position_rates_columns_present = all(col in column_names for col in ["raRate", "decRate"])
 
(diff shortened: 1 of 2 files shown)


Little-Ryugu (approved):
LGTM

matthewholman:
Grouped with the **BK orbit-fit stack** for context (overview in #326), but an **independent** bugfix with no dependency on the BK PRs.

PR #330, opened 2026-05-17
Add real-data validation against JPL Horizons

matthewholman:
implementation I had started.

Adds an end-to-end validation test that runs `orbitfit` on real MPC
astrometry and asserts the recovered Cartesian state matches the
JPL Horizons reference at the same epoch.  Every other test in the
layup suite uses synthetic observations (either generated via
`predict_sequence` or from the diagnostic-scan dataset built by
integrating known orbits with ASSIST) -- this is the first
real-world ground-truth check.

## What

Three numbered asteroids from the existing `4_random_mpc_ADES_provIDs_no_sats.csv`
test fixture, each compared to JPL Horizons:

| provID   | name      | obs  | arc       | layup vs JPL drift |
|---|---|---:|---|---:|
| 119839   | 2002 CX17 | 587  | 27 years  | ~150 km            |
| 742428   | 2007 TC75 | 117  | shorter   | ~130 km            |
| 609631   | 2005 HE12 | 109  | shorter   |  ~20 km            |

All three objects are mainbelt at ~2-3 AU (i.e., ~3 × 10⁸ km), so
relative position agreement is 1×10⁻⁷ to 1×10⁻⁶.  Velocity agreement
is 1×10⁻⁹ AU/day or better.  Test tolerances (1×10⁻⁵ AU pos,
1×10⁻⁷ AU/day vel) sit comfortably above the observed accuracy so
benign numerical drift won't flag false positives.

## How

`tests/data/jpl_reference_states.json` -- captured once via JPL
Horizons' public API on 2026-05-17 (see the `_comment_` field for
regeneration instructions: query
`https://ssd.jpl.nasa.gov/api/horizons.api` with
`CENTER='@0'`, `REF_PLANE='FRAME'`, `REF_SYSTEM='ICRF'`,
`VEC_TABLE='2'`, `OUT_UNITS='AU-D'` at the layup-recovered epoch).

The test runs hermetically against this committed fixture -- no
network access during CI.

It also catches future drift in layup's epoch-selection logic: the
test asserts that layup's chosen epoch matches the fixture's
recorded epoch to 10⁻⁹ days.  If layup changes how it picks the fit
epoch for a given observation set, the fixture needs regeneration
(the test message says so).

## Independent

Doesn't depend on any of the other open BK-related PRs -- branches
off `main`, touches only `tests/`.

- [x] Have you written a unit test for any new functions? — `tests/layup/test_real_data_validation.py`
- [x] Do all the units tests run successfully? — 3 parametrized cases, all green
- [x] Does Layup run successfully on a test set of input files/databases? — uses the existing real-MPC fixture

### DIFF
--- a/tests/data/jpl_reference_states.json
+++ b/tests/data/jpl_reference_states.json
@@ -1,0 +1,41 @@
+{
+  "_comment_": "JPL Horizons reference state vectors for objects in 4_random_mpc_ADES_provIDs_no_sats.csv. Queried 2026-05-17 via the public Horizons API at the layup-recovered epochs (epoch values match exactly so the comparison is at one definite epoch). Frame: barycentric, ICRF (== layup's BCART_EQ default).  To regenerate: run orbitfit on each provID to get its epoch, then query https://ssd.jpl.nasa.gov/api/horizons.api with COMMAND=<provID>, CENTER='@0', REF_PLANE='FRAME', REF_SYSTEM='ICRF', VEC_TABLE='2', OUT_UNITS='AU-D'.",
+  "objects": {
+    "119839": {
+      "_name_": "2002 CX17",
+      "epoch_jd_tdb": 2459546.75963587081,
+      "state_au_au_per_day": [
+        1.344744268651191,
+        2.442670284100593,
+        1.527065730993102,
+        -8.398959551991471e-03,
+        3.798491654270669e-03,
+        1.566943188232538e-03
+      ]
+    },
+    "742428": {
+      "_name_": "2007 TC75",
+      "epoch_jd_tdb": 2459533.7747232849,
+      "state_au_au_per_day": [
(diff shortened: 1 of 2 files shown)


hannorein (commented):
Looks good. Just one question: if layup recovers each tested object's barycentric position to a relative agreement of ~1e-7 why set the test threshold to 1e-5 and not something closer to 1e-7?

matthewholman:
Good catch, @hannorein.  The value of 1e-5 was arbitrary.  We can change that to something closer to 1e-7.

matthewholman:
@hannorein Good point — I reframed the tests to assert a **relative** tolerance (drift / |reference|).  So the real relative agreement is ~4e-7 (position) / ~6e-7 (velocity).

hannorein (approved):

PR #331, opened 2026-05-17
Add iod='auto': Gauss with BK-IOD fallback

matthewholman:
implementation I had started.

Adds `iod="auto"` to `do_fit`: runs Gauss IOD first, falls back to
the universal-BK 5-parameter linear IOD (`run_bk_iod`) if every
Gauss root fails to seed the LM.  `iod="gauss"` (the default) is
unchanged.

## Motivation

The diagnostic-scan sweep in `tools/bk_iod_sweep.py` (added in
PR #328) shows that **Gauss and BK-IOD are complementary**.
Across 98 cases, the IOD→LM pipelines succeed on:

| IOD            | converges |
|---|---:|
| Gauss alone    | 82 / 98   |
| BK-IOD alone   | 72 / 98   |
| **Either**     | **90 / 98** |

So a single "try Gauss, fall back to BK" path picks up an extra
~8 cases in this dataset.  The regimes where BK rescues Gauss are
mostly short-arc distant objects -- exactly where BK-IOD's
parallax-based linear fit handles geometries that Gauss's
three-point method finds degenerate.

## What changed

- `do_fit`'s primary-segment seed loop is refactored from the
  previous "try root 0, elif root 1, elif root 2" chain (which had
  a logic bug -- the second `elif` could never fire) into a clean
  for-loop over all Gauss roots that breaks on first LM convergence.
- After the Gauss-roots loop, if `iod='auto'` and no Gauss root
  succeeded, `do_fit` calls `run_bk_iod` on the primary segment
  (using the middle observation's epoch, matching Gauss's
  `idx1` convention) and re-runs the LM with that seed.
- `iod='gauss'` still raises `ValueError` on any unknown value;
  `iod='auto'` is now also accepted.

## Tests

`tests/layup/test_iod_auto.py` (4 tests):

  - `test_do_fit_unknown_iod_raises`: ValueError on typo'd iod.
  - `test_iod_auto_matches_gauss_on_easy_case`: on a mainbelt arc
    where Gauss converges, `iod='auto'` produces a state
    bit-identical to `iod='gauss'` (the BK fallback never fires).
  - `test_iod_auto_recovers_when_gauss_fails`: on
    `sednoid_80AU_arc_001.00d`, `iod='gauss'` returns `flag=3`
    (primary interval failed) under the default Cartesian LM,
    while `iod='auto'` returns `flag=0` by falling back to BK-IOD.
  - `test_iod_auto_engine_choice_propagates`: `iod='auto'` is
    independent of the engine choice (works with both
    `engine='cartesian'` and `engine='bk_native'`).

## Dependencies

Stacks on `feat/bk-iod` (PR #328 -- provides `run_bk_iod`).  Once
PR #328 lands, this PR collapses to its 1 new commit.

- [x] Have you written a unit test for any new functions? — `tests/layup/test_iod_auto.py`
- [x] Do all the units tests run successfully? — 4 new tests + 56 BK + IOD tests

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -15,6 +15,7 @@ from layup.routines import (
     Observation,
     gauss,
     get_ephem,
+    run_bk_iod,
     run_bk_native_fit,
     run_from_vector_with_initial_guess,
 )
@@ -472,7 +473,9 @@ def do_fit(
     cache_dir : str
         Directory holding the ASSIST kernels.
     iod : str
-        Name of the registered IOD method. Default "gauss".
+        Name of the registered IOD method (default "gauss"), or "auto" to run
+        Gauss and fall back to the BK 5-parameter linear IOD (run_bk_iod) on the
+        primary segment when every Gauss root fails to seed a converged fit.
     engine : str
         Which LM fitter to dispatch to.  Supported:
           - 'cartesian' (default): the existing 6D Cartesian-state fit.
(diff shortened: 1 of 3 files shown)


Little-Ryugu (commented):
The code and logic look good.

Just two things:

1. Should we change the iod to default to 'auto' instead of guass now for orbitfit?

https://github.com/Smithsonian/layup/blob/a0b4d2e79573057a7c5e99eb9fcb70b4215e11b8/src/layup_cmdline/orbitfit.py#L73-L80

2. unit tests are being skipped on both github and my laptop. It would be better to have them availble somewhere accessible for github and other devs for future pull requests.  
https://github.com/Smithsonian/layup/blob/8acd7347486055bdcc869ab14c342829bb4385ec/tests/layup/test_bk_everywhere.py#L1-L13

unit tests are being skipped in test_bk_everywhere.py, test_bk_iod.py and test_iod_auto.py

matthewholman:
I agree that 'auto' should be the new default.  I'll get that changed.

Good catch regarding the skipped unit tests.  I'll address that too.

matthewholman:
📦 **Top of the BK orbit-fit stack** — full overview in #326's description.

This PR **stacks on #328 (→ #326)**, so its diff vs `main` includes both parents; review after them. Note: the `iod='auto'` recovery test is gated behind `LAYUP_SLOW_TESTS` (a real LM fit on a degenerate distant orbit can grind the ASSIST integrator on Linux); default CI keeps the two fast `iod='auto'` tests.

Little-Ryugu (approved):
LGTM. All unit tests work now except the one that is skipped, but it can be run locally.

PR #334, opened 2026-06-20
Fix NaN KEP/BKEP covariances for hyperbolic orbits (#288)

### LINKED ISSUE #288: converting orbits with e>1 to KEP gives null values for covariances.
When converting orbits with e > 1 to KEP using `layup convert`, the covariance matrix is not converted and is output as null values. Should we fix this or give a warning/error to the user?

matthewholman:
## Summary
Converting an orbit with `e>1` to **KEP** or **BKEP** produced an all-NaN covariance matrix (issue #288).

`universal_keplerian` computed the mean motion as `sqrt(mu / a**3)`. For a hyperbolic orbit `a = q/(1-e) < 0`, so this is the square root of a negative number → `NaN`, which flows through the mean anomaly `M` into `jac_keplerian_xyz` and poisons the KEP/BKEP covariance transform (both share `covariance_keplerian_xyz`).

The hyperbolic mean anomaly is well-defined with a **real** mean motion `n = sqrt(mu / |a|**3)` (= `sqrt(-mu/a**3)` — the same form already used in `hypanom`), with `M = n * (t - tp)`.

## Changes
- `universal_keplerian`: `M = (epoch - tp) * sqrt(mu / |a|**3)`
- `covariance_xyz_keplerian`: `tp = epoch - M * sqrt(|a|**3 / mu)`

For elliptic orbits `|a| = a`, so results are unchanged.

## Verification
- Numerically confirmed the hyperbolic covariance is now finite **and** agrees with a finite-difference propagation of the input covariance to ~4e-7 (elliptic agrees to the same tolerance).
- Added a regression test (`test_keplerian_covariance_finite_for_hyperbolic_orbit`) that is NaN without the fix.
- `test_convert.py`: 10 passed on a fresh build.

Closes #288.


kjnapier (approved):
Yes, the new math looks good

PR #335, opened 2026-06-20
Warn when an orbit fit fails on a sub-24h arc (#312)

matthewholman:
## Summary
Orbit determination needs an observational baseline longer than ~24 hours. When a fit failed on a single night of data, the user previously got an opaque failure (issue #312).

On a failed fit, `_orbitfit` now checks the observation arc span and, if it is under a day, logs a clear warning naming the object and the (too-short) arc length so the likely cause is obvious.

## Changes
- New `_warn_if_short_arc(jds, obj_id)` helper (threshold `_MIN_ARC_DAYS = 1.0` day), called where the fit result is known to have failed.
- Unit tests for the sub-day (warns), multi-day (silent), and empty (no crash) cases.

## Verification
- `test_orbit_fit.py`: 9 passed on a fresh build, including the 3 new tests.

Closes #312.


kjnapier (approved):
Looks good

PR #255, opened 2025-05-15
For test_orbit_fit_cli set num_workers to 1

wilsonbb:
Fixes a test error observed since https://github.com/Smithsonian/layup/pull/241 where the ubuntu github runner hangs on test_orbit_fit_cli

https://github.com/Smithsonian/layup/issues/256 is filed to follow-up on debugging and fixing the underlying issue. However here we simply make the tests usable again by setting num_workers=1

Describe your changes.

## Review Checklist for Source Code Changes

- [ ] Does pip install still work?
- [ ] Have you written a unit test for any new functions?
- [ ] Do all the units tests run successfully?
- [ ] Does Layup run successfully on a test set of input files/databases?
- [ ] Have you used black on the files you have updated to confirm python programming style guide enforcement?

PR #336, opened 2026-06-20
Fix Linux fork+JAX worker deadlock by using the 'spawn' start method (#256)

### LINKED ISSUE #256: Fix ubuntu github runner hanging for test_orbit_fit_cli
A test error observed since https://github.com/Smithsonian/layup/pull/241 is that the ubuntu github runner hangs on test_orbit_fit_cli when n_workers > 1

Example https://github.com/Smithsonian/layup/actions/runs/15050986557/job/42305483912

This PR introduced calls to convert into the orbit fit workflow. However in test_convert we are currently only testing `num_workers=1`

I suspect that increasing the number of workers in test_convert will produce the same error which will simplify debugging. Otherwise it might be an issue between calling orbit fit and convert

matthewholman:
## Summary
The ubuntu CI runner intermittently **hangs** in orbit-fit / convert workflows with `num_workers > 1` (issue #256). This is the classic *fork-in-a-multithreaded-process* deadlock.

`ProcessPoolExecutor` was created with the platform-default start method, which on **Linux is `fork`**. `fork()` clones the whole parent process — including any mutex another thread holds at the instant of the fork, in its **locked** state. `orbitfit` imports JAX (via `layup.convert` → `orbit_conversion`) at module load, and JAX/XLA runs background threads, so a forked worker can inherit a permanently-locked JAX mutex and deadlock the first time it touches it. CPython's *"os.fork() … JAX is multithreaded … will likely lead to a deadlock"* `RuntimeWarning` (#302) is the same root cause.

macOS already defaults to **`spawn`**, which is exactly why the hang was Linux-only.

## Changes
- Pin both `ProcessPoolExecutor` sites in `data_processing_utilities.py` to a **`spawn`** context, so every platform launches a fresh interpreter per worker — inheriting none of the parent's locks or threads, so the deadlock cannot occur. (`forkserver` would also work and is cheaper, since it forks workers from a clean single-threaded helper; `spawn` is the most portable and matches the macOS default. Easy to switch if the team prefers `forkserver`.)
- Revert #255's test-only bandaid (`num_workers` `1 → 2` in `test_orbit_fit_cli`) so the multi-worker fit→convert path is exercised again.

## Verification
On a fresh build:
- `test_orbit_fit_cli[num_workers=2]` (the reverted #255 case, i.e. the exact #256 scenario) — passes.
- `test_data_processing_utilities` with `n_workers ∈ {1,2,4,8}` — 39 passed.
- In-process predict tests — pass.

Note: macOS already used `spawn`, so local macOS runs confirm the explicit context doesn't break the multi-worker paths; the Linux deadlock resolution itself is validated by the ubuntu CI legs on this PR. (Corroborating data point: on a recent PR the same commit hung on `ubuntu-3.14` once and passed on re-run — the nondeterministic race this change eliminates.)

Closes #256. Related: #302 (same root cause), #255 (bandaid reverted here).

### DIFF
--- a/src/layup/utilities/data_processing_utilities.py
+++ b/src/layup/utilities/data_processing_utilities.py
@@ -1,4 +1,5 @@
 import logging
+import multiprocessing
 from concurrent.futures import ProcessPoolExecutor
 
 import numpy as np
@@ -10,6 +11,23 @@ from layup.routines import FitResult
 from layup.utilities.layup_configs import LayupConfigs
 
 """ A module for utilities useful for processing data in structured numpy arrays """
+
+# Start worker processes with "spawn" rather than the platform default.
+#
+# On Linux the default start method is "fork", which clones the whole parent
+# process -- including any mutexes other threads hold at the instant of the
+# fork, in their *locked* state.  layup imports JAX (via layup.convert ->
+# orbit_conversion) at module load, and JAX/XLA runs background threads, so a
+# forked worker can inherit a permanently-locked JAX mutex and deadlock the
+# first time it touches it.  This is the cause of the ubuntu CI hangs in
+# orbit-fit/convert workflows with num_workers > 1 (see issues #256 and #302).
+#
+# "spawn" launches a fresh interpreter per worker, inheriting none of the
+# parent's locks or threads, so the deadlock cannot occur.  macOS already
+# defaults to "spawn" (which is why the hang was Linux-only); pinning it here
+# makes every platform behave the same.  ("forkserver" would also work and is
+# cheaper, but "spawn" is the most portable and matches the macOS default.)
+_MP_CONTEXT = multiprocessing.get_context("spawn")
 
 AU_M = 149597870700
 AU_KM = AU_M / 1000.0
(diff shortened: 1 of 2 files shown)


mschwamb:
I thought this isn't true in the new way that v3.13 and onward does its multithreading as this has changed in v3.13 onward

mschwamb:
The ubuntu v3.13 CI is hung so I'm not sure this has fixed the issue.

matthewholman:
Thanks @mschwamb — I think CI is showing something more specific than a hang:

**"3.13 is hung":** it actually ran to completion (it's a `failure`, not a hang). 23 of its 25 failures are `requests.ConnectTimeout` to `minorplanetcenter.net` — that runner couldn't reach the MPC to download obscodes; the long runtime is timeout×retries, not a deadlock. That's a transient upstream outage (also hit `naif.jpl.nasa.gov` on #334/#335), addressed by #337 (bundled-obscodes fallback) and #338 (cache the data in CI).

**The 3.13+ threading change:** this PR sidesteps it by not forking at all — with `spawn`, workers don't call `os.fork()`, and the `os.fork()` warning from #302 is gone from these runs (0 occurrences). It also covers the versions we still support that *do* default to fork (3.11–3.13, per `requires-python`). Empirically, `ubuntu-3.14` — the leg that hung on #328 — passes here in ~12 min.

mschwamb (approved):

mschwamb:
This is approved @matthewholman  do you want to merge?

matthewholman:
Thanks, @mschwamb!  Let me check if this PR depends on merging earlier PRs.

matthewholman:
It doesn't.  I'll merge it.

PR #337, opened 2026-06-20
Fall back to a bundled obscodes file when the MPC download fails

matthewholman:
## Summary
`LayupObservatory` downloads the MPC observatory-codes file (`obscodes_extended.json.gz`) from `minorplanetcenter.net` on first use. When that server is unreachable the **whole run fails** — this recently took down a CI leg where every obscode-dependent test errored with `requests.exceptions.ConnectTimeout` (23 failures on one ubuntu runner).

Observatory codes change rarely, so this ships a fallback copy in the package and uses it when the live download fails.

## Changes
- Bundle `layup/data/ObsCodes.json.gz` (77 KB, 2708 codes).
- `write_fallback_obscodes()`: decompress the bundled file to a plain JSON path.
- `LayupObservatory.__init__`: try the normal MPC download; on a `requests.exceptions.RequestException`, log a warning and re-initialize via sorcha's `Observatory(..., oc_file=...)` argument, which reads the decompressed JSON directly and bypasses the download.

A slightly stale fallback is far better than a hard failure on a transient outage.

## Verification (fresh build)
- `test_write_fallback_obscodes` — the bundled file decompresses to valid JSON (X05/500 present).
- `test_layup_observatory_falls_back_on_mpc_failure` — a simulated `ConnectTimeout` falls back to the bundled copy and populates `ObservatoryXYZ`.
- Full `tests/layup/test_data_processing_utilities.py`: **41 passed**, no regressions.

## Notes
- This fixes the *failure* mode; it does not shorten the download timeout, so an outage still costs the (one-time per worker) pooch retry/timeout before falling back. A follow-up could shorten that timeout or have the fallback also seed the on-disk cache.
- Independent of the `spawn` fix in #336 (network resilience vs. the fork/JAX deadlock); both touch `data_processing_utilities.py` but in non-overlapping regions, so they rebase cleanly in either order.


kjnapier (approved):
Looks good

PR #339, opened 2026-06-20
Fix #286: handle observatories with zero-valued parallax constants

### LINKED ISSUE #286: observatory code coming up nans
Checking cache for existing files.
All expected files were found in the local cache: /Users/mschwamb/Library/Caches/layup/
2025-07-02 14:50:44,253 - layup.orbitfit - 71836 - INFO - Processing 110 rows for ['A11pl3Z']
Traceback (most recent call last):
  File "/Users/mschwamb/miniforge3/envs/layup/bin/layup-orbitfit", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/mschwamb/Research/python/lsst/layup/src/layup_cmdline/orbitfit.py", line 148, in main
    return execute(args)
           ^^^^^^^^^^^^^
  File "/Users/mschwamb/Research/python/lsst/layup/src/layup_cmdline/orbitfit.py", line 205, in execute
    orbitfit_cli(
  File "/Users/mschwamb/Research/python/lsst/layup/src/layup/orbitfit.py", line 830, in orbitfit_cli
    fit_orbits = orbitfit(
                 ^^^^^^^^^
  File "/Users/mschwamb/Research/python/lsst/layup/src/layup/orbitfit.py", line 666, in orbitfit
    pos_vel = layup_observatory.obscodes_to_barycentric(data)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mschwamb/Research/python/lsst/layup/src/layup/utilities/data_processing_utilities.py", line 449, in obscodes_to_barycentric
    obscode_cache_key = self.populate_observatory(obscode, et, row)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mschwamb/Research/python/lsst/layup/src/layup/utilities/data_processing_utilities.py", line 387, in populate_observatory
    raise ValueError(f"Observatory {obscode} has invalid coordinates at epoch {et}: {coords}")
ValueError: Observatory R17 has invalid coordinates at epoch 804687686.9920872: [nan nan nan]
Error: Command 'layup-orbitfit' failed with exit code 1.

Using files: 

[I3.80col.txt](https://github.com/user-attachments/files/21023851/I3.80col.txt)

matthewholman:
## Summary

Fixes #286. `layup orbitfit` crashed with `ValueError: Observatory <code> has invalid coordinates` for several real observatory codes — most importantly the common **geocentric code 500**, plus **000 (Greenwich)**, **782 (Quito)**, and **244/248**.

## Root cause

The crash originates in Sorcha's `Observatory.convert_to_geocentric` (`simulation_parsing.py`), which gates on the **truthiness** of the parallax constants:

```python
if (obs_location.get("Longitude", False)
    and obs_location.get("cos", False)
    and obs_location.get("sin", False)):
```

A parallax constant that is legitimately `0.0` is falsy, so the observatory is wrongly reported as having no fixed position. Layup then routes it to the moving-observatory path, which demands per-observation `pos1/2/3` fields and raises for plain MPC input that carries none. The affected codes all have a constant that is exactly `0.0`:

| code | name | zero constant |
|------|------|---------------|
| 500 / 244 / 248 | Geocentric / Occultation / Hipparcos | Longitude = cos = sin = 0 |
| 000 | Greenwich | Longitude = 0 |
| 782 | Quito (equatorial) | sin = 0 |

## Fix

Override `convert_to_geocentric` in `LayupObservatory` to test for the **presence** of the constants (`is not None`) rather than their truthiness. Since `super().__init__` calls `self.convert_to_geocentric`, the override is used when the observatory table is built — no upstream release needed.

- Codes with no position keys at all (roving observer 247, space telescopes WISE/TESS/HST) still return `(None, None, None)` and correctly take the ADES per-observation position path.
- The geocenter resolves to a `(0, 0, 0)` offset from Earth's center, which is exactly right.

The same bug should also be fixed upstream in Sorcha (a separate one-line PR); this override makes layup correct immediately and is robust regardless of the installed Sorcha version.

## Verification

- New regression test `test_fixed_observatory_with_zero_parallax_constant` — passes with the fix, **fails without it**.
- Full `tests/layup/test_data_processing_utilities.py`: **40 passed**.
- End-to-end: a geocentric `500` observation now flows through `obscodes_to_barycentric` and resolves to Earth's barycentric position; Greenwich `000` lands exactly one Earth radius (4.25e-5 AU) from the geocenter.

### DIFF
--- a/src/layup/utilities/data_processing_utilities.py
+++ b/src/layup/utilities/data_processing_utilities.py
@@ -320,6 +320,44 @@ class LayupObservatory(SorchaObservatory):
         # A cache of barycentric positions for observatories of the form {obscode: {et: (x, y, z)}}
         self.cached_obs = {}
 
+    def convert_to_geocentric(self, obs_location: dict) -> tuple:
+        """Convert an observatory's parallax constants to geocentric coordinates.
+
+        This overrides Sorcha's ``Observatory.convert_to_geocentric``, which gates
+        on the truthiness of the parallax constants
+        (``obs_location.get("sin", False)``). A constant that is legitimately
+        ``0.0`` -- e.g. the geocenter (codes 500/244/248: Longitude=cos=sin=0),
+        Greenwich (code 000: Longitude=0), or an equatorial station such as Quito
+        (code 782: sin=0) -- is falsy, so the base class wrongly reports the
+        observatory as having no fixed position. Layup then routes it to the
+        moving-observatory path and raises "invalid coordinates" for plain MPC
+        input that carries no per-observation position (issue #286).
+
+        We instead test for the *presence* of the constants. Codes that have no
+        position keys at all (roving observer 247, space telescopes WISE/TESS/HST,
+        ...) still return ``(None, None, None)`` and are correctly routed to the
+        ADES per-observation position path. The geocenter resolves to a (0, 0, 0)
+        offset from Earth's center, which is exactly right.
+
+        Parameters
+        ----------
+        obs_location : dict
+            Dictionary with Longitude and the sin/cos of the observatory latitude.
+
+        Returns
+        -------
+        tuple
+            Geocentric position (x, y, z), or (None, None, None) when the
+            observatory has no fixed position.
+        """
+        longitude = obs_location.get("Longitude")
+        cos = obs_location.get("cos")
+        sin = obs_location.get("sin")
+        if longitude is not None and cos is not None and sin is not None:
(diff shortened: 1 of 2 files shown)


kjnapier (approved):
The new gate logic looks good to me. I did not check the correctness of the old expressions, but assuming they did not change, all is good.

PR #340, opened 2026-06-20
Fix predict obs_cov_yy index + Monte-Carlo covariance validation (#278)

### LINKED ISSUE #278: Test observational covariance provided by predict
The idea is to test the observational covariance provided by predict by repeatedly adding noise to the observations, refitting, and predicting to the time of observation.  If the orbit covariance and the observational covariance are implemented correctly and are a good representation, the scatter in the predicted RA/Dec values should be consistent with the uncertainty ellipse from the observational covariance.

An example of this can be found in the scatter.c routine in B&K.

matthewholman:
## Summary

Closes #278. Adds the Monte-Carlo covariance validation requested in the issue (the `scatter.c` check from Bernstein & Khushalani), and fixes the `predict` covariance bug that the validation uncovered.

## The bug

`predict`'s per-observation result mapped the flattened 2×2 sky-plane covariance into the output struct with the wrong index. The C++ packs the 2×2 as:

```
obs_cov[0]=(0,0)  obs_cov[1]=(0,1)  obs_cov[2]=(1,0)  obs_cov[3]=(1,1)
```

but `_predict` read `obs_cov_yy` from `pred.obs_cov[2]` — the `(1,0)` **off-diagonal** term — instead of `pred.obs_cov[3]` = `(1,1)`. So the reported Dec variance was actually the RA/Dec covariance, which:

- under-estimated the minor-axis (≈Dec) uncertainty by ~10× in variance (~3× in sigma), and
- made `obs_cov_yy == obs_cov_xy` identically.

The 2×6 partials, the 6×6 orbit covariance, and the `B·cov·Bᵀ` projection are all correct — only the result index was wrong. The fix is one line.

## How it was found and verified

A three-way comparison localized the bug precisely:

| | σ_RA | σ_Dec |
|---|---|---|
| Empirical refit scatter (Monte Carlo) | 0.33″ | 0.28″ |
| Finite-difference Jacobian × fit cov | 0.29″ | 0.275″ |
| `predict` (before fix) | 0.29″ ✓ | **0.085″ ✗** |
| `predict` (after fix) | 0.29″ | **0.275″ ✓** |

The analytic `B` matrix matches a finite-difference Jacobian of predict's own forward model element-for-element, and the refit **state** scatter matches the fitted 6×6 covariance (`mean χ²₆ = 6.16`, expected 6) — so both the fit covariance and the projection partials are correct; the discrepancy was purely the output index.

## The test

`test_predict_observational_covariance_matches_refit_scatter` implements the issue's procedure: fit → predict to an epoch → perturb the observations by their assumed 1″ uncertainty → refit → re-predict, over 100 trials. The squared Mahalanobis distances of the predicted positions against the predicted covariance follow χ²(2), so their mean is ≈2.

- **With the fix:** `mean ≈ 2.3` ✅
- **Without the fix:** `mean ≈ 11–13` ❌ (the test fails)

It also validates that the fitted 6×6 state covariance is correct, and asserts `obs_cov_yy != obs_cov_xy`. Runs in ~9 s (fixed RNG seed; uses the in-process fit/predict workers so it stays fast). This lets the covariance comparison in `test_predict_output` (currently commented out "until the covariance calculation is solidified") be revisited in a follow-up.


matthewholman:
Follow-up commit (`b3778e3`): re-enabled the covariance verification in `test_predict_output` that was commented out "until the covariance calculation is solidified".

Root cause it was disabled: `holman.csv` had an all-zero input covariance, so `predict` emitted a zero sky-plane covariance — and the golden `obs_cov0–3` columns were externally sourced (next to the hand-added JPL columns) and not reproducible from a zero-covariance input. The stale column names (`obs_cov0` vs the actual `obs_cov_xx/yy/xy`) were a second problem.

Change:
- `holman.csv` now carries a realistic anisotropic diagonal covariance. Covariance does not affect ra/dec/rho, so the existing golden asserts are unchanged (verified). This also exercises the COM→BCART_EQ covariance conversion in the CLI path, which I confirmed produces a correct, well-conditioned 2×2 (`yy/xy ≈ 4`, positive determinant).
- Replaced the commented golden-value asserts with physical-property checks: finite, positive variances, positive-definite, and `obs_cov_yy != obs_cov_xy` (a direct regression guard for the bug fixed here). These are robust to the ~1e-7 platform variation in the propagated covariance, which has no independent reference in the repo. The full numerical validation lives in the new Monte-Carlo test.

Separately noted for a follow-up (not in this PR): `skyplane_cov_to_radec_cov` (the `a_arcsec/b_arcsec/PA_deg` ellipse helper) calls `np.cos(dec)` with `dec` in **degrees**, and appears to double-apply `cos(dec)` since `predict`'s `obs_cov` is already great-circle (projected onto the unit `a_vec`). That governs only the derived ellipse columns, not the raw `obs_cov_*`.

matthewholman:
Follow-up commit (`c10c9f8`): fixed the error-ellipse helper (`a_arcsec`/`b_arcsec`/`PA_deg`) that I flagged above. It had **three compounding bugs**:

1. **Call-site argument order** — `predict` called `skyplane_cov_to_radec_cov(ra, dec, obs_cov_xx, obs_cov_yy, obs_cov_xy)` but the signature was `(ra, dec, cov_xx, cov_xy, cov_yy)`, so the Dec variance and the RA/Dec off-diagonal were swapped.
2. **`cos(dec)` in degrees** — the function called `np.cos(dec)` with `dec` in degrees.
3. **`cos(dec)` shouldn't be there at all** — predict's sky-plane covariance is already on-sky (great-circle): its RA axis is the unit `a_vec`, which already carries the `cos(dec)` geometry. Re-applying `cos(dec)` double-counts.

The error ellipse is just the eigen-decomposition of the 2×2 sky-plane covariance. The helper now computes that directly (dropping the unused `ra`/`dec` args and the `cos(dec)` scaling, with a clamp against a tiny negative minor eigenvalue), and the call site is fixed.

Verified: the corrected output matches `numpy.linalg.eigh` exactly (a, b, and PA) across random matrices, and matches the Monte-Carlo empirical ellipse on the micro-arc fit (≈0.27–0.30″, vs the old code's ~146″ with NaN warnings). Added a unit test for the helper plus an eigen-decomposition consistency assert in `test_predict_output` (guards the argument-order regression). Both fail on the respective bug and pass with the fix.

PR #340 now covers the full covariance path: the `obs_cov_yy` index fix, the Monte-Carlo validation, the re-enabled `test_predict_output` covariance checks, and this error-ellipse fix.

kjnapier (approved):
Looks good

PR #341, opened 2026-06-21
Fix barycentric placement of space-based observatories + TNO validation (#149)

### LINKED ISSUE #149: read in and parse space-based observatory stack-and-shift observations
read in and parse space-based observatory stack-and-shift observations

matthewholman:
## Summary

Fixes the barycentric placement of **space-based / moving observatories** (relates to #149), and adds the TNO regime to the real-data validation suite (2000 FV53 vs JPL Horizons).

## The bug

A space-based observation carries the observatory's geocentric position per-observation (`sys`/`ctr`/`pos1..3`); `populate_observatory` stores it in **km, in the J2000/ICRF frame**. But `obscodes_to_barycentric` then fed that vector through `sorcha.barycentricObservatoryRates`, which is written for **fixed ground stations** — it rotates the stored vector from the Earth-fixed frame to J2000 and **scales it by the Earth radius**. Applied to an already-J2000, already-km position, that misplaces the observer by a factor of the Earth radius (~6378×):

| | barycentric distance from Earth |
|---|---|
| HST (actual, low-Earth orbit) | ~6,948 km |
| **layup (before fix)** | **~44,000,000 km** |

So **any fit including a space-based observation diverged**. Concretely, 2000 FV53 fits cleanly from its 27 ground observations but diverged (flag=4) the moment its single HST observation was included.

## The fix

Compute a moving observatory's barycentric state directly — **Earth's barycentric position + the supplied geocentric offset**, with no rotation and no Earth-radius scaling. A per-epoch cache key distinguishes moving observatories from fixed stations. The observatory's own geocentric velocity isn't supplied per observation, so Earth's barycentric velocity is used (the satellite's orbital velocity is a small, aberration-only term).

## Why this wasn't caught

The existing space-based test (`test_layup_observatory_obscodes_to_barycentric` over a sats fixture) only checks the output **shape/dtype**, never the actual barycentric values — so a 44-million-km error passed silently.

## Tests

- **`test_moving_observatory_barycentric_position`** — a moving observatory must sit at its supplied geocentric offset from Earth (~6,948 km), not ~6378× further; fails by ~3×10⁷ km without the fix.
- **`test_tno_validation`** — end-to-end: fit the classical TNO **2000 FV53** from the Bernstein et al. (2004) HST-survey astrometry (28 obs, ground + 1 HST, ~19-yr arc) and match JPL Horizons. Recovered state agrees to **~2×10⁻⁵** relative, with a 6D Mahalanobis distance **~1.8** (consistent with the fit's own covariance). Diverges without the fix. Reference stored offline in `tests/data/jpl_reference_2000_FV53.json`.

Closes #149.


matthewholman:
Closes #149.

matthewholman:
Closes #149.

kjnapier (approved):
Looks good

PR #344, opened 2026-06-21
Keep the best IOD seed in the residual prefilter (fixes 609631 macOS divergence)

matthewholman:
## Summary

Fixes the macOS CI divergence on `test_real_data_validation.py[609631]` (`flag=3`/NaN) — a regression from the #324 IOD picker prefilter.

`filter_candidates_by_residual` rejects any Gauss root whose *raw seed* misses a single observation by more than `threshold_sigma` (1000σ). But a **valid-but-rough** 3-point seed propagated across a long arc legitimately drifts well past that. Instrumenting object 609631 on macOS:

| Gauss root | worst residual | outcome |
|---|---|---|
| **2.142 AU** (correct main-belt orbit) | **9462σ** | ❌ dropped by the 1000σ cut |
| 0.732 AU (phantom) | 647990σ | ❌ dropped |
| 0.967 AU (phantom) | — | ✅ kept via the close-Earth bypass |

So the picker received **only a phantom** and LM diverged. The prefilter docstring's premise ("right root within a few σ; phantoms at 10⁵–10⁶σ") is false for a rough seed over a 52-day / 109-obs arc — the right root lands at ~10⁴σ, between the two. Crucially, **LM converges cleanly from that 9462σ seed** (csq=15.8, r_helio=2.140), so the threshold was discarding a recoverable seed.

## Fix

The prefilter is a performance optimization (skip LM on obvious garbage) and must never discard the single best-fitting seed. Track each candidate's worst residual and **always retain the lowest-residual one**, even if above threshold. The common case is unchanged — phantoms far worse than a within-threshold good root are still cut, so the picker still usually gets 1 candidate.

## Validation

- 609631 now converges **with the prefilter active**: `flag=0, csq=15.81, r_helio=2.140 AU`.
- `test_real_data_validation.py -k 609631` passes; `test_iod_picker.py`, `test_orbit_fit.py` all green.
- Adds two integrator-free regression tests (`test_prefilter_keeps_best_root_even_above_threshold`, `test_prefilter_drops_phantom_below_good_root`) pinning the invariant.

Regression introduced by #324.

### DIFF
--- a/src/layup/iod.py
+++ b/src/layup/iod.py
@@ -189,16 +189,30 @@ def filter_candidates_by_residual(
     observations,
     ephem,
     threshold_sigma: float = 1000.0,
+    residual_percentile: float = 80.0,
     min_obs_for_filter: int = 4,
     close_earth_AU: float = _CLOSE_EARTH_AU,
 ):
     """Drop IOD candidates whose predicted positions miss the observations
     by more than `threshold_sigma` times the per-axis astrometric σ.
 
-    The right Gauss root predicts the observations within a few σ; phantom
-    roots are typically off by 10⁵-10⁶ σ. A loose threshold (1000σ
-    default) keeps the right root in every realistic case while throwing
-    out the obviously-wrong ones before LM ever runs on them.
+    The right Gauss root predicts the bulk of the observations within a few
+    σ; phantom roots are typically off by 10⁵-10⁶ σ on essentially every
+    observation. A loose threshold (1000σ default) keeps the right root in
+    every realistic case while throwing out the obviously-wrong ones before
+    LM ever runs on them.
+
+    The per-candidate metric is the `residual_percentile`-th percentile of
+    the per-observation residuals (in σ), *not* the worst single point. A
+    max-residual criterion is brittle: one contaminating observation (or a
+    rough multi-point seed propagated across a long arc) can throw a single
+    large residual that exceeds the threshold even for the correct root.
+    Using a high percentile (80th by default) tolerates a minority of bad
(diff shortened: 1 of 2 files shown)


kjnapier (commented):
Rather than using the wort-fitting point as a rejector for the IOD solution, I would choose some percentile threshold. Maybe the median, 80%, whatever. This will defend against contamination that can be later taken care of by a robust fitter.

It may be a good idea to toss the point that was very inconsistent with the IOD solution, but if there are many such points (for example two groups of points that don't actually go together), that can get messy.

matthewholman:
@kjnapier good call — switched the prefilter from a worst-point criterion to a percentile of the per-observation residuals (`residual_percentile`, default **80th**; `=100` recovers the old max behavior, `=50` gives the median). A single contaminating point no longer rejects an otherwise-good root, and the downstream robust LM cleans up the actual outliers as you note.

I kept it at a high percentile rather than the median deliberately, to guard the failure mode you flagged: with the median, a seed that matches only one of two mis-linked tracklet groups (≤50% of points) could slip through; at the 80th percentile a candidate still has to fit the *majority* of the arc. The keep-best-seed invariant stays as a backstop.

Bonus: this fixes 609631 at the root cause rather than via the backstop — its correct 2.14 AU root had a ~9500σ *worst* point over the 52-day arc but fits the bulk well, so the percentile metric keeps it directly. Added a test for the contaminated-arc case (plus the `=100` legacy equivalence).

Pushed in 301aef4.

kjnapier (approved):
Looks good to me

Issue #140, opened 2025-04-24
C++ orbit fit code handling of streak observations. (needed for KBMOD)


mschwamb:
C++ orbit fit code handling of streak observations. (needed for KBMOD)

matthewholman:
While auditing the streak path I confirmed the current state here, in case it's useful for this issue:

**The fit silently ignores the streak rates.** `Observation.from_streak_with_id` stores `ra_rate`/`dec_rate`, but `compute_residuals` in `orbit_fit.cpp` never reads them (there's no `StreakObservation`/`ra_rate`/`std::visit` handling — only the TODO at `orbit_fit.cpp:16`). Concretely:

- For N streak observations the fit reports `ndof = 2N − 6` (ra/dec only), not `4N − 6` — so the rates contribute no residuals.
- Multiplying `raRate`/`decRate` by 100 and refitting is a **no-op** (identical state, csq, ndof).

So a single tracklet — which *should* be solvable from its rates — is under-constrained and the IOD seed degenerates. That, in turn, surfaced a separate, general bug: such a fit was reported as `flag = 0` (success) with a NaN state. That part is fixed in #343 (`converged()` no longer treats a NaN step as convergence). This issue (actually using the rate information in the fit) remains the real work.

Note the existing streak test (`test_orbitfit_with_streak_observations`) only asserts the *number* of returned orbits, so it passes regardless of whether the rates are used or the state is finite — worth strengthening when the rate residuals land.

matthewholman:
Resolved by #345, which wired streak (sky-motion-rate) observations through the fit end-to-end. The bug noted above — rates parsed but ignored, `ndof = 2N − 6`, the TODO at `orbit_fit.cpp:16` — is fixed on `main`:

- **C++ (`orbit_fit.cpp`):** real `StreakObservation` handling (`std::holds_alternative`) — rate residuals (`ra_rate_resid`/`dec_rate_resid`), closed-form rate partials, variable-row packing (4 rows per streak vs 2), `ndof = 4N − 6`, and rate-uncertainty weighting. The TODO is gone.
- **Python (`orbitfit.py`):** reads `raRate`/`decRate` (and `rmsRArate`/`rmsDecrate`), converts arcsec/hr → rad/day on the great-circle convention (matching Sorcha's `RARateCosDec`), and builds `Observation.from_streak_with_id(...)` with the rates.
[...]

PR #345, opened 2026-06-21
Fit streak (sky-motion-rate) observations (#140)

matthewholman:
## Summary

Wires `StreakObservation` sky-motion rates (ra/dec rates) into the orbit fit, closing the `orbit_fit.cpp:16` TODO behind #140 — previously the rates were parsed and stored but **silently ignored** by the fit. Each streak observation now contributes **four** residual rows (ra/dec position + ra/dec rate) instead of two, so the rates constrain the velocity — the dimension a short position-only arc leaves poorly determined.

Additive and low-risk: the state vector and covariance stay 6×6; astrometry-only fits are unchanged.

## What's in it

**Convention & units (pinned to Sorcha).** ADES `raRate`/`decRate` are arcsec/hour in the great-circle convention (`raRate` = cos(Dec)·dRA/dt), matching Sorcha's `RARateCosDec`/`DecRate`. They were passed through raw while ra/dec were converted to radians; now converted to rad/day at ingest (`ARCSEC_PER_HOUR_TO_RAD_PER_DAY`), with `rmsRArate`/`rmsDecrate` → rate weights. (`predict` computes no rate, so Sorcha — which generates the observed rates and whose `A`/`D` basis equals the fitter's `a_vec`/`d_vec` — is the consistency target.)

**C++ rate residual + partials.** For a `StreakObservation`, `compute_single_residuals` models the great-circle rates ω = d(ρ̂)/dt projected onto the observed `a_vec`/`d_vec`, including the `(1 − dδ/dt)` light-time-rate factor (the residual is fully light-time-consistent, like the position residual). The partials use the closed-form geometric rate Jacobian chained through the position **and velocity** variational particles — the velocity STM was integrated but previously unread (`orbit_fit.cpp:301-303`).

**Variable-row packing.** `residuals`/`partials` gain rate fields + an `n_resid` count (2 or 4); `compute_dX` and `get_weight_matrix` pack a variable number of rows via a prefix-sum offset; `ndof = total_rows − 6`.

## Validation

- **Partials vs autodiff:** the closed-form rate Jacobian matches a jax / finite-difference reference to **~5e-9**.
- **Residual correct:** on a synthetic noise-free arc the fit reaches **csq ≈ 7e-14** across all 4N residual rows (a bug in any rate row would blow this up).
- **Partials correct & live:** velocity covariance tightens monotonically vs position-only as the arc shrinks — **~24× at a 3-minute baseline**, negligible at 10 days (exactly where rates should/shouldn't matter); corrupting a rate raises chi-squared.
- **Physically sane:** modeled rates peak at opposition (~37″/hr for the test object) and vanish at the stationary point.
- Two deterministic offline tests added (`test_streak_validation.py` + `tests/data/streak_synthetic.json`); existing `test_orbit_fit.py` unchanged and green.

Closes the rate-fitting part of #140.


src/lib/detection.cpp
kjnapier, line 194: This is not part of this commit, but are these always hard coded?

src/lib/detection.cpp
```diff
@@ -225,16 +233,21 @@ namespace orbit_fit
             obs.d_vec = d_vec_from_rho_hat(obs.rho_hat);
             obs.ra_unc = 1.0 / 206265;
```
kjnapier, line 234: Always hard coded?

kjnapier (commented):
I have some questions about the astrometry uncertainties. Looking at the from_xxx methods, they seem to always be 1 arcsecond at the moment. Do they get updated somewhere else?

matthewholman:
The astrometric uncertainties get loaded from astrometric_uncertainty_Veres2017, but I think that should be updated/improved in the future.

kjnapier (approved):
Okay, got it re-uncertainties. Modulo that, it looks okay to me.

matthewholman:
@kjnapier good catch — and the name was the confusion: `ra_unc`/`dec_unc` are σ (uncertainties); the fit forms the weight `1/σ²` in `get_weight_matrix`. The `from_*` 1″ is a default that the Veres-2017 σ overrides, but only when `weight_data=True` — which currently defaults to `False`, so by default everything is 1″. Renamed the misleading `data_weight_Veres2017` → `astrometric_uncertainty_Veres2017` in #347. (For streaks, the *rate* σ does come from the data; the *position* σ doesn't yet.)

PR #347, opened 2026-06-22
Rename data_weight_Veres2017 -> astrometric_uncertainty_Veres2017

matthewholman:
The function is named "weight" but returns a one-sigma astrometric **uncertainty** in arcseconds (0.2-1.5"), not a 1/σ² weight. The actual weight is formed only in `orbit_fit.cpp`'s `get_weight_matrix` as `1/σ²`. The misnomer (and the `dw` / `data_weight_*` locals) made the uncertainty-vs-weight distinction confusing — surfaced in the #345 review (kjnapier).

- Rename the function in both copies (`astrometric_uncertainty.py` and the standalone `process_obs80.py`) and all call sites.
- Rename the `dw` accumulator → `sigma_arcsec`; `orbitfit.py`'s `data_weight_arcsec`/`_rad` → `sigma_arcsec`/`sigma_rad`; clarify comments + docstring (one-sigma uncertainty, not a weight).
- Keep `data_weight_Veres2017` as a **back-compat alias** in both modules so no external caller breaks.

Behavior unchanged: `test_weight_data.py` passes; the alias resolves to the same object.

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -30,7 +30,7 @@ except ImportError:  # extension not rebuilt yet
 from layup.convert import convert
 from layup.iod import filter_candidates_by_residual, get_iod, iod_methods
 
-from layup.utilities.astrometric_uncertainty import data_weight_Veres2017
+from layup.utilities.astrometric_uncertainty import astrometric_uncertainty_Veres2017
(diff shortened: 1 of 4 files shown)


kjnapier (approved):
Looks good. No logic changes, just renaming

PR #348, opened 2026-06-23
Drop os.chdir import hack in layup.routines (RPATH resolves the libs) (#75)

### LINKED ISSUE #75: Determine the cause of dynamic linking problems on linux
**Bug report**
discovered and somewhat mitigated during development of #67. 

after we build the `assist` and `rebound` code and compile the `layup` C++ code, python is unable to find the static object files, despite creating symlinks for these files in the project root directory as well as the src directory. There's a workaround by adding the directory to the `$LD_LIBRARY_PATH`, but that's somewhat messy for our linux users.

**Error message**
```
 tests/layup/test_cpp_import.py:1: in <module>
    from layup import routines
src/layup/routines/__init__.py:3: in <module>
    from _layup_cpp._core import __doc__, __version__, hello_world
E   ImportError: librebound.so: cannot open shared object file: No such file or directory
```

**Before submitting**
Please check the following:

- [x] I have described the situation in which the bug arose, including what code was executed, information about my environment, and any applicable data others will need to reproduce the problem.
- [x] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a description of what I expected instead.
- [x] If I have a solution in mind, I have provided an explanation and/or pseudocode and/or task list.

matthewholman:
Small follow-up cleanup to the RPATH fix from #320 that resolved #75.

`src/layup/routines/__init__.py` used to `os.chdir()` to the repo root before
importing `_layup_cpp._core`, "to help MacOS find the library objects." Since
#320 the extension is linked with an RPATH (`$ORIGIN/..` on Linux,
`@loader_path/..` on macOS — see `CMakeLists.txt`), so the dynamic loader finds
`librebound`/`libassist` in site-packages regardless of the working directory.
The chdir is therefore no longer needed, and it had a side effect: it mutated
the process-global cwd at import time.

Verified the extension now imports from an unrelated working directory with no
`chdir` and no `LD_LIBRARY_PATH`:
- `otool` on the built `_core` shows `LC_RPATH @loader_path/..` and
  `@rpath/lib{assist,rebound}` install names;
- `import layup.routines` succeeds from `/tmp` with cwd unchanged afterward;
- `tests/layup/test_cpp_import.py` passes.

No functional change beyond removing the import-time chdir.

### DIFF
--- a/src/layup/routines/__init__.py
+++ b/src/layup/routines/__init__.py
@@ -1,11 +1,7 @@
 from __future__ import annotations
-import os
-from pathlib import Path
 
-# temporarily change the working directory to help MacOS find
-# the library objects, if running in a different directory.
-cwd = os.getcwd()
-os.chdir(Path(__file__).parent.parent.parent.parent)
+# The compiled extension is linked with an RPATH (``$ORIGIN/..`` on Linux,
+# ``@loader_path/..`` on macOS, set in CMakeLists.txt) so the dynamic loader
+# finds librebound/libassist in site-packages regardless of the working
+# directory. The previous os.chdir() dance is therefore no longer needed (#75).
 from _layup_cpp._core import *
-
-os.chdir(cwd)


kjnapier (approved):
Okay

Issue #146, opened 2025-04-27
Add ability to read in and fit radar/doppler observations into layup orbitfit


mschwamb:
Add ability to read in and fit radar/doppler observations into layup orbitfit  - given the rarity of these observations this is an enhancement to be added in after astrometry and stack and shift have been handled properly in orbitfit

matthewholman:
Radar/Doppler observation support has fully landed: C++ core + two-leg light time (#354) and Python ingest + end-to-end/real-data tests (#355), on top of the chi-square fix (#353). All three merged to `main`. Closing.

PR #349, opened 2026-06-23
Use user-supplied observer velocities (ADES vel1/vel2/vel3) in orbit fit (#147)

### LINKED ISSUE #147: Read in and use user provided observatory velocities in layup orbitfit
ADES allows optional velocity columns for the observatory. No examples in the MPC have this information stored, but it could be provided by the user.  See the ADES manual - [OBS_VEL_ADES_Description-2.pdf](https://github.com/user-attachments/files/19931432/OBS_VEL_ADES_Description-2.pdf)

 See the discussion in #144 for further details

matthewholman:
Closes #147.

ADES allows optional observer **velocity** columns (`vel1/vel2/vel3`) for space-based observers. The fit already consumes an observer velocity end-to-end — C++ `Observation.observer_velocity` (fed from the `vx/vy/vz` columns, used for aberration and the streak-rate residual from #345) — but until now that velocity was always the *computed* value, and for a moving observatory #341 approximated it with **Earth's** velocity. This wires the user-supplied velocity through.

### Changes (all in `data_processing_utilities.py`; no C++ changes)
- `populate_observatory` caches an optional geocentric velocity (`vel1/vel2/vel3`) under the same per-epoch key as the position, reusing the `sys`/`ctr` already validated for the position. Absent/NaN → cache untouched (back-compatible).
- `_obs_vel_to_km_s` converts to km/s. **Unit convention: `ICRF_KM` → km/s, `ICRF_AU` → au/day** (confirmed against the ADES OBS_VEL spec).
- `_barycentric_moving_observatory` returns `earth_vel + obs_vel` when supplied, else falls back to Earth's velocity (the prior #341 behavior).
- Readers need no change — `vel1/vel2/vel3` pass through by header name (Obs80 has no velocity; ADES-only, per spec).

### Why
Removes the Earth-velocity approximation #341 noted for space-based observatories (the satellite's ~km/s orbital velocity is not negligible for precise astrometry), tightens streak-rate fits for moving observers, and is the prerequisite for radar/Doppler (#146), where the observer's true velocity is mandatory.

A natural follow-up (#55) can populate the same `ObservatoryVel` cache from a JPL/SPICE query for known spacecraft.

### Tests
Static unit conversion, ICRF_KM + ICRF_AU end-to-end (vs Earth's velocity), absent→Earth-velocity fallback, and a same-epoch inconsistency guard. Full `test_data_processing_utilities.py` passes (49); black/isort clean.

### DIFF
--- a/src/layup/utilities/data_processing_utilities.py
+++ b/src/layup/utilities/data_processing_utilities.py
@@ -378,6 +378,13 @@ class LayupObservatory(SorchaObservatory):
         # A cache of barycentric positions for observatories of the form {obscode: {et: (x, y, z)}}
         self.cached_obs = {}
 
+        # Optional user-supplied geocentric observer velocities (km/s, ICRF),
+        # keyed by the same per-epoch cache key as ObservatoryXYZ (issue #147).
+        # Empty unless ADES vel1/vel2/vel3 columns are provided for a moving
+        # observer; _barycentric_moving_observatory falls back to Earth's
+        # velocity for keys that are absent here.
+        self.ObservatoryVel = {}
+
     def convert_to_geocentric(self, obs_location: dict) -> tuple:
         """Convert an observatory's parallax constants to geocentric coordinates.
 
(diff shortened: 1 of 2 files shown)


kjnapier (approved):
As long as the ADDS velocities are conventionally in au/day, this looks good

matthewholman:
I couldn't actually find the velocity units in the ADES document.

PR #353, opened 2026-06-23
Fix orbit_fit reporting a stale chi-square when LM hits iter_max

matthewholman:
## What
`orbit_fit()` left `chi2_final` unset when the Levenberg-Marquardt loop ran to `iter_max` without converging, so a non-converged fit reported a garbage/zero chi-square. Now `chi2_final` is initialized and refreshed every iteration, so the reported csq is always the honest value for the returned state.

## Why
Same family as #343. It's a pre-existing bug independent of any new feature, but radar's stiff weights (1 µs ≈ 150 m) hit the non-converged path constantly and exposed it, so it's pulled out here as its own small fix ahead of the radar stack (#146).

## Note
A deeper issue — `converged()` reporting success when the LM stalls with dX→0 against a gross outlier — is **not** addressed here; it's a separate concern.

### DIFF
--- a/src/lib/orbit_fit/orbit_fit.cpp
+++ b/src/lib/orbit_fit/orbit_fit.cpp
@@ -744,6 +744,12 @@ namespace orbit_fit
 
         double chi2_prev = HUGE_VAL;
 
+        // Initialise so that a fit which runs to iter_max without converging (or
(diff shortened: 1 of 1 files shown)


Little-Ryugu (approved):
Looks good to me!

PR #354, opened 2026-06-23
Add radar (delay/Doppler) observation fitting — C++ core (#146)

matthewholman:
## What
Radar (delay/Doppler) observation fitting in the C++ core (issue #146, P1), now with a **two-leg light-time model** validated against real JPL radar.

- `RadarObservation` variant + `from_radar`/`from_radar_with_id` factories and pybind bindings; `delay_unc`/`doppler_unc` and a new `observer_acceleration` field (default zero, so optical/streak paths are unaffected).
- Delay/Doppler residual + partials and generalized variable-row packing (1 row for delay-only/Doppler-only, 2 for both) in `orbit_fit.cpp`/`.h`.

## Two-leg light time
A real-data cross-check against (99942) Apophis radar astrometry showed the naive single-position model (`delay = 2 rho/c`, one station position/time) is off by **thousands of µs in delay and ~100 Hz in Doppler** — far above the ~1 µs / sub-Hz measurement precision. The dominant error is the round-trip station displacement (the up-leg was transmitted ~one round-trip earlier, when Earth had moved ~3400 km). The model now solves both legs: down leg with the station at receive, up leg with the station Taylor-extrapolated (pos+vel) to the transmit time via the observer acceleration. Partials keep the single-leg×2 Jacobian (legs parallel to ~2e-4 rad; convergence is unaffected). Shapiro delay (~2 µs) is the next refinement.

## Tests
`test_radar_validation.py` (3) + noise-free fixture (regenerated to the two-leg convention): recovery from a perturbed seed to csq<1e-6, liveness, row counts.

## Stacking
Based on the chi2_final fix PR; the Python ingest + real-data Apophis validation are in the next PR on top. Merge that one first.


Little-Ryugu (approved):
Looks good to me! I'm not fully confident on my c++ skills yet but i walked through what you did @matthewholman and it looks all correct. 

Unit test and code coverage / build (ubuntu-latest, 3.14) seems to have gotten stuck and cancelled so I set it to rerun. All the unit tests worked on my laptop and on the github except ubuntu 3.14 so probably just a github software hicupp.

matthewholman:
Thank you, @Little-Ryugu!

Issue #128, opened 2025-04-21
unit tests after sending to various  C++ components of layup orbitfit


mschwamb:
unit tests after sending to various  C++ components of layup orbitfit

matthewholman:
Closing — the C++ core now has unit-test coverage across its components, all merged to `main`:

- **#363** — ephem-free tests for the Gauss IOD solver and the IAS15 integrator settings (+ a `d0` degeneracy guard in `gauss.cpp`).
- **#372** — the pybind11 binding surface: `numpy_to_eigen` round-trips, `FitResult`/`PredictResult` round-trips (including the non-grav fields), and the `Observation` streak/radar factory methods.
- **#374** — the fit/predict integration tests that exercise the C++ paths end-to-end (nongrav A2, radar, streak, and the MBA/TNO/3I-ATLAS validation cases) now actually run on Linux CI rather than being silently skipped.

Together these cover IOD, integration, the Python↔C++ marshalling layer, and the full fit/predict pipeline. The one deeper item not in scope here — finite-difference checks of the radar/streak *analytic partials* via a thin debug binding — is worth a dedicated follow-up; note those partials were already validated against JAX when the radar (#146) and streak (#140) features landed. Filing a separate enhancement if/when we want the FD cross-check.

PR #363, opened 2026-06-29
Ephem-free C++-core unit tests (gauss IOD, IAS15) + gauss d0 guard

### LINKED ISSUE #359: Run the ephem-gated fit tests in CI (currently skipped)
The C++ fit core currently has **~no CI coverage**: the radar, streak, and `bk_fit` end-to-end tests are all gated behind the ~600 MB ASSIST ephemeris (`@pytest.mark.skipif(...)`) and are therefore **skipped in CI**. They report green as "skipped", so coverage *looks* fine while the headline features and the core fit loop are never actually exercised before release.

**Task:**
- [ ] Make the fit core testable without the full ephemeris — a small cached ephemeris slice fixture, a mock, or an ephem-free path — so these tests run in CI rather than skip.
- [ ] Add ephem-free unit tests for the internal residual/partials (radar delay/Doppler, streak rates) via finite-difference-vs-analytic (reuse the `test_bk_basis.py` idiom), adding a thin debug binding for `compute_single_residuals`/`integrate_light_time`/`compute_dX` if needed.
- [ ] Direct numeric `gauss()` test (today only registry-presence is checked).
- [ ] Cover the untested ias15 setters and `numpy_to_eigen`/result round-trips.

Closely related to #128 (C++ unit tests). This is the biggest test-confidence gap before v1.0.

_Tier 0 item of the v1.0 code-clarity audit._

matthewholman:
Closes #359. Part of #128.

- `tests/layup/test_gauss.py`: direct numeric test of the Gauss angles-only IOD
  binding -- builds a known two-body orbit, propagates it with a universal-variable
  Kepler solver, synthesizes exact lines of sight, and checks `gauss()` recovers
  the middle state. No ASSIST ephemeris required, so it runs in CI.
- `tests/layup/test_ias15_settings.py`: round-trip coverage for the four IAS15
  tuning knobs (previously untested).
- `gauss.cpp`: return `nullopt` for degenerate/coplanar lines of sight (d0 floor)
  instead of a non-finite state; the degenerate test asserts this.

### DIFF
--- a/src/lib/gauss/gauss.cpp
+++ b/src/lib/gauss/gauss.cpp
@@ -89,6 +89,19 @@ namespace orbit_fit
 
         double d0 = rho1.dot(p1);
 
+        // Gauss's method divides through by d0, the scalar triple product of the
+        // three lines of sight rho1 . (rho2 x rho3). It vanishes when the three
+        // pointings are coplanar/degenerate (e.g. identical directions), leaving
+        // the topocentric range indeterminate. A real triplet -- even a 1-day arc
+        // -- keeps |d0| >~ 1e-8 thanks to the slight curvature of the apparent
+        // path, so this small floor rejects only the singular geometry instead of
+        // propagating a division by ~0 into a non-finite state.
+        constexpr double D0_MIN = 1e-12;
+        if (std::abs(d0) < D0_MIN)
+        {
+            return std::nullopt;
+        }
+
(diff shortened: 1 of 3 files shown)


kjnapier (approved):
Looks good to me

matthewholman:
Thank you for the quick turnaround

Issue #360, opened 2026-06-28
v1.0 code-clarity & simplification audit (tracking)


matthewholman:
Tracking issue for the pre-v1.0 clarity/maintainability pass (audit done 2026-06-28 over `main`). Goal: make the code easy for a researcher to read/maintain and easy for JOSS reviewers to evaluate. **Quality-only — no behavior changes.** The BK math layer (`bk_basis`, `bk_iod`) is already well-documented; debt is concentrated elsewhere.

**Tier 0 (highest leverage)** — tracked separately: #358 (constants), #359 (run ephem-gated tests in CI), plus splitting `compute_single_residuals` (below).

### C++ (`src/lib`)
- [ ] `orbit_fit.cpp`: split `compute_single_residuals` (~300L, 3-way optical/streak/radar branch with 3 copy-pasted light-time partial blocks); name `206265` (~15 sites) + LM magic numbers; remove stale to-do banner (L1–36); delete dead `detection`-based path (`read_detections`/`print_residuals`/`read_initial_conditions`, empty `identify_outliers`); fix `converged()` predicate (L744); `obs_kind` → `ObsKind`; fix `.epoch` "utc jd" comment (used TDB-relative)
- [ ] `bk_fit.cpp`: document/guard astrometry-only 2N row layout (silently drops streak/radar rows); extract shared LM-step helper (currently a copy of `orbit_fit()`)
- [ ] `gauss.cpp`: name/cite the polynomial-derivation vars; explain `1e-10` root tol
- [ ] `predict.cpp`: remove `printf("barf…")` debug; dedup `drho_*` partials
- [ ] `detection.cpp`: dedup `Observation` constructors/factories; magic `24.0/206265.0`
- [ ] cross-cutting: pick one error-handling convention; reconsider unity-build `#include "*.cpp"`

### Python (`src/layup`)
- [ ] split long functions: `do_fit` (~180L, name the flag values), `orbitfit_cli` (~200L), `convert._apply_convert` (~250L), `comet._apply_comet`
- [ ] remove dead/commented blocks + debug `print`s (orbit_conversion `tp` block, predict recarray, comet prints→logger, stale `jax.lax.cond` alts, unused imports)
- [ ] name `250 au` reference distance (comet, 6+ sites)
- [ ] fix docstring/param mismatches (predict `aux`, comet `_direction_of_integration`, orbit_conversion `root_function`); `MJD_TO_JD_CONVERSTION` typo; mutable default args; `== True` idioms
[...]

Issue #93, opened 2025-03-16
optimize layup convert further


mschwamb:
If we assume it takes 100,000 mathematical operations to convert one orbit (probably a 10x overestimate?), for ~10M orbits it should take ~100 GFLOPs to do it.

A single Apple M4 core (just as an example, they're all similar within 2x) has about 8-10 GFLOPS of double-precision power available to it. Let's conservatively take 5 GFLOPs.

Dividing the two, it should be possible to write an implementation to convert 10M orbits in ~20 seconds, on a single core.

Gauntlet thrown :).

mschwamb:
@astjoephysics this would be something you could help with. Is  test how long it takes if you try and read your entire file in as one chunk and request only one core - as worth testing how long the transform is versus how long the write out time is. You might new logging/print statements to help estimate that.

drewoldag:
FWIW I was observing the processing using `top` this morning, and noted that for the following configuration (~9.5M row CSV, 10 workers, 1M chunk size) that a significant amount of time is spent between chunks (namely all the file I/O).

The actual processing time for any given chunk was pretty small (2-5ish seconds) then several seconds of non-processing time during file I/O. 

This wasn't meant to be a detailed study of where the time was spent, simply an observation made while implementing the `--num-workers` CLI argument.

matthewholman:
**Profiling (measurement-first), 2026-06-28**

`convert` runs at ~7,000 rows/s single-worker (~0.14 ms/row). Findings:

- `_create_assist_ephemeris` is effectively cached (~0.001 s/call), so the per-chunk/per-worker rebuild in `convert_cli` is **not** a bottleneck (rules out the obvious build-once-and-pass-in optimization).
- Dominant per-row cost is **jax scalar → float conversion (~31%)**: `universal_keplerian`/`universal_cometary`/`universal_cartesian` return tuples of jax scalars, which numpy converts one-by-one (6/row) when building the output array. Then `parse_orbit_row` (~17%) and `ephem.get_particle` (~12%).
- No cheap single win remains: pre-converting the row tuple does not reduce the jax-scalar count.

The only **substantial** speedup is vectorizing `_apply_convert` (batch the per-row jax calls over whole columns, ~1 dispatch instead of N), which is a real rewrite of the element/covariance path with corresponding risk. Since `convert` is not a hot path (unlike fitting), **vectorization is deferred past v1.0**.

[...]

PR #364, opened 2026-06-29
convert: deduplicate _apply_convert output-format dispatch

matthewholman:
Part of #360 (clarity audit). (Perf optimization of `convert` is tracked
separately in #93 and deferred past v1.0 -- see profiling notes there.)

- Collapse the four near-identical cometary/Keplerian output branches into one
  block keyed on (barycentric?, cometary?); extract `_subtract_sun`/`_to_ecliptic`.
- Drop two redundant Sun-particle re-fetches; fix the `MJD_TO_JD_CONVERSTION` typo;
  replace mutable default args with `None`.

Behavior-preserving: all `test_convert.py` tests pass.

### DIFF
--- a/src/layup/convert.py
+++ b/src/layup/convert.py
@@ -85,7 +85,24 @@ def _scale_degree_cov(arr, fmt, factor, mask=None):
 
 
 # Add this to MJD to convert to JD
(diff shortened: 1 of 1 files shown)


kjnapier (approved):
Looks good, this removed a lot of logic branches

PR #366, opened 2026-06-29
orbit_fit.cpp clarity: dead-code removal + ARCSEC_PER_RAD

### LINKED ISSUE #367: converged(): reduced-chi2 check compares a bool to the threshold (dead branch)
### Summary
`converged()` in `src/lib/orbit_fit/orbit_fit.cpp` has a reduced-chi-square check that can never fire because it compares a **boolean** to the threshold:

```cpp
if ((chi2 > ndof) > thresh)   // (chi2 > ndof) is 0 or 1; thresh == 10
    return 2;
```

`(chi2 > ndof)` evaluates to `0` or `1`, and the caller passes `thresh = 10`, so `(0|1) > 10` is **always false** — the `return 2` branch is dead.

### Evidence of intent
The very same file already writes the intended reduced-chi-square test correctly in the post-loop check:
```cpp
double thresh = 10.0;
if ((chi2_final / ndof) > thresh)   // <- correct form
```
So `converged()` almost certainly meant:
```cpp
if ((chi2 / ndof) > thresh)
    return 2;
```

### Impact
- The `return 2` ("converged but reduced-chi2 too high") signal never occurs.
- Practical effect is currently limited because the post-loop check (`chi2_final / ndof > thresh`) catches the poor-fit case separately, but `converged()`'s own contract is broken.

### Decision needed (why this isn't just a silent fix)
The lone caller does `int cflag = converged(...); if (cflag) { ... }` — it treats the return as truthy/falsy, so once `return 2` starts firing it would be handled **identically to `return 1` (converged)**. So fixing the predicate also requires deciding the intended semantics of flag/return 2 (distinct "converged-but-bad-fit" outcome vs. just converged). Hence filing rather than silently changing.

### Suggested fix
Change the predicate to `(chi2 / ndof) > thresh` and make the caller distinguish `2` from `1` if a separate poor-fit flag is intended.

Found during the #360 clarity audit (the predicate was flagged but intentionally left unchanged in PR #366 since it is a behavior change).

matthewholman:
Part of #360 (clarity audit). Quality-only, behavior-preserving.

- Remove the stale top-of-file to-do/compile banner.
- Delete the dead legacy detection-based path (read_initial_conditions, print_initial_condition, read_detections, the empty identify_outliers stub, print_residuals) -- zero call sites; superseded by the Observation type. -178 lines.
- Name 206265 -> ARCSEC_PER_RAD for the live sites (streak-rate default uncertainties, LM initial lambda).

Full tests/layup suite: 298 passed / 1 skipped (only the pre-existing global-~/assist shell-out failures).

Deferred intentionally: the converged() `(chi2 > ndof) > thresh` predicate is a latent logic bug (needs intent, not a clarity change); splitting compute_single_residuals is largely obviated by the radar shared-variational block; obs_kind->ObsKind and the .epoch comment live in headers/detection.cpp.

---
Also removes the dead reduced-chi2 branch in converged() (was comparing a bool to thresh; the correct check already runs post-loop). **Closes #367.**

### DIFF
--- a/src/lib/orbit_fit/bk_fit.cpp
+++ b/src/lib/orbit_fit/bk_fit.cpp
@@ -178,10 +178,8 @@ FitResult run_bk_native_fit(
         }
 
         // --- Convergence (same predicate as the Cartesian fit) ---
-        const size_t ndof = detections.size() * 2 - 6;
-        const double thresh = 10.0;
         Eigen::MatrixXd dX_mat = dX;
-        if (converged(dX_mat, eps, chi2_cur, ndof, thresh))
+        if (converged(dX_mat, eps, chi2_cur))
         {
             result.flag = 0;
             result.csq = chi2_cur;
--- a/src/lib/orbit_fit/orbit_fit.cpp
+++ b/src/lib/orbit_fit/orbit_fit.cpp
@@ -617,20 +617,18 @@ namespace orbit_fit
     }
 
 
-    int converged(Eigen::MatrixXd dX, double eps, double chi2, size_t ndof, double thresh)
+    int converged(Eigen::MatrixXd dX, double eps, double chi2)
     {
         // A NaN chi-square or step means the fit has diverged, not converged.
         // Without this guard the loop below reports convergence on a NaN step,
         // because abs(NaN) > eps is false for every element, so the function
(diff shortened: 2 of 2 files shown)


kjnapier (approved):
Just dead code removal, no issues

PR #369, opened 2026-06-29
CI: add pytest-timeout so a wedged test fails fast (#368)

### LINKED ISSUE #368: CI: add a pytest timeout so a wedged test fails fast instead of hanging for hours
### Symptom
The `Unit test and code coverage` CI job intermittently **hangs for hours** on the
`Run unit tests with pytest` step. It strikes a *random* matrix leg (observed:
ubuntu-3.14 on #354, then ubuntu-3.13 on #365) while every other leg passes green,
so the PR sits at `blocked`/`unstable` until someone manually cancels the run and
re-runs the failed job. Seen at least 3 times in the last few days.

### Likely cause
A test blocks on a **network/ephemeris fetch with no timeout** (e.g. a pooch /
NAIF / MPC download). When the remote is slow or unresponsive the test waits
indefinitely instead of failing, wedging the whole job (GitHub's 6h job timeout is
the only backstop).

### Proposed fix
- Add **`pytest-timeout`** to the dev/test dependencies and set a per-test cap in
  the pytest config, e.g.:
  ```toml
  [tool.pytest.ini_options]
  timeout = 300
  timeout_method = "thread"
  ```
  A wedged test then fails fast with a traceback (pointing at the exact hang)
  instead of hanging the run.
- Optionally, give the network/ephemeris fetches their own explicit timeouts.

### Related
#359 (de-gate the ephem-gated fit tests so they don't reach out to the network at
all) addresses the root cause for the fit suite; this issue is the broad backstop
for *all* tests so a single slow fetch can't hang CI again.

matthewholman:
Closes #368.

Adds `pytest-timeout` (dev dep) and a 300s per-test cap in `[tool.pytest.ini_options]` (`timeout_method="thread"`, compatible with pytest-xdist and all platforms), so a test that wedges on a network/ephemeris fetch fails fast with a traceback instead of hanging the CI job for hours (observed on ubuntu-3.14 and ubuntu-3.13 across recent PRs).

Verified locally: pytest reads the config (`timeout: 300.0s / method: thread`) and enforcement fires (a 5s sleep fails at ~1s under `--timeout=1`).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,6 +59,7 @@ dev = [
     "pytest",
     "pytest-cov", # Used to report total code coverage
     "pytest-xdist", # Used to parallelize unit tests
+    "pytest-timeout", # Fail a wedged test fast instead of hanging CI (issue #368)
 ]
 
 [build-system]
@@ -80,6 +81,11 @@ write_to = "src/layup/_version.py"
 testpaths = [
     "tests",
 ]
+# Cap any single test so a wedged test (e.g. a network/ephemeris fetch with no
+# timeout of its own) fails fast with a traceback instead of hanging the CI job
+# for hours (issue #368). "thread" works with pytest-xdist and on all platforms.
+timeout = 300
+timeout_method = "thread"
 
 [tool.black]
 line-length = 110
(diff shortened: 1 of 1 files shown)


kjnapier (approved):
Good

matthewholman:
Thank you!

Issue #361, opened 2026-06-28
JOSS: worked-example notebook(s) + docs for submission (tracking)


matthewholman:
JOSS requires a runnable worked example and clear usage docs; reviewers run the code. This tracks the notebook/doc deliverables for the JOSS submission and clarifies which existing notebook issues are in vs. out of JOSS scope.

**In scope for JOSS (required):**
- [ ] #289 — notebook running orbitfit through direct API calls → the canonical end-to-end **worked example** JOSS reviewers will execute
- [ ] #295 — orbit-plotting / visualization demo notebook → also supplies **figures for the methods paper**
- [ ] README quickstart that points to the example notebook
- [ ] Notebooks run top-to-bottom in CI (or binder) — no skipped/manual cells — so the example is verifiably reproducible

**Explicitly NOT JOSS-blocking** (deferred; nice-to-have polish): #19, #34 (interactive reject/refit widgets), #305/#307/#308 (visualize polish).

Related: the JOSS `paper.md`/`paper.bib` themselves (Sorcha used as the model).

matthewholman:
Closing — the JOSS worked-example / docs deliverables tracked here are complete:

- **#289** — orbit-fit worked-example notebook via the Python API (merged in #370)
- **README quickstart** pointing at the example (merged in #379)
- **#295** — orbit-plotting/visualization demo notebook, pre-executed for docs (PR #391, in review)

The notebooks render with committed outputs (the ephemeris/Dash-dependent plotting notebook lives in `docs/pre_executed/` with `execute:never`; the API notebook runs in the docs build). Remaining JOSS work is the `paper.md`/`paper.bib` themselves, tracked separately. Deferred/interactive notebook ideas (#19, #34) stay out of JOSS scope.

PR #370, opened 2026-06-29
docs: JOSS worked-example notebook for the orbit-fit API (#289)

### LINKED ISSUE #289: Jupyter notebook to run layup orbitfit through direct calls to the function APIs
Get an example notebook using 3I/ATLAS as a good use case where the structured numpy arrays are made and sent through the initial orbitfitting routines (included IOD steps) conversation of elements etc...

matthewholman:
Closes #289. Part of #361 (JOSS worked-example/docs deliverable).

Adds `docs/notebooks/orbit_fitting_api.ipynb`, an end-to-end walk-through of layup's Python API:

1. **Load** MPC/ADES astrometry with `CSVDataReader`.
2. **Fit** an orbit with `orbitfit` (Gauss IOD + Levenberg-Marquardt / ASSIST refinement).
3. **Inspect** the fitted BCART_EQ state and reduced chi-square.
4. **Convert** to Keplerian elements with `convert`.
5. **Predict** future RA/Dec (with the on-sky 1-sigma error ellipse) via `predict`.

Self-contained: ships a 587-observation example file for (119839) 2002 CX17 (`docs/notebooks/example_observations.csv`) so the notebook needs no external data, and is committed pre-executed (matching the docs convention that excludes `docs/notebooks/` from output-clearing). Registered in `docs/notebooks.rst`.

Verified top-to-bottom with `jupyter nbconvert --execute` (fit converges, reduced chi-square 0.24).

> Reviewer note: running it requires the ASSIST ephemeris (`layup bootstrap`), as the notebook states up front.


Little-Ryugu (approved):
Notebook looks good and walkthrough is very clear. 

Also shows up on the documation link with no issues. 

I noticed that this warning below, is appearing whenever orbitfit and predict are ran. Not an issue for this pull request but we might need to replace astropy.tests.runner with pytest in those files 
WARNING: AstropyDeprecationWarning: The TestRunner class is deprecated and may be removed in a future version.
        Use pytest instead. [astropy.tests.runner]
WARNING: AstropyDeprecationWarning: The TestRunnerBase class is deprecated and may be removed in a future version.
        Use pytest instead. [astropy.utils.decorators]

Little-Ryugu:
The next steps section at the end was a little confusing though. In the future when we add the rest of the documention we could link those parts to the places in the documentation that discuss them in more detail.

matthewholman:
Thanks for the review, @Little-Ryugu!

**"Next steps" section** — good catch, it read as a bare options list with no recap. I've reworked it (`d66c8d3`): renamed to "Recap and next steps", added a one-line summary of the `load → orbitfit → convert → predict` workflow, and noted the options will be documented more fully as the docs grow (we can link each to its docs section once those pages exist).

**`AstropyDeprecationWarning` (TestRunner)** — I traced this and it's *not* coming from `orbitfit`/`predict`; those don't touch `astropy.tests.runner`. The source is **`sbpy`** (`sbpy/_astropy_init.py` builds an astropy-style `TestRunner` at import time), which we pull in transitively. `import sbpy` alone reproduces it. So the real fix is upstream in sbpy; on our side the only lever is to suppress that specific warning for clean CLI output. I'll open a separate follow-up issue to track that rather than hold up this PR.

matthewholman:
Filed the `AstropyDeprecationWarning` (sbpy `TestRunner`) follow-up as #376 so it doesn't hold up this PR.

PR #362, opened 2026-06-29
Single source of truth for physical constants + predict/data_processing clarity

matthewholman:
Closes #358. Part of #360 (clarity audit).

- New `src/layup/constants.py` with one definition each of `AU_M`, `AU_KM`,
  `SPEED_OF_LIGHT`, `MU_SUN`, `GMtotal` (with provenance comments). Migrates
  `iod.py`, `predict.py`, `data_processing_utilities.py` to import from it.
  (`SPEED_OF_LIGHT` was previously derived two different ways that differed by
  ~1 ULP; now a single canonical value.)
- Clarity in `predict.py` (malformed docstring + `aux`->`configs`, dead recarray
  block, `Layup_observatory` rename, placeholder columns) and
  `data_processing_utilities.py` (`parse_fit_result` mutable default/double-assign/
  `==True`; `populate_observatory` 5-block->loop + typo; `get_format` builtin
  shadow/dead init).

Quality-only, behavior-preserving. `orbitfit.py`'s own constant copies are left for
a small follow-up (that file was in flight at the time).

Issue #358, opened 2026-06-28
Introduce layup.constants as single source of truth for physical constants


matthewholman:
Physical constants are currently redefined independently across ≥4 modules (`orbitfit.py`, `predict.py`, `iod.py`, `utilities/data_processing_utilities.py`): `AU_M`, `GMtotal`/`MU_SUN`, `SPEED_OF_LIGHT`.

**Correctness risk:** `SPEED_OF_LIGHT` is derived two different ways — `orbitfit`/`iod` compute it via `AU_M`, `predict` via `AU_KM`. These should be one value.

**Task:** create `src/layup/constants.py` with one authoritative definition of each constant (with provenance comments / citations — e.g. IAU AU, GM values), and import from it everywhere. Removes the duplicate definitions flagged in the v1.0 clarity audit, and gives the methods paper a single citable source.

**Acceptance:** no constant defined in more than one place; `c` identical everywhere; tests green.

_Tier 0 item of the v1.0 code-clarity audit._

PR #371, opened 2026-06-30
orbitfit: use layup.constants (finish constants consolidation)

matthewholman:
Part of #360. Finishes the `layup.constants` single-source-of-truth migration begun in #362 (#358).

orbitfit.py kept its own `AU_M` / `SPEED_OF_LIGHT` / `_MU_SUN` copies because it was in flight when #362 merged. Now:
- import `SPEED_OF_LIGHT` and `MU_SUN as _MU_SUN` from `layup.constants`;
- drop the local `AU_M` (only used to derive `SPEED_OF_LIGHT`) and the dead `GMtotal` (unused in orbitfit; iod.py has its own).

The `_MU_SUN` and `SPEED_OF_LIGHT` names remain importable from `layup.orbitfit` (tests rely on that). Behavior-preserving, values unchanged; test_bk_fit / test_bk_everywhere / test_orbit_fit (30 tests) pass.

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -29,6 +29,9 @@ try:
 except ImportError:  # extension not rebuilt yet
     get_ias15_adaptive_mode = lambda: -1
     set_ias15_adaptive_mode = lambda m: None
+# _MU_SUN (= heliocentric GM = k^2) is used by the BK-native fit for the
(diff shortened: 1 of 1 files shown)


Little-Ryugu (approved):
Looks good to me!

matthewholman:
Thank you!

Issue #351, opened 2026-06-23
Fit for non-grav parameters


matthewholman:
Although ASSIST covers the terms for the Marsden representation of non-gravitational acceleration, including the partial derivatives/variational equations for those terms, we are not presently fitting for them in layup.   This is challenging because we represent the state of a small body with a 6D vector for its combined position+velocity vectors.  All the current internals work with 6D vectors, etc.  To include additional fitted parameters (A1, A2, A3, for example), we either need to expand that 6D vector or we need to carry around a secondary vector for the non-grav parameters and make sure we gracefully handle the cases with and without non-gravs.

matthewholman:
Non-gravitational parameter fitting (A1/A2/A3, Yarkovsky) landed in #356 (merged). Closing.

PR #372, opened 2026-06-30
Add ephem-free unit tests for the C++ binding surface (#128)

matthewholman:
Part of #128 (unit tests for C++ components). Pure ephem-free test additions; no source changes.

Adds `tests/layup/test_cpp_bindings.py` covering binding surface that had no direct coverage and needs no ASSIST ephemeris (runs in CI):
- `numpy_to_eigen` — flat-array -> m x n matrix round-trip (row-major) + size-mismatch error path.
- `FitResult` / `PredictResult` — field round-trips, including the A1/A2/A3 non-grav fields (#351) and their zero defaults.
- `Observation.from_streak` / `from_radar` — streak (ra_rate/dec_rate) and radar (delay/doppler/has_*) variant payloads + computed rho_hat.

Complements #363 (gauss/ias15). 11 tests, all pass.

> The radar/streak *partial* FD tests (which need ASSIST integration) belong with the de-gating work in #359, not here.

### DIFF
--- a/tests/layup/test_cpp_bindings.py
+++ b/tests/layup/test_cpp_bindings.py
@@ -1,0 +1,126 @@
+"""Unit tests for the small C++ <-> Python binding surface of the layup core.
+
+These cover bindings that previously had no direct coverage and need no ASSIST
+ephemeris, so they run anywhere (including CI without the ephemeris cache):
+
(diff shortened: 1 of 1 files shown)


kjnapier (approved):
All good

PR #373, opened 2026-06-30
CI: surface skipped tests with pytest -rs (#359 step 1)

matthewholman:
First step toward #359 (visibility before the de-gating fix).

Adds `-rs` to the pytest invocations (testing-and-coverage + smoke-test) so skipped tests + reasons appear in the CI log.

**Why:** a green CI run shows 283 passed / **19 skipped**, but the default `pytest -n auto` output only prints a count. Locally (ASSIST ephem present) the same invocation gives 298 passed / 1 skipped, so ~18 ephem-gated suites (BK engine, radar, non-grav, JPL validation) silently skip on CI. `-rs` will show exactly which suites skip and why on the next run, so the real de-gating fix (ensure the ASSIST ephem is at the guard'''s cache path on CI / fail loudly if absent) can be targeted.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -51,4 +51,5 @@ jobs:
         pip list
     - name: Run unit tests with pytest
       run: |
-        python -m pytest -n auto
+        # -rs surfaces skipped tests + reasons in the CI log (issue #359).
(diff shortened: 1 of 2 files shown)


kjnapier (approved):
No problems here

PR #377, opened 2026-06-30
Resolve space-based observatories via JPL Horizons (Closes #55)

### LINKED ISSUE #55: Convert Rust code for querying JPL for special case observatories

matthewholman:
Closes #55. Scoped in https://github.com/Smithsonian/layup/issues/55#issuecomment-4846807848.

Spacecraft observatory codes (HST, JWST, WISE, ...) have no MPC parallax constants, so layup could only place them when the user supplied an explicit ADES per-observation position; otherwise the fit raised `ValueError`. This ports the spacecraft `obscode -> NAIF-id` map from the Rust `spacerocks` project and queries **JPL Horizons** for the observer's geocentric state when no position is given.

### What it does
- **New `utilities/special_observatories.py`** — the 9-entry `SPACE_OBSERVATORIES` map + `query_horizons_geocentric()`. Asks Horizons for a **geometric** (`VEC_CORR=NONE`), **geocentric** (`CENTER=@399`), **ICRF-equatorial** state in **km / km·s⁻¹** (`OUT_UNITS=KM-S` — layup's internal moving-observer units), and parses the CSV vector table. Epochs are **batched** (one request per spacecraft).
- **`populate_observatory()`** — in the no-parallax-constants branch, a space-based obscode *without* a usable ADES position is routed to Horizons; the geocentric position + velocity are cached under the same per-epoch key #147 uses, so `_barycentric_moving_observatory` adds Earth's barycentric state to it. **A user-supplied ADES position always takes priority.**
- **`obscodes_to_barycentric()`** — pre-pass warms the cache for all space-based epochs, one batched HTTP call per spacecraft.

### Design decisions (per the issue)
1. **Horizons HTTP** (not SPK kernels) — the case is rare; avoids per-spacecraft kernel management. SPK-backed resolution can be a future enhancement.
2. **Geocentric query** (`CENTER=@399`) — drops straight into the existing #147 "geocentric + add Earth" path; no new downstream code.
3. **In-run batched cache** only; cross-run disk caching deferred.

Scope boundaries: roving (`247`/`270`) and occultation (`275`) codes stay ADES-supplied (not Horizons-resolved).

### Validation
- 63 tests pass; black clean.
- **Live Horizons cross-check** (HST): |r| ≈ 6916 km (≈537 km LEO altitude ✓), |v| ≈ 7.59 km/s ✓ — confirms units, frame, and parsing against the real server.
- `test_tno_validation` (real HST data) still passes — confirmed no-op, since its fixture supplies ADES positions (which keep priority).

### Tests
Pure-logic tests (map, JD conversion, CSV parser, HTTP client with the network **mocked**) run everywhere. `LayupObservatory` integration tests (ephem-gated, HTTP mocked) cover the Horizons path, the ADES-override priority, request batching, and end-to-end barycentric placement. No test touches the network.

The one existing-test change swaps the generic moving-observer code `C51` → the roving code `247` in `test_moving_observatory_coordinate_cache` (since `C51` is now Horizons-resolved, it no longer exercises the "ADES position required" error path).

### DIFF
--- a/src/layup/utilities/special_observatories.py
+++ b/src/layup/utilities/special_observatories.py
@@ -82,6 +82,25 @@ def query_horizons_geocentric(naif_id, jd_tdb_list, timeout=30):
         state at the requested TDB epoch is returned, matching how layup
         supplies every other observer position (light time is handled later, in
         the integrator).
+
+    Notes
+    -----
+    We deliberately query the **geocentric** spacecraft state (center = Earth's
+    body center) rather than the barycentric one, so this plugs into the same
+    "geocentric observer offset + Earth's barycentric state" path the ground
+    stations use: ``_barycentric_moving_observatory`` adds Earth's barycentric
+    state (from layup's own SPICE kernel) to this offset. Referencing the offset
+    against a physical body (Earth) is more reproducible than referencing against
+    the solar-system barycenter, whose definition shifts between ephemeris
+    versions.
+
+    Caveat for high-precision work: Earth's geocenter as realized by layup's
+    SPICE kernel may differ slightly from the geocenter implied by the
+    spacecraft's Horizons reference ephemeris. The geocentric offset cancels
+    Earth's own position to first order, but a small residual inconsistency
+    (~meters) can remain. That is negligible for normal astrometry, but could
+    matter for very high precision applications such as space-based stellar
+    occultations -- flagged here as a known limitation (see PR #377 discussion).
     """
     out = {}
     jd_list = list(jd_tdb_list)
@@ -97,7 +116,7 @@ def _query_chunk(naif_id, jd_chunk, timeout):
         "OBJ_DATA": "NO",
         "MAKE_EPHEM": "YES",
         "EPHEM_TYPE": "VECTORS",
-        "CENTER": "'@399'",  # geocentric (Earth body center)
+        "CENTER": "'@399'",  # geocentric offset (see precision caveat in the docstring)
         "REF_PLANE": "FRAME",  # ICRF equatorial (not the ecliptic)
         "REF_SYSTEM": "ICRF",
         "VEC_CORR": "NONE",  # geometric state at the epoch
--- a/src/layup/utilities/special_observatories.py
+++ b/src/layup/utilities/special_observatories.py
@@ -1,0 +1,151 @@
+"""Resolve space-based ("special case") MPC observatory codes via JPL Horizons.
+
+A handful of MPC observatory codes denote *spacecraft* rather than fixed ground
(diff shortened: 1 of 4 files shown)


src/layup/utilities/special_observatories.py
kjnapier, line 1: Why are we querying the geocentric state of the vectors? That involves another call to get Earth's position, and will lead to potential mismatches in the ephemeris. Might it be better to query the barycentric spacecraft state

kjnapier (commented):
See comment re-Geocentric vs Barycentric queries

src/layup/utilities/special_observatories.py
matthewholman: The code that computes the observatory positions from obscodes has a part that computes the barycentric position of the geocenter and another that computes the geocentric position of the observatory.  This is a plug in for the second part.

kjnapier: I'm a little concerned about possible mismatches between the state vector of the Geocenter in the internal SPICE kernel, vs the state of the Geocenter in the spacecraft's reference kernel. I think it would be best not reference spacecraft against Earth's position, since they are not attached to Earth like the ground-based observatories are. 

Doing this would require changing the pattern for spacecraft

matthewholman: I understand what you are saying, but referencing a position from that of an actual physical object, rather than the barycenter (which changes from ephemeris to ephemeris), is more dependable, I think.

kjnapier: That's probably true; I guess that the perturbers' positions in the simulations are not pulled from the same spice kernel as the spacecraft, so the barycenter will be slightly different. 

I guess using the Geocentric offset should be fine for now, but in the future maybe we will have to be more careful. If, for instance, there are space-based occultation observations, offsets of several meters may matter (right?). If that's true, we should flag it as a potential problem with a comment in the code, and call it good enough for now.

matthewholman: Flagging this as a potential problem for high precision work makes sense.

matthewholman: Done in 0912b50 — added a `Notes` section to `query_horizons_geocentric` explaining the geocentric-offset choice (reproducible offset against Earth vs the ephemeris-dependent barycenter) and flagging the residual ~meter-level geocenter-realization mismatch as a known limitation for very high precision work (e.g. space-based occultations), per this discussion.

matthewholman: This has been done.

kjnapier (approved):
Okay thanks for adding that comment, looks good to me

PR #378, opened 2026-06-30
Add ADES XML observation reader (Closes #44)

### LINKED ISSUE #44: read in ADES ephemeris format file as input for orbit fitting
Current MPC [ADES format](https://www.minorplanetcenter.net/iau/info/ADES.html), in either 
XML or PSV format 

Image

Looks like there are some utilities in here for parsing the xml/PSV version https://github.com/IAU-ADES/ADES-Master/tree/master/Python/bin

matthewholman:
Closes #44.

ADES XML was already a recognized input format, but its reader slot was a `None` placeholder (`"ADES_xml": (None, None)`), so selecting it failed. This implements a native reader so `input_file_format="ADES_xml"` (and the CLI `ADES_xml` type) actually works.

### What it does
- **New `utilities/file_io/ADESXMLReader.py`** — `ADESXMLDataReader` + `parse_ades_xml()`. Streams the file with stdlib `xml.etree.iterparse` (no new dependency), collecting every `<optical>`/`<radar>` record. Handles both the **flat** form (`<ades><optical>…</optical></ades>`) and the **`<obsBlock>`-wrapped** form, plus XML namespaces (localname matching). Each record's child tags are the ADES field names; rows become a structured array whose numeric fields are coerced to float while the **primary-id and station columns stay strings** — i.e. **identical to what `CSVDataReader` produces**, so downstream code is format-agnostic. Implements `_build_id_map`/`obj_id_counts` so `create_chunks`/`read_objects` work.
- Register `ADESXMLDataReader` for `ADES_xml` in orbitfit's format dispatch and export it from `file_io`.

### Why a native reader (vs docs-only)
@mschwamb suggested documenting an XML→PSV conversion instead. We went with a native reader because the `ADES_xml` slot was already reserved (and was a latent crash), the XML is flat and parses with the stdlib, and it makes the documented format work directly with no external tooling. Roving/occultation handling is unchanged.

### Validation
- 11 new tests + 37 existing reader tests pass; black clean.
- **CSV↔XML equivalence test** on a real ADES fixture — proves parity with the established reader.
- **Real-data cross-check** against @mschwamb's actual Holman `3666.xml` (4312 records, parses in ~0.1 s; columns, dtypes, ra/dec ranges, string `stn`/`obsTime` all correct).
- Exercises the exact `orbitfit` instantiation path (`sep=None` + `required_columns`).

### Tests
`tests/layup/test_ADESXMLReader.py` + `tests/data/ades_sample.xml` cover parsing, namespaces, dtypes (`stn` `"024"` stays a string, not `24`), the space-based satellite-position columns (`sys`/`ctr`/`pos1..3`), `get_row_count`, `read_objects`, `create_chunks`, the orbitfit dispatch path, and the CSV↔XML equivalence check. No network.

### Known parity behavior
In real ADES files, older `<optical>` records that carry only `permID` (no `<provID>`) land under `'nan'` when `provID` is the primary id — the **same** behavior the CSV/PSV reader has for a missing id field (consistent, not a regression).

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -46,7 +46,12 @@ from layup.utilities.data_processing_utilities import (
 )
 from layup.utilities.datetime_conversions import convert_tdb_date_to_julian_date
 from layup.utilities.debiasing import debias, generate_bias_dict
-from layup.utilities.file_io import CSVDataReader, HDF5DataReader, Obs80DataReader
+from layup.utilities.file_io import (
+    ADESXMLDataReader,
+    CSVDataReader,
+    HDF5DataReader,
+    Obs80DataReader,
+)
 from layup.utilities.file_io.file_output import write_csv, write_hdf5
 
 logger = logging.getLogger(__name__)
@@ -100,7 +105,7 @@ INPUT_FORMAT_READERS = {
     "MPC80col": (Obs80DataReader, None),
     "ADES_csv": (CSVDataReader, "csv"),
     "ADES_psv": (CSVDataReader, "psv"),
-    "ADES_xml": (None, None),
+    "ADES_xml": (ADESXMLDataReader, None),
     "ADES_hdf5": (HDF5DataReader, None),
 }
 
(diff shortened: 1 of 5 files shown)


kjnapier (approved):
Looks good

PR #385, opened 2026-07-01
updated visualize to fix BCART_EQ bug

Little-Ryugu:
linter removed dead space in file_access utils

Fixes #384 .

Fixed bug that prevented orbital format BCART_EQ from working for visualize. 

## Review Checklist for Source Code Changes

- [x] Does pip install still work?
- [x] Have you written a unit test for any new functions?
- [x] Do all the units tests run successfully?
- [x] Does Layup run successfully on a test set of input files/databases?
- [x] Have you used black on the files you have updated to confirm python programming style guide enforcement?

Issue #384, opened 2026-07-01
layup visualize does not allow 'BCART_EQ' format


Little-Ryugu:
It might good to add this format as it's the default format outputted by orbitfit

Little-Ryugu:
Hang on just double checked the code, its from a small bug. input_format_infer is set as the new orbit format variable but orbit format is still used to check required columns on line 249

https://github.com/Smithsonian/layup/blob/391762bbe3c30d1f0bbf187bcc2cbe57d6ff75b9/src/layup/visualize.py#L230-L249

PR #386, opened 2026-07-01
tests: BCART_EQ regression test for visualize (#384; depends on #385)

matthewholman:
Adds a regression test for #384 (BCART_EQ — orbit-fit's default output format — was not visualizable). This is the first test for `visualize`.

`test_visualize_cli_accepts_bcart_eq` feeds a BCART_EQ orbit file through `visualize_cli` and asserts it reaches the render step. The ephemeris-backed figure build (`build_fig_caches`) and the blocking Dash server (`run_dash_app`) are stubbed, so it exercises only the read / format-inference path in ~1s with no ephemeris needed.

Validated both ways locally: **passes with #385's fix; fails (KeyError on `REQUIRED_COLUMN_NAMES['BCART_EQ']`) without it.**

### ⚠️ Depends on #385
This test asserts the behavior that #385 fixes, so **on `main` as-is it reproduces the bug and CI here will stay red until #385 merges.** Once #385 lands, re-run CI and this goes green. (Kept as a separate PR so #385 stays untouched; merge this after #385.)

### DIFF
--- a/tests/layup/test_visualize.py
+++ b/tests/layup/test_visualize.py
@@ -1,0 +1,42 @@
+"""Regression tests for ``layup visualize`` (issue #384 / PR #385)."""
+
+from layup.utilities.data_utilities_for_tests import get_test_filepath
+from layup.visualize import visualize_cli
+
+
+def _bcart_eq_file_with_objid(src_name, tmp_path):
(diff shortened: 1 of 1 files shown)


Little-Ryugu (approved):
Looks good to me.

PR #387, opened 2026-07-01
visualize: configurable primary-id column via -pid (Closes #383)

### LINKED ISSUE #383: layup visualize requires ObjID column
layup visualize requires an ObjID column to work. It would be good to add a -pid PRIMARY_ID_COLUMN_NAME command line arg similar to other layup commands.

matthewholman:
Closes #383.

`visualize` hard-coded an `ObjID` column, so an orbit-fit result (which uses `provID`) couldn't be visualized without renaming the column. This adds a `primary_id_column_name` argument — CLI **`-pid`/`--primary-id-column-name`**, mirroring `orbitfit`/`predict`.

### What it does
- Threads `primary_id_column_name` through both public entry points (`visualize_cli`, `visualize_notebook`) into **all six reader constructions** (probe / full / special × csv / hdf5) and into `build_fig_caches` (which hard-coded `ObjID` for the special-file dedup).
- Adds the `-pid` CLI flag.

### Default choice: `provID`
Per discussion, the default is **`provID`** — the column `orbitfit` and `predict` write — so `orbitfit → visualize` works with **no flag** (the friction #383/#384 describe). This is a behavior change: `visualize` previously assumed `ObjID`; files keyed by `ObjID` now pass `-pid ObjID`. (The ecosystem is already mixed — orbitfit/predict default `provID`, `convert` defaults `ObjID`.)

The **demo notebook is unaffected**: it uses the numpy-array path with no `special` file, where the id column is never consulted.

### Validation
- New `tests/layup/test_visualize_primary_id.py` (3 tests, ~1s, ephemeris + Dash stubbed): a provID-keyed **BCART_EQ** file (native orbit-fit output) visualizes with no flag; an `ObjID` file works via `-pid ObjID`; a wrong id column raises a clear error. All pass; black-clean; `-pid` shows in `layup visualize --help`.

### ⚠️ Merge-order note vs #386
This flips visualize's default to `provID`, so **#386's** `test_visualize.py::test_visualize_cli_accepts_bcart_eq` (which renames its data to `ObjID` and relies on the default) will need `primary_id_column_name="ObjID"` once both are on `main`. I'll reconcile in whichever PR merges second (a one-line test edit).

### DIFF
--- a/src/layup/visualize.py
+++ b/src/layup/visualize.py
@@ -31,6 +31,7 @@ def build_fig_caches(
     n_points: int = 500,
     r_max: float = 50.0,
     cache_dir: Optional[str] = None,
+    primary_id_column_name: str = "provID",
 ):
     """ """
     # get the assist ephem object and build epoch array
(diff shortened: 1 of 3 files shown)


Little-Ryugu (approved):
Looks good to me!

PR #389, opened 2026-07-01
Don't block on the MPC obscodes download in LayupObservatory (Closes #388)

### LINKED ISSUE #388: CI hang: unguarded MPC ObsCodes.json.gz download (25-retry) wedges test runs
## Symptom
CI Unit jobs intermittently **hang** at "Run unit tests with pytest" for 20–76 min (normal is a few minutes), with the log spamming:
```
Failed to download 'ObsCodes.json.gz'. Will attempt the download again 25 more times.
```
This has wedged many PRs and requires manual cancel+rerun each time. `pytest-timeout` (300s) does **not** catch it.

## Root cause
- The MPC observatory-codes file is fetched from `minorplanetcenter.net` via **pooch with `retry_if_failed=25`** (`bootstrap_utilties/download_utilities.py:make_retriever`), with no short per-request timeout. A slow/flaky MPC turns that into a 25-retry stall of tens of minutes.
- The download is triggered lazily by **Sorcha's `Observatory.__init__`** (`retriever.fetch(observatory_codes_compressed)`) whenever the *decompressed* file isn't already cached — i.e. the first `LayupObservatory()` a test builds if the CI cache missed.
- Because that happens at **test-collection / session-fixture time** (outside a test body), `pytest-timeout`'s per-test signal can't interrupt it → it wedges the whole job.
- `LayupObservatory` already has a bundled fallback, but only *after* the 25 retries exhaust and raise — so it doesn't prevent the long hang.

## Fix
Only use the network obscodes when it's **already cached** (put there by `layup bootstrap`); otherwise hand Sorcha the **bundled copy immediately** — no blocking download on the fit/test path. `layup bootstrap` remains the way to refresh the codes.

Being fixed in the linked PR.

matthewholman:
Closes #388 — the durable fix for the recurring CI hang that's needed cancel+rerun on many PRs today.

### Root cause
Sorcha's `Observatory.__init__` downloads the MPC observatory-codes file via a pooch retriever with **`retry_if_failed=25`** and no short timeout, whenever the *decompressed* file isn't already cached. A slow/flaky MPC turns that into a tens-of-minutes stall. Because the first `LayupObservatory()` is often built at **pytest collection / session-fixture time**, it escapes `pytest-timeout` and wedges the whole job for up to ~76 min (the "Will attempt the download again 25 more times" hang).

### Fix
Decide the obscodes source *before* handing off to Sorcha: only use the network copy when it's **already cached** (put there by `layup bootstrap`); otherwise hand Sorcha the **bundled copy immediately** — so there is never a blocking MPC download on the fit/test path. The existing try/except bundled fallback is kept defensively for a cached-but-corrupt file. `layup bootstrap` remains the way to refresh the codes.

Net effect: a flaky/slow MPC costs ~nothing (falls back to the shipped copy) instead of wedging CI. Observatory codes change rarely, so the bundled copy is a fine default between bootstraps.

### Validation
- 3 new tests in `tests/layup/test_obscodes_fallback.py`, **no ephemeris/network** (SPICE furnishing and Sorcha's `Observatory.__init__` are stubbed), so they run everywhere: `_cached_obscodes_present` detects the cached file; an **uncached** dir makes `LayupObservatory` hand Sorcha the bundled copy (never `None`, which would allow the download); a **cached** file uses the local copy (`oc_file=None`).
- The 12 existing `LayupObservatory` tests still pass; black-clean.

### Scope note
I deliberately did **not** lower `retry_if_failed=25` globally in `make_retriever` — it's shared with the large ephemeris downloads where more retries are reasonable. This targets only the obscodes path that actually wedges. A short per-request timeout on `make_retriever` could be a good follow-up hardening for the `layup bootstrap` path itself.

### DIFF
--- a/src/layup/utilities/data_processing_utilities.py
+++ b/src/layup/utilities/data_processing_utilities.py
@@ -8,6 +8,7 @@ from concurrent.futures import ProcessPoolExecutor
 from importlib.resources import files
 
 import numpy as np
+import pooch
 import requests
 import spiceypy as spice
 from sorcha.ephemeris.simulation_geometry import barycentricObservatoryRates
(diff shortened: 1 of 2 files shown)


kjnapier (approved):
No problem here

PR #390, opened 2026-07-01
tests: fix visualize BCART_EQ test for the provID default (main hotfix)

matthewholman:
**Hotfix for a red `main`.** #387 changed visualize's default primary-id column to `provID`, which broke `test_visualize.py::test_visualize_cli_accepts_bcart_eq` on `main`: that test renamed the fixture's `provID` column to `ObjID` and relied on the old default, so the reader can no longer find the id column (this was the merge-order interaction flagged in #387).

Now that `main` has **both** #385's BCART_EQ fix **and** #387's `provID` default, the fix is to drop the ObjID rename and use the `provID`-keyed fixture directly — which also turns this into the real "a fresh orbit-fit output (provID + BCART_EQ) visualizes with no flag" regression. Removes the now-unused helper.

Verified locally: `test_visualize.py` + `test_visualize_primary_id.py` both pass; black-clean.

### DIFF
--- a/tests/layup/test_visualize.py
+++ b/tests/layup/test_visualize.py
@@ -4,20 +4,7 @@ from layup.utilities.data_utilities_for_tests import get_test_filepath
 from layup.visualize import visualize_cli
 
 
-def _bcart_eq_file_with_objid(src_name, tmp_path):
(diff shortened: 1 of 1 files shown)


kjnapier (approved):
Looks okay to me

matthewholman:
Thanks!

PR #393, opened 2026-07-01
Replaces hard coded ObjID from orbit_maths.py to a variable

### LINKED ISSUE #392: visualise pid bug
With the addition of the -pid command arg I noticed a bug. ObjID is hardcoded into file src/layup/orbit_maths.py. Therefore, ObjID is the only primiary id column that can work for layup visualize currently.

Little-Ryugu:
Fixes #392  .

Changes hard-coded ObjID into pid variable in orbit_maths.py

### DIFF
--- a/docs/notebooks/orbit_visualizer_demo.ipynb
+++ b/docs/notebooks/orbit_visualizer_demo.ipynb
@@ -123,11 +123,7 @@
    "outputs": [],
    "source": [
     "conic_cache, lines_cache, sunpos_cache, _ = prepopulate_orbit_variants(\n",
-    "    orbits,\n",
-    "    orbit_format=\"KEP\",\n",
-    "    input_plane=\"ecliptic\",\n",
-    "    input_origin=\"heliocentric\",\n",
-    "    pid= \"ObjID\"\n",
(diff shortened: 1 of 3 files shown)


matthewholman (commented):
Thanks @Little-Ryugu — this is the right follow-up; it completes the -pid work by getting the hard-coded ObjID out of orbit_maths.py too. A couple of things to tighten up before merge:

  1. pid is currently required (no default), which breaks the visualizer notebook. The docstrings say pid: str, optional, but the signature has no default, so prepopulate_orbit_variants(...) in orbit_visualizer_demo.ipynb (which passes no pid) will hit a TypeError. Notebooks aren't run in the pytest CI, so this stays green while the docs example breaks. Suggest giving it a  default — pid: str = "provID" (matching visualize's default) — and, if easy, placing it after the required data args rather  than as a 3rd positional.
  2. The --special path isn't updated. The special-file call to prepopulate_orbit_variants in visualize.py still doesn't pass  pid, so layup visualize --special … will TypeError on the missing arg. Only the main call got it.

  Minor: REQUIRED_COLUMN_NAMES is now defined both in orbit_maths.py and as a local in visualize_cli — worth de-duplicating. And a small regression test for a non-ObjID file through prepopulate_orbit_variants (and the --special path) would lock both of the above down.

matthewholman (approved):
This looks good to me!

PR #394, opened 2026-07-01
Suppress the cosmetic sbpy TestRunner deprecation warning (Closes #376)

### LINKED ISSUE #376: Suppress AstropyDeprecationWarning (TestRunner) emitted via sbpy on orbitfit/predict
## Summary
Running `orbitfit` or `predict` prints two `AstropyDeprecationWarning`s on import:

```
WARNING: AstropyDeprecationWarning: The TestRunner class is deprecated and may be removed in a future version.
        Use pytest instead. [astropy.tests.runner]
WARNING: AstropyDeprecationWarning: The TestRunnerBase class is deprecated and may be removed in a future version.
        Use pytest instead. [astropy.utils.decorators]
```

These are cosmetic (they don't affect results) but add noise to the CLI output. Reported by @Little-Ryugu in review of #370.

## Root cause
The warning does **not** come from layup's `orbitfit`/`predict` code — neither references `astropy.tests.runner`. It originates in **`sbpy`**: `sbpy/_astropy_init.py` constructs an astropy-style `TestRunner` at import time, which trips astropy's own deprecation of that class. layup pulls `sbpy` in transitively, so importing anything that reaches `sbpy` surfaces it.

Reproduce with just:
```python
import sbpy   # emits both warnings
```
(`import astropy.units` alone does **not**.)

## Options
1. **Suppress in layup** (smallest, fully in our control) — add a targeted filter early in `layup/__init__.py`:
   ```python
   import warnings
   from astropy.utils.exceptions import AstropyDeprecationWarning
   warnings.filterwarnings("ignore", message=".*TestRunner.*", category=AstropyDeprecationWarning)
   ```
   Narrow `message=` so we don't mask unrelated astropy deprecations.
2. **Upstream fix in sbpy** — sbpy drops the deprecated `TestRunner` in `_astropy_init.py` (the real cure; outside this repo). Worth tracking / a heads-up to sbpy, and may resolve on its own with a future sbpy release.
3. **Pin / bump deps** if a newer sbpy already removed it.

Recommendation: do (1) as a one-line cleanup for v1.0 CLI polish, and optionally file/track (2) upstream.

## Notes
- Cosmetic only; not a v1.0 blocker.
- Found during #370 review; see https://github.com/Smithsonian/layup/pull/370#issuecomment-4845520306

matthewholman:
Closes #376.

Running `orbitfit`/`predict` (or importing any layup module that pulls in `sbpy`) printed two `AstropyDeprecationWarning`s about the deprecated astropy `TestRunner`. The source is **sbpy** (`sbpy/_astropy_init.py` builds a `TestRunner` at import) — a transitive dependency, not layup code — so it's pure CLI/notebook noise (flagged by @Little-Ryugu in review of #370).

Installs a **narrow** warnings filter (matches only the `TestRunner` message, so unrelated astropy deprecations still show) at the top of `layup/__init__.py`, before any layup submodule imports sbpy. `import layup` doesn't pull sbpy itself, so the filter is always set first.

**Test** (subprocess, no ephem — runs everywhere): importing `layup.orbitfit` emits no `TestRunner` warning. Verified locally the warning count drops 2 → 0; black-clean.

### DIFF
--- a/src/layup/__init__.py
+++ b/src/layup/__init__.py
@@ -1,3 +1,13 @@
+import warnings as _warnings
+
+# Silence a cosmetic AstropyDeprecationWarning emitted at import time by sbpy
+# (sbpy/_astropy_init.py builds a deprecated astropy TestRunner). sbpy is a
+# transitive dependency of layup, so the warning is pure noise on the layup CLI
+# and in notebooks. The filter is narrow -- it matches only the TestRunner
+# message, so unrelated astropy deprecation warnings are unaffected -- and is
+# installed here, before any layup submodule imports sbpy. See issue #376.
+_warnings.filterwarnings("ignore", message=r".*TestRunner.*")
+
 try:
     from ._version import version as __version__
     from ._version import version_tuple
--- a/tests/layup/test_warning_suppression.py
+++ b/tests/layup/test_warning_suppression.py
@@ -1,0 +1,20 @@
+"""Regression test for the sbpy TestRunner warning suppression (issue #376)."""
+
+import subprocess
+import sys
+
+
+def test_no_sbpy_testrunner_warning_on_import():
+    """Importing a layup module that pulls in sbpy must not surface the cosmetic
+    ``AstropyDeprecationWarning`` about the deprecated ``TestRunner``.
+
+    Run in a fresh subprocess so the check isn't affected by warning state (or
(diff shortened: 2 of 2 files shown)


kjnapier (approved):
Looks fine

PR #396, opened 2026-07-01
Add a throughput benchmark harness (quality #2, part 1)

matthewholman:
Quality initiative #2 (integration + performance), **part 1: a benchmark harness**.

Layup had no benchmark infrastructure — performance was asserted from one-off profiling rather than *measured*. This adds a dependency-free, runnable harness that produces the throughput numbers for the paper and to catch regressions.

### What's here
- **`benchmarks/run_benchmarks.py`** — measures `convert` rows/s (KEP→CART on synthetic orbits) and a full `orbitfit` wall-clock on the bundled demo observations. Prints a table, and `--json` emits machine-readable results.
  - JAX JIT-compiles on the first call, so each timed benchmark **warms up** before measuring (this matters — cold vs warm was ~184 vs ~17,800 rows/s in testing).
  - Any benchmark that errors (e.g. ephemeris missing) is reported as **skipped**; the rest still run.
- **`benchmarks/README.md`** — usage + notes.
- **`tests/layup/test_benchmarks.py`** — smoke tests so the harness can't bit-rot as the APIs it drives evolve. **No performance-floor assertions** (CI machine speed varies → flaky); it only checks the harness runs and reports a positive rate.

### Example output (M-series mac)
```
convert (KEP->CART, 1 worker)                17,806 rows/s   (0.281s, n=5000)
orbitfit (holman_data_working.csv, 1 obj)      0.33 fits/s   (2.997s, n=1)
```

### Design notes
Repo-root `benchmarks/` (not shipped in the package); no `pytest-benchmark` dependency (custom timing keeps it simple); no CI perf gates. Follow-ups noted in the README: a **short-tracklet fits/s** benchmark matching the paper's "fits/sec" figure (the demo file is one big object, so its fits/s isn't the throughput number), and a `predict` rows/s benchmark.

**Part 2** of this initiative (separate PR) is the full-pipeline integration test (`observations → orbitfit → convert → predict`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #397, opened 2026-07-01
Add a full-pipeline integration test (quality #2, part 2)

matthewholman:
Quality initiative #2, **part 2: the full-pipeline integration test**.

The per-feature suites test each stage in isolation — which is exactly how `visualize` shipped unable to read the fitter's own default output (#384). This drives the whole workflow end-to-end through **real compute**, so a hand-off break between verbs can't slip through:

```
observations → orbitfit → convert (KEP) → predict (on-sky)
```

on the bundled demo observations, with the shared default primary-id (`provID`) and no per-stage flags. It asserts at each hand-off:
- **fit** keeps the object id, emits `BCART_EQ` with a finite state + chi-square (i.e. it actually converged);
- **convert** preserves `provID` and lands in `KEP` with sane elements (`e ≥ 0`, finite `a`/`e`/`inc`);
- **predict** yields one row per object per time with `ra_deg`/`dec_deg` in valid ranges.

Ephemeris-gated (`requires_ephem`), single-worker so it runs in-process (no spawned workers). Verified locally end-to-end (~8s).

This completes initiative #2 alongside the benchmark harness (#396): the numerical core was well-tested, but the **cross-verb pipeline** was the gap — now it's covered.

### DIFF
--- a/tests/layup/test_pipeline_integration.py
+++ b/tests/layup/test_pipeline_integration.py
@@ -1,0 +1,73 @@
+"""Full-workflow integration test: observations -> fit -> convert -> predict.
+
+The per-feature suites test each stage in isolation; this drives the whole
+pipeline end-to-end through real compute so a hand-off break between verbs (the
+class of bug behind #384: `visualize` couldn't read the fitter's own output)
+can't slip through. It runs on the bundled demo observations with the shared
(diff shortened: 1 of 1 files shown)


kjnapier (approved):
Looks good

PR #403, opened 2026-07-02
Fail-fast timeout + bounded retries for data downloads (Closes #388)

### LINKED ISSUE #388: CI hang: unguarded MPC ObsCodes.json.gz download (25-retry) wedges test runs

matthewholman:
## Summary

Durable fix for the recurring CI download flake (issue #388): make every layup data download **fail fast** instead of hanging for 20-76 minutes.

The root cause is in `make_retriever` (`bootstrap_utilties/download_utilities.py`): pooch was configured with `retry_if_failed=25` and **no request timeout**, so a stalled MPC/JPL connection could hang long enough to escape `pytest-timeout` when it happened at fixture-collection time, wedging CI runs (and, off-CI, wedging a fit for ~an hour). #389 patched only the observatory-codes path; this addresses the general cause for **every** data download (JPL ephemeris in `orbit_maths`, the meta-kernel kernels, and the `bootstrap` fetch).

## Changes

- **`layup_downloader()`** — a pooch `HTTPDownloader` with a `(connect, read) = (15, 30)s` timeout. The read timeout is the max gap between received bytes, so it bounds a *stalled* connection **without** interrupting a slow-but-progressing large download (the ~600 MB ephemeris keeps flowing, so it is never falsely cut off).
- **`make_retriever`**: `retry_if_failed` **25 -> 3**. Combined with the timeout, a flaky host now fails in about a minute instead of hours.
- Threaded `layup_downloader()` through every fetch site: `download_files_if_missing` (bootstrap), `build_meta_kernel_file` (via a new optional `downloader` arg), and the two ephemeris fetches in `orbit_maths`.
- **New network-free test** (`test_download_utilities.py`) pinning the bounded-retry + fail-fast-timeout policy so it cannot silently regress.

## Notes

- Backward-compatible: `make_retriever`'s signature is unchanged; `build_meta_kernel_file` gains an optional `downloader=None` arg.
- The read timeout only fires on a genuine stall (no bytes for 30s), so legitimate large downloads are unaffected.
- Should end the flake class that has been repeatedly hanging the PR queue (e.g. #398's three consecutive ~2-hour hangs).

Closes #388.

### DIFF
--- a/.github/workflows/testing-and-coverage.yml
+++ b/.github/workflows/testing-and-coverage.yml
@@ -18,6 +18,12 @@ jobs:
         python-version: ['3.13', '3.14']
 
     runs-on: ${{ matrix.os }}
+    # Fail a wedged leg fast instead of running to the 6-hour default. The unit
+    # suite is ~5-6 min warm (~12 min on a cold cache); a leg far exceeding this
+    # is the recurring flaky-download hang (issue #388), which escapes
+    # pytest-timeout when it happens in a spawned worker. Capping keeps reruns
+    # cheap instead of burning ~2 hours per hang.
+    timeout-minutes: 20
     steps:
     - uses: actions/checkout@v7
     - name: Set up - ${{ matrix.os }} - Python ${{ matrix.python-version }}
--- a/src/layup/orbit_maths.py
+++ b/src/layup/orbit_maths.py
@@ -5,7 +5,7 @@ from typing import Literal, Optional
 import numpy as np
 
 from layup.utilities.layup_configs import LayupConfigs
-from layup.utilities.bootstrap_utilties.download_utilities import make_retriever
+from layup.utilities.bootstrap_utilties.download_utilities import make_retriever, layup_downloader
 
 from sorcha.ephemeris.simulation_geometry import equatorial_to_ecliptic
 from sorcha.ephemeris.simulation_constants import ECL_TO_EQ_ROTATION_MATRIX, EQ_TO_ECL_ROTATION_MATRIX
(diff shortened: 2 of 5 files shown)


kjnapier (approved):
Looks good

PR #405, opened 2026-07-02
Don't crash debiasing on blank/unknown star-catalog codes (Closes #401)

### LINKED ISSUE #401: debias() raises KeyError on blank/unknown star-catalog codes (crashes fits on real obs80 data)
### Summary

`debias()` in `src/layup/utilities/debiasing.py` raises `KeyError` when an observation's star-catalog code is blank, `None`, or otherwise not a key in `MPC_CATALOGS`. Because debiasing runs inside the per-object fit loop, this aborts the **entire** fit for any object that contains such an observation — which is common in real MPC data (pre-CCD / uncatalogued astrometry, and any obs where the catalog column is absent so `catalog=None` is passed).

### Root cause

```python
def debias(ra, dec, epoch_jd_tdb, catalog, bias_dict, nside=256):
    catalog_key = MPC_CATALOGS[catalog]        # line 79: unguarded -> KeyError
    if catalog_key not in bias_dict.keys():    # line 80: guard is AFTER the lookup
        return ra, dec
```

The guard on line 80 only handles a *mapped* catalog that is missing from `bias_dict`; it never protects the `MPC_CATALOGS[catalog]` lookup itself. A blank code `''`, `None` (passed by `orbitfit._orbitfit` when no `astCat` column is present, orbit_fit.py ~L913-919), or any unrecognized single-char code therefore crashes.

### Reproduction

```
KeyError: np.str_('')
  File ".../layup/utilities/debiasing.py", line 79, in debias
    catalog_key = MPC_CATALOGS[catalog]
```

Fit any real obs80 dataset containing observations with a blank star-catalog code with `orbitfit(..., debias=True)`. (Hit while fitting historical astrometry — e.g. 1930s NEO obs — in the MPC full-catalog demo.)

### Proposed fix

Treat an unknown/blank/`None` catalog as "no debias correction available" and return the astrometry unchanged, guarding before the lookup:

```python
def debias(ra, dec, epoch_jd_tdb, catalog, bias_dict, nside=256):
    catalog_key = MPC_CATALOGS.get(catalog)
    if catalog_key is None or catalog_key not in bias_dict:
        return ra, dec  # no reference-catalog bias model -> leave astrometry as-is
    ...
```

This matches the intended semantics (bias correction is only defined for observations tied to a known reference catalog) and makes `debias=True` robust on real, catalog-scale data.

### Impact

Blocks `debias=True` on essentially any historical or full-catalog dataset. Low-risk fix (behavior only changes for the currently-crashing cases, which become no-ops).

_Found during Phase 0a of the MPC full-catalog fitting demo._

matthewholman:
## Summary

`debias()` raised `KeyError` when an observation's star-catalog code was blank, `None`, or otherwise not a known catalog name. Because debiasing runs inside the per-object fit loop, that aborted the **entire** object's fit — which happens routinely on real/historical astrometry (pre-catalogue observations, or any observation with no `astCat`).

Found while fitting the MPC catalog with `debias=True`:

```
KeyError: np.str_('')
  File ".../layup/utilities/debiasing.py", line 79, in debias
    catalog_key = MPC_CATALOGS[catalog]
```

## Fix

Guard the lookup with `MPC_CATALOGS.get(...)` and treat an unknown / blank / `None` catalog as "no bias model available", returning the astrometry unchanged. This matches the intended semantics — a bias correction is only defined for a known reference catalog — and mirrors the existing guard just below it for a known catalog that is absent from `bias_dict`.

```python
catalog_key = MPC_CATALOGS.get(catalog)
if catalog_key is None or catalog_key not in bias_dict:
    return ra, dec
```

## Tests

Added regression tests (network-free) covering blank, `None`, and unrecognized catalog codes (all return the astrometry unchanged rather than raising), plus the known-catalog-absent-from-bias_dict case.

Closes #401.

### DIFF
--- a/src/layup/utilities/debiasing.py
+++ b/src/layup/utilities/debiasing.py
@@ -76,8 +76,12 @@ def generate_bias_dict(cache_dir=None):
 
 def debias(ra, dec, epoch_jd_tdb, catalog, bias_dict, nside=256):
 
-    catalog_key = MPC_CATALOGS[catalog]
-    if catalog_key not in bias_dict.keys():
+    # A blank, None, or unrecognized star-catalog code has no bias model (common
+    # in historical / uncatalogued astrometry, or when no astCat column is
+    # present). Leave the astrometry unchanged rather than raising KeyError, which
+    # would otherwise abort the whole object's fit (issue #401).
+    catalog_key = MPC_CATALOGS.get(catalog)
+    if catalog_key is None or catalog_key not in bias_dict:
         return ra, dec
 
     # find pixel from RADEC
--- a/tests/layup/test_debias.py
+++ b/tests/layup/test_debias.py
@@ -70,3 +70,21 @@ def test_debias():
 
     # Check that the returned values are different from the input values (indicating debiasing occurred)
     assert debiased_ra != ra or debiased_dec != dec, "Debiased values should differ from input values."
+
+
+@pytest.mark.parametrize("catalog", ["", None, "NOT_A_REAL_CATALOG"])
+def test_debias_unknown_catalog_returns_unchanged(catalog):
+    """A blank, None, or unrecognized star-catalog code has no bias model and must
+    leave the astrometry unchanged instead of raising KeyError -- which would
+    otherwise abort the whole object's fit on real/historical data (issue #401).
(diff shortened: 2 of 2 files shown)


kjnapier (approved):
Looks good

PR #406, opened 2026-07-02
Add residuals_at_state: fast per-observation residuals at a fixed state

matthewholman:
## Summary

Adds `residuals_at_state(ephem, fit, detections, cov)` — a fast per-observation residuals + mapped-covariance evaluator at a fixed state — and a regression test pinning it against `predict_sequence`.

## Motivation

`predict_sequence` computes each observation's model direction and mapped covariance `B·cov·Bᵀ` by integrating **fresh from the epoch** to that observation — N separate integrations. A full sweep over all observations (e.g. the per-observation χ² that drives Carpino/OrbFit-style outlier rejection) therefore costs **~10× a whole fit** and dominates any reject-refit loop.

## What it does

`residuals_at_state` reuses the fit's own `create_sequences` + `compute_residuals` — **one forward + one backward pass** over the sorted arc, the identical machinery the LM uses each iteration — and returns, per detection:

```
[x_resid, y_resid, oc00, oc01, oc11, n_resid]
```

the raw tangent-plane residual (radians) and the 2×2 mapped state covariance `B·cov·Bᵀ`.

## Results

Measured on a 630-observation arc:

| | predict_sequence | residuals_at_state |
|---|---|---|
| per-obs residual + covariance pass | 2.75 s | **0.026 s** (~105×) |

The resulting per-observation χ² is **identical** to the `predict_sequence` path to a relative 1e-7. End-to-end in an outlier-rejection loop this drops per-object cost ~75–115× (a 3,300-observation NEO: 23 min → 12 s) with bit-for-bit identical rejection decisions.

## Scope / notes

- Optical / astrometry rows only for now — radar and streak partials have a different row layout, so those observations fall back to zero mapped covariance (noted in the C++ comment). Extending to them is straightforward follow-up.
- `residuals_at_state` needs a forward declaration of `create_sequences` (defined later in the same translation unit).
- New regression test `test_residuals_at_state_matches_predict_sequence` fits a real orbit and asserts `residuals_at_state` reproduces `predict_sequence` (same `B·cov·Bᵀ`, same tangent-plane residual).

### DIFF
--- a/src/lib/orbit_fit/orbit_fit.cpp
+++ b/src/lib/orbit_fit/orbit_fit.cpp
@@ -524,6 +524,82 @@ namespace orbit_fit
                                    nongrav_mask, a123);
     }
 
+    // Forward declaration: create_sequences is defined later in this translation
+    // unit (after compute_dX), but residuals_at_state below needs it.
+    void create_sequences(std::vector<double> &times, double epoch,
+                          std::vector<size_t> &reverse_in_seq, std::vector<size_t> &reverse_out_seq,
+                          std::vector<size_t> &forward_in_seq, std::vector<size_t> &forward_out_seq);
+
+    // Per-observation residuals + mapped covariance at a fixed state, computed
+    // with the fit's own efficient forward+backward single-pass integration.
+    // This is a fast replacement for predict_sequence in the outlier-rejection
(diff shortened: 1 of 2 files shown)


kjnapier (approved):
Much better

PR #408, opened 2026-07-02
Benchmark residuals_at_state vs predict_sequence in the harness

matthewholman:
## Summary

Follow-up to the benchmark harness (#396) and `residuals_at_state` (#406): teach the throughput harness to measure the perf win those two enabled.

Adds `bench_residuals` to `benchmarks/run_benchmarks.py`. It fits an orbit, then times the **per-observation residual + mapped-covariance sweep** — the pass that drives Carpino/OrbFit-style outlier rejection — both ways:

- `predict_sequence` integrates **fresh from the epoch to each observation** (N integrations)
- `residuals_at_state` reuses the fit's **single forward + backward pass**

and reports `residuals_at_state` throughput (obs/s) plus its speedup over `predict_sequence`.

## Result

On the `holman_data_working.csv` demo set (4,135 obs), the harness prints:

```
  residuals_at_state (702x vs predict_sequence)   65,121 obs/s   (0.063s, n=4135)
```

(The bigger the arc, the bigger the win — `predict_sequence`'s cost is ~O(N × arc), `residuals_at_state`'s is ~O(arc).)

## Tests

Adds an ephem-gated smoke test (`test_bench_residuals_runs`) asserting the benchmark runs, reports positive `obs/s`, and that the measured speedup exceeds 1 — never a fixed throughput floor (CI machine speed varies), matching the existing benchmark tests' style.

### DIFF
--- a/benchmarks/run_benchmarks.py
+++ b/benchmarks/run_benchmarks.py
@@ -86,7 +86,87 @@ def bench_fit(obs_file="holman_data_working.csv"):
     return {"name": f"orbitfit ({obs_file}, {n_obj} obj)", "n": n_obj, "unit": "fits/s", "seconds": dt}
 
 
-BENCHMARKS = [bench_convert, bench_fit]
+def bench_residuals(obs_file="holman_data_working.csv", reps=3):
+    """Speedup of ``residuals_at_state`` over ``predict_sequence`` for a per-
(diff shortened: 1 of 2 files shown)


kjnapier (approved):
Looks good

PR #410, opened 2026-07-02
Apply debiasing for obs80 catalog codes, not just names (Closes #409)

### LINKED ISSUE #409: debias() silently no-ops on obs80 catalog codes (single-char codes vs MPC_CATALOGS names)
### Summary

`debias()` silently applies **no correction** when the star-catalog is given as an obs80 single-character **code** (e.g. `W`, `L`, `q`) rather than a catalog **name** (e.g. `Gaia3`, `2MASS`, `UCAC4`). Since `Obs80DataReader` supplies the single-char code, **debiasing never fires on obs80 input** — including the bulk MPC observation archive.

### Root cause

`MPC_CATALOGS` maps **name -> code**:

```python
MPC_CATALOGS = {"USNOA1": "a", ..., "UCAC4": "q", ..., "Gaia3": "W"}
```

and `bias_dict` (from `generate_bias_dict`) is keyed by the **codes** (`MPC_CATALOGS.values()`). But `debias()` looks the catalog up as a **key**:

```python
catalog_key = MPC_CATALOGS.get(catalog)   # name -> code
if catalog_key is None or catalog_key not in bias_dict:
    return ra, dec
```

So a **name** works (`"UCAC4"` -> `"q"` -> in `bias_dict`), but a **code** (`"q"`) is not a key -> `.get` returns `None` -> the guard returns the astrometry unchanged. No debiasing.

### Evidence

Fitting the same objects with `debias=False` vs `debias=True` gives **byte-identical** elements, because every obs80 catalog code no-ops:

```
    obj    da(deb=0)    da(deb=1)  changed?
  20000    3.829e-04    3.829e-04        no
  90377    4.078e-02    4.078e-02        no
    944    5.271e-08    5.271e-08        no
```

The obs80 `astCat` values (`L, N, S, U, W, b, c, d, o, q, r, u, v, w`) all match `MPC_CATALOGS` **values**, none match its **keys**.

### Impact

The full-catalogue run ingests **obs80**, so the "faithful to JPL" weighting+debiasing story is incomplete: weighting applies, debiasing does not. (It does not move the headline element agreement much — debias is a sub-arcsec correction and well-observed objects are already tight — but it should actually run, and matters for bias-sensitive objects.)

### Proposed fix

Accept a code as well as a name — `bias_dict` is already code-keyed, so let a code pass through:

```python
catalog_key = MPC_CATALOGS.get(catalog, catalog)   # name -> code, or a code passes through
if catalog_key is None or catalog_key not in bias_dict:
    return ra, dec
```

Names still map to their code; obs80 codes now resolve directly; blank/`None`/unknown still fall through the `not in bias_dict` guard unchanged (preserving #401). Companion to #401.

_Found while enabling debiasing on the MPC full-catalogue fit._

matthewholman:
## Summary

`debias()` silently applied **no correction** on obs80 input: `Obs80DataReader` supplies the single-char catalog **code** (`W`, `q`, …), but `debias()` looked the catalog up as an `MPC_CATALOGS` **key** (a name like `UCAC4`). Since `bias_dict` is keyed by the **code**, a code never resolved → the guard returned the astrometry unchanged. So debiasing never fired on obs80 — including the bulk MPC archive the catalogue fit ingests.

## Fix

```python
catalog_key = MPC_CATALOGS.get(catalog, catalog)  # name -> code, or a code passes through
if catalog_key is None or catalog_key not in bias_dict:
    return ra, dec
```

- **Name** (`"UCAC4"`) → its code `"q"` → in `bias_dict` ✓ (unchanged behaviour for ADES-style input)
- **Code** (`"q"`) → passes through → in `bias_dict` ✓ (obs80 now debiases)
- **blank / None / unknown** → falls through the `not in bias_dict` guard unchanged ✓ (preserves #401)

## Validation

End-to-end, fitting the same objects with `debias=False` vs `debias=True` now **differs** (it was byte-identical before the fix):

| obj | Δa (debias off) | Δa (debias on) |
|---|---|---|
| 20000 | 3.83×10⁻⁴ | 2.16×10⁻⁴ |
| 90377 | 4.08×10⁻² | 1.14×10⁻¹ |
| 944 | 5.27×10⁻⁸ | 6.89×10⁻⁸ |

Corrections are small (sub-arcsec), as expected — but they now actually apply, which the "faithful to JPL" weighting+debiasing story needs on obs80 data.

## Tests

Adds a regression test asserting a code (`"p"`) yields the same correction as its name (`"PPM"`) and actually changes the astrometry. All 8 `test_debias` tests pass.

Closes #409.

### DIFF
--- a/src/layup/utilities/debiasing.py
+++ b/src/layup/utilities/debiasing.py
@@ -76,11 +76,18 @@ def generate_bias_dict(cache_dir=None):
 
 def debias(ra, dec, epoch_jd_tdb, catalog, bias_dict, nside=256):
 
-    # A blank, None, or unrecognized star-catalog code has no bias model (common
-    # in historical / uncatalogued astrometry, or when no astCat column is
-    # present). Leave the astrometry unchanged rather than raising KeyError, which
-    # would otherwise abort the whole object's fit (issue #401).
-    catalog_key = MPC_CATALOGS.get(catalog)
+    # ``catalog`` may be a catalog NAME (an MPC_CATALOGS key, e.g. "UCAC4", as
+    # ADES provides) or the single-char CODE that obs80 supplies (an MPC_CATALOGS
+    # value, e.g. "q"). bias_dict is keyed by the code, so map a name to its code
+    # and let a code pass through -- otherwise debiasing silently no-ops on all
+    # obs80 input (issue #409).
+    #
+    # A blank, None, or unrecognized catalog has no bias model (common in
+    # historical / uncatalogued astrometry, or when no astCat column is present);
+    # it falls through the ``not in bias_dict`` guard and the astrometry is left
+    # unchanged rather than raising KeyError, which would abort the whole object's
+    # fit (issue #401).
+    catalog_key = MPC_CATALOGS.get(catalog, catalog)
     if catalog_key is None or catalog_key not in bias_dict:
         return ra, dec
 
--- a/tests/layup/test_debias.py
+++ b/tests/layup/test_debias.py
@@ -88,3 +88,22 @@ def test_debias_known_catalog_absent_from_bias_dict_returns_unchanged():
     returns the astrometry unchanged (the pre-existing second guard)."""
     ra, dec = 50.0, 10.0
     assert debias(ra, dec, 2451545.0, "PPM", bias_dict={}) == (ra, dec)
+
+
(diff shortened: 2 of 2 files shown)


kjnapier (approved):
Looks good

PR #412, opened 2026-07-02
Actionable error for unknown/stale observatory codes (Closes #404)

### LINKED ISSUE #404: Unknown/stale observatory codes hard-crash the fit (misread as non-fixed observatory)
### Summary

Two related problems in observatory-code handling that make real, catalog-scale astrometry fail:

1. **Staleness.** The bundled/cached `ObsCodes.json` lags the live MPC list. On a machine whose cache dated to Jul 2025, **106** ground stations were missing (added since).
2. **Bad failure mode.** When `populate_observatory` meets an obscode it doesn't know, it assumes a *non-fixed* (satellite/roving) observatory, finds NaN positions, and **hard-crashes the entire object's fit** with a misleading `ValueError: Observatory <code> has invalid coordinates`.

Together: *any* object with an observation from a station added after the cache date dies, with an error that points the user in the wrong direction.

### Reproduction

Fit numbered asteroid **(434) Hungaria** (or essentially any well-observed object with recent astrometry). Its observations include obscodes **R67** and **R83**:

```
ValueError: Observatory R67 has invalid coordinates at epoch <et>: [nan nan nan]
  data_processing_utilities.py:522 in populate_observatory
```

`R67` and `R83` are real ground stations — *Chinon Private Observatory* and *Observatoire Banon La Tuilerie* — present in the current MPC list with full parallax constants (`Longitude/cos/sin`), just absent from the cached file.

### Root cause

`populate_observatory` (`data_processing_utilities.py`, ~L496-522):

```python
coords = self.ObservatoryXYZ.get(obscode, None)
if coords is None or None in coords or np.isnan(coords).any():
    ...
    if obscode in SPACE_OBSERVATORIES and not self._has_ades_position(data):
        ...  # spacecraft
    # else: assume a NON-FIXED observatory that must carry sys/ctr/pos1/pos2/pos3
    for field in ("sys", "ctr", "pos1", "pos2", "pos3"):
        if field not in data.dtype.names: raise ...
    coords = np.array([data["pos1"], data["pos2"], data["pos3"]])
    if coords is None or np.isnan(coords).any():
        raise ValueError(f"Observatory {obscode} has invalid coordinates ...")  # <-- here
```

An unrecognized **ground** code (missing from a stale `ObsCodes.json`) is indistinguishable, in this path, from a genuinely non-fixed observatory. So it falls through to the pos1/pos2/pos3 branch, reads NaN (a normal ground obs has no embedded position), and raises.

### Proposed fixes

- **Graceful degradation for unknown ground codes.** Distinguish "unknown fixed station" from "non-fixed observatory" (the latter is signalled by the data actually carrying `sys/ctr/pos1/pos2/pos3`). For an unknown code with no embedded position, either:
  - emit a clear, actionable error (`obscode R67 not found in the local ObsCodes.json; run 'layup bootstrap' to refresh`), and/or
  - **refresh obscodes on demand** and retry (now safe/bounded via the fail-fast downloader in #403), and/or
  - skip that observation with a warning rather than aborting the whole object.
- **Freshness.** A full-catalog run must be able to refresh the obscodes; make the staleness obvious (the cached file's age / a mismatch against the registry), so users know to `bootstrap`.

### Impact

Blocks fitting any object with post-cache-date astrometry — i.e. most actively-observed objects — and the error message misdirects debugging. Found while fitting the MPC catalog (companion to #401 debias and #402 roving-observer reader).

matthewholman:
## Summary

An observatory code absent from the local `ObsCodes.json` — most often a **ground station added to the MPC list after the cached file** — has no parallax constants, so `populate_observatory` fell into the *moving-observatory* branch, read NaN `pos1/pos2/pos3`, and raised the misleading `Observatory X has invalid coordinates`. The code is valid; only the cached codes file is stale.

## Fix

A **genuine** non-fixed observatory (satellite/roving) carries a real `pos1/pos2/pos3`; NaN there instead means an **unrecognized fixed station**. So on the NaN case, raise a clear, actionable error that names the code and points at `layup bootstrap` to refresh `ObsCodes.json`:

```
Observatory code 'R67' is not in the local observatory-codes file (ObsCodes.json)
and no position was supplied with the observation. If it is a ground station,
refresh the codes with `layup bootstrap` -- the cached list may predate it;
otherwise the code may be invalid.
```

This resolves #404's core complaint (a misleading hard-crash) and makes the staleness obvious and fixable — one of the remedies the issue proposed.

## Scope

This is the clear-degradation fix. Fully *fitting* an object through an unknown code (on-demand obscode refresh, or skip-with-warning) is a larger enhancement; at batch scale the per-object failure bucket now carries an actionable message instead of a misleading one.

## Tests

Adds a regression test: an unknown code with no supplied position raises an error that mentions the code and `bootstrap`, and is not the old "invalid coordinates". All 51 `test_data_processing_utilities` tests pass.

Closes #404.

### DIFF
--- a/src/layup/utilities/data_processing_utilities.py
+++ b/src/layup/utilities/data_processing_utilities.py
@@ -517,9 +517,20 @@ class LayupObservatory(SorchaObservatory):
                         f"The data must have a '{field}' field for non-fixed position observatory {obscode}."
                     )
             coords = np.array([data["pos1"], data["pos2"], data["pos3"]])
-            # If any of the coordinates are None or NaN, raise an error
+            # NaN here means this obscode has no parallax constants in the local
+            # ObsCodes.json AND the observation carried no position. That is almost
+            # always a *ground* station added to the MPC list after the cached
+            # ObsCodes.json -- not a genuinely non-fixed observatory (those carry a
+            # real pos1/pos2/pos3). Give an actionable message pointing at the stale
+            # codes file rather than the misleading "invalid coordinates" (issue #404).
             if coords is None or np.isnan(coords).any():
-                raise ValueError(f"Observatory {obscode} has invalid coordinates at epoch {et}: {coords}")
+                raise ValueError(
+                    f"Observatory code '{obscode}' is not in the local observatory-codes "
+                    f"file (ObsCodes.json) and no position was supplied with the observation. "
+                    f"If it is a ground station, refresh the codes with `layup bootstrap` -- "
+                    f"the cached list may predate it; otherwise the code may be invalid. "
+                    f"(epoch et={et})"
+                )
 
             # Check if the coordinates are in a reference frame that we support.
             if data["sys"] not in ["ICRF_KM", "ICRF_AU"]:
--- a/tests/layup/test_data_processing_utilities.py
+++ b/tests/layup/test_data_processing_utilities.py
@@ -305,6 +305,33 @@ def test_moving_observatory_coordinate_cache():
             observatory.populate_observatory(obscode, et, row_inconsistent)
 
 
+def test_unknown_obscode_gives_actionable_error():
+    """An obscode absent from the local ObsCodes.json with no observation-supplied
+    position -- typically a ground station added to the MPC list after the cached
+    codes file -- must raise a clear, actionable error pointing at `layup bootstrap`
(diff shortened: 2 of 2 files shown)


kjnapier (approved):
All makes sense

PR #414, opened 2026-07-03
Vectorize convert() covariance propagation; unify on one path (~98x)

### LINKED ISSUE #93: optimize layup convert further

matthewholman:
Closes #93 (with #413).

Follow-up to #413 (which vectorized the no-covariance path). This extends `_apply_convert_vectorized` to carry **covariance**, so `convert()` is now vectorized for every case and the codebase is back to a **single conversion path**.

## What

- `_parse_cov_to_bcart_eq` batches the input covariance -> equatorial Cartesian (a rotation congruence for Cartesian inputs; the per-row numba `parse_covariance_row_to_CART` for element inputs, whose Jacobian goes through `universal_cartesian`).
- Output covariance uses **vmapped** `covariance_keplerian_xyz` / `covariance_cometary_xyz` (pure jax) for element outputs and batched block-diagonal rotations for Cartesian outputs.
- Input/output degree<->radian covariance rescaling applied batched.
- The old per-row loop is retained **verbatim** as `_apply_convert_rowwise`, no longer on the production path -- it's the equivalence oracle for the test.

## Performance

| case | before | after | speedup |
|---|---|---|---|
| BCART_EQ + cov -> KEP + cov | 2.2k rows/s | ~227k rows/s | **~98x** |

That's the flagship demo's output stage (fitted orbits **with** covariances -> KEP). Element-input covariances stay per-row (numba Jacobian), same tradeoff as the value path.

## Correctness

`test_convert_vectorized_matches_rowwise` now pins **both the converted state and the propagated covariance** (non-trivial SPD input covariance, all output formats) to `_apply_convert_rowwise` -- matches to ~1e-8. The existing covariance round-trip and `test_convert_covariance_angle_units` (finite-difference Jacobian) tests continue to pass. Full `test_convert.py` green (13 passed).

## Stacking

Stacked on #413 (`perf/convert-vectorized`). Merge #413 first; this then retargets/rebases to `main`.


kjnapier (approved):
Nice

PR #415, opened 2026-07-03
Fix predict() crash for a single prediction time

matthewholman:
### Problem
`predict(..., times=<one time>)` crashes with `TypeError: iteration over a 0-d array`.

Root cause is in `obscodes_to_barycentric` (`src/layup/utilities/data_processing_utilities.py`):

```python
return np.squeeze(np.array(res)) if len(res) > 1 else res[0]
```

For a single observation `len(res) == 1`, so it returns `res[0]` — a **0-d structured scalar**. `predict` → `_predict` then iterates it (`for i, pos in enumerate(obs_pos_vel)`) and raises `iteration over a 0-d array`. Two or more times work, which is why this went unnoticed.

### Fix
```python
return np.array(res)
```
`np.array(res)` already yields shape `(N,)` for any `N >= 1` (the `len > 1` branch's `np.squeeze` was a no-op there), so the special-case was both unnecessary and wrong.

### Why it matters
Predicting at a single epoch is a normal request — e.g. attributing known objects to one visit, which is exactly what the downstream ssolink known-object identification does per exposure.

### Test
Adds `test_predict_single_time`: a length-1 `times` returns one row per object (no crash), the positions are finite, and they match the earliest epoch of a two-time `predict` per object.

### DIFF
--- a/src/layup/utilities/data_processing_utilities.py
+++ b/src/layup/utilities/data_processing_utilities.py
@@ -780,7 +780,7 @@ class LayupObservatory(SorchaObservatory):
             res.append(np.array((x, y, z, vx, vy, vz), dtype=output_dtype))
 
         # Combine all of our results into a single structured array
-        return np.squeeze(np.array(res)) if len(res) > 1 else res[0]
+        return np.array(res)
 
 
 def get_format(data):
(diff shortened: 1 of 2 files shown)


kjnapier (approved):
Looks good

PR #416, opened 2026-07-04
Warn (not crash) on same-epoch observer-position repeats across co-observed objects

matthewholman:
## Problem

`LayupObservatory` caches a moving/space-based observer's geocentric position per `(obscode, epoch)`, and **that cache is shared across every object it processes** (both here and in `orbitfit`, which builds one `LayupObservatory` per chunk). A repeat position at the same key had to match **exactly** (`np.allclose` default) or it raised `ValueError` and crashed the fit.

Space telescopes image **many asteroids in a single exposure**, so different objects each carry a record for the **same (observatory, epoch)**. The satellite's one true position at that instant is reported per record at finite (~km) precision, so those records differ slightly. Surfaced by the MPC full-catalog fit:
- **C57 = TESS** (~lunar distance): same-epoch positions differ by ~1.4 km
- **C51 = WISE** (low Earth orbit): differ by up to ~4 km

(This is *cross-object*, not two observations of one object — fitting either object alone never collides; it only appears once a shared observatory cache accumulates entries from co-observed objects.)

## Fix

The colliding positions are the **same body at the same instant**, astrometrically indistinguishable at the km level (1 km at 1 au ≈ 1.4 mas, far below observational noise), and a genuinely corrupt observer position (unit/frame error) shows up as a diverging fit anyway. So **warn instead of raise**, keeping the most recent position. Only a difference beyond any plausible reporting spread — `_OBSERVER_POSITION_WARN_KM = 100 km` — is logged (a likely units/frame error). Never fatal.

## Impact

~0.07% of numbered objects (all TESS/WISE-observed) crashed on this in the 10k-object sample; at full catalog scale that's ~600 objects — now fit. Confirmed the previously-crashing objects (39799, 79671, f1997) now converge (flag=0, rel|da| ~1e-9).

## Test

`test_moving_observatory_same_epoch_position_not_fatal` (uses the C57/TESS values): a few-km same-epoch difference must not raise **or** warn; a >100 km difference is logged but still does not raise. The former hard-raise assertion in `test_moving_observatory_coordinate_cache` is updated to the not-fatal behaviour. Full `test_data_processing_utilities.py` green (51 passed).

### DIFF
--- a/src/layup/utilities/data_processing_utilities.py
+++ b/src/layup/utilities/data_processing_utilities.py
@@ -26,6 +26,16 @@ from layup.utilities.special_observatories import (
 
 """ A module for utilities useful for processing data in structured numpy arrays """
 
+# Two observations at the same instant from the same moving/space-based observer
+# should report the same geocentric position. MPC 80-column satellite records
+# (and some ADES sources) carry that position at finite precision, so repeats at
+# one epoch can differ at the ~0.01 km reporting level. Treat positions agreeing
+# to this absolute tolerance as consistent: 1 km at 1 au is ~1.4 mas of
+# astrometric shift, far below observational noise, while gross unit/sign/center
+# errors differ by orders of magnitude more and still raise. (Surfaced by
+# object 47451, obscode C57 = TESS, in the MPC full-catalog fit.)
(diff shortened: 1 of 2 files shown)


kjnapier (approved):
Really good

PR #417, opened 2026-07-04
Adaptive non-grav inclusion: fit_nongrav="auto" (issue #357)

### LINKED ISSUE #357: Adaptive non-grav inclusion: fit_nongrav="auto" (only when warranted)
Follow-up to #351 / #356 (which add the ability to *fit* non-grav A1/A2/A3).

Standard orbit-determination practice is to introduce a non-grav parameter only when a gravity-only fit is unacceptable **and** the parameter is statistically warranted. Today that decision is manual: `fit_nongrav` is an explicit, off-by-default flag, and when set we adopt the requested params as long as the joint fit converges and is well-conditioned (the weak-constraint guard catches *can't constrain*, not *not needed*).

Add an adaptive `fit_nongrav="auto"` mode (mirroring `iod="auto"`):

1. Gravity-only fit; if reduced chi-square is acceptable, stop (no non-gravs).
2. Otherwise try non-gravs (A2 first; A1/A3 as the arc warrants), adopting one only if it (a) converges, (b) passes the conditioning guard, and (c) clears a significance / chi-square-improvement test (F-test or chi2 drop, plus a per-param |A|/sigma cut).
3. Return the most parsimonious acceptable model.

Building blocks already exist from #356 (gravity-only fit, joint refinement, the correlation-matrix conditioning guard, and formal parameter uncertainties); this issue is just the decision logic on top.

Related: COM/KEP joint covariance for non-grav fits, and seeding non-gravs from a supplied initial_guess (both noted as #356 follow-ups).

matthewholman:
Closes #357.

Follow-up to #351/#356 (which added the ability to *fit* non-grav A1/A2/A3). Adds an adaptive `fit_nongrav="auto"` mode, mirroring `iod="auto"`: introduce a non-grav parameter only when a gravity-only fit is unacceptable **and** the parameter is statistically warranted.

## Behaviour

`orbitfit(..., fit_nongrav="auto")`:
1. Keep the gravity-only fit if its reduced chi-square is acceptable (`_gravity_fit_acceptable`, ≤ 1.5).
2. Otherwise try non-grav models of increasing complexity — **A2; A1+A2; A1+A2+A3** — and adopt the first that:
   - converges and is well-conditioned (`flag == 0`, i.e. passes the #356 weak-constraint guard),
   - drops chi-square by more than ~3σ per added parameter (Δχ² > 9·k), **and**
   - has every added parameter individually significant (`|A| > 3·σ_A`).
3. Return the **most parsimonious** acceptable model.

The result schema always carries `a1/a2/a3` (+ `_unc`); each row reports only the adopted params, the rest `NaN`. Reuses the existing joint fit, conditioning guard, and formal uncertainties — this PR is just the decision layer on top.

## Tests

The decision layer is factored into pure helpers (`_gravity_fit_acceptable`, `_nongrav_warranted`, `_select_nongrav_auto`) and unit-tested with crafted fit results in **`test_nongrav_auto.py`** (7 tests, no ephemeris needed): the acceptance gate, the χ²-drop + per-parameter significance test, the parsimony ladder (adopt A2 / escalate to A1+A2), and the ill-conditioned fallback. `test_nongrav_a2.py` adds the end-to-end `orbitfit()` 'auto' schema + keep-gravity-when-not-warranted path.

*Note on test design:* the existing synthetic arcs are noise-free, where a gravity fit already has ~0 χ² and so is (correctly) statistically acceptable — adopting a param there would be over-fitting. The adopt/escalate logic is therefore validated with realistic crafted χ²/σ values in the unit tests rather than noise-free arcs.

## Catalog relevance

Non-gravs matter for the NEO tail of the MPC full-catalog fit; `auto` lets a bulk run apply them only where warranted, without a per-object manual flag.

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -157,6 +157,58 @@ def _parse_nongrav(fit_nongrav):
     return sum(_NONGRAV_BITS[n] for n in names), names
 
 
+# fit_nongrav="auto" decision thresholds (issue #357). Non-grav models are tried
+# only when the gravity-only reduced chi-square exceeds _AUTO_ACCEPT_REDUCED_CHI2,
+# in increasing complexity (_AUTO_NONGRAV_LADDER); the first model that converges
+# and is well-conditioned (flag 0), drops chi-square by more than
+# _AUTO_DELTA_CHI2_PER_PARAM per added parameter, and whose every added parameter
+# is individually significant (|A| > _AUTO_NSIGMA * sigma) is adopted. Otherwise
+# the gravity-only fit is kept. Standard practice: introduce a non-grav only when
+# gravity is unacceptable and the parameter is statistically warranted.
+_AUTO_NONGRAV_LADDER = (("A2",), ("A1", "A2"), ("A1", "A2", "A3"))
+_AUTO_ACCEPT_REDUCED_CHI2 = 1.5
+_AUTO_DELTA_CHI2_PER_PARAM = 9.0  # ~3-sigma chi-square drop per added parameter
+_AUTO_NSIGMA = 3.0
+
+
+def _gravity_fit_acceptable(csq, ndof):
+    """Whether a gravity-only fit is good enough that no non-grav is warranted.
+
+    True when the reduced chi-square is at or below the acceptance threshold (or
+    there are no degrees of freedom to judge it by).
+    """
+    return ndof <= 0 or csq / ndof <= _AUTO_ACCEPT_REDUCED_CHI2
+
+
+def _nongrav_warranted(csq_gravity, res_ng, names):
+    """Whether adopting the (converged, well-conditioned) non-grav fit ``res_ng``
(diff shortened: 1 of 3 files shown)


kjnapier (approved):
Okay. Might be nice to make the threshold tunable, but no big deal

matthewholman:
Good point, @kjnapier.

matthewholman:
Made the thresholds tunable in 5c27ddf, @kjnapier. The three `fit_nongrav="auto"` decision knobs are now bundled in a frozen `NongravAutoThresholds` dataclass (defaults unchanged: acceptance reduced-chi2 1.5, chi-square drop per added parameter 9.0, per-parameter significance 3-sigma) and threaded through as an optional `orbitfit(..., nongrav_auto_thresholds=...)`. Pass a customized instance to tune how readily a non-grav is adopted; omit it for the current behavior. Added unit tests for each knob.

Issue #419, opened 2026-07-04
Incremental / steady-state catalog updates (skip-unchanged, warm-start, sequential update)


matthewholman:
## Motivation

In steady state, most observations of an object are already in a previous fit and only a small fraction of objects receive new astrometry each cycle (a nightly/weekly MPC delta). Re-fitting the entire catalog from scratch every cycle is wasteful. layup should support **incremental / steady-state catalog updates**: fit the catalog once, then each cycle touch only what changed.

## Three levers (in increasing effort / payoff)

**1. Skip unchanged objects (exact; the big win).**
Store an *obs fingerprint* in the fit result -- e.g. a hash of the observations used (count + hash of the sorted records). On a new cycle, an object whose fingerprint is unchanged has an identical fit -> carry the prior forward, don't refit.
- *Needs:* a fingerprint column in the orbitfit result schema + an incremental driver that loads a prior catalog and diffs by fingerprint. Python-only.

**2. Warm-start the refit from the prior fit (exact; ~free).**
For changed objects, seed the LM from the prior fit's state rather than from IOD/MPCORB. `orbitfit` already accepts a warm-start `initial_guess`, so this is mostly plumbing. Reduces LM/rejection iterations (and their IAS15 ramp-ups) -- marginal for well-behaved objects (they converge in ~1 iteration cold), but real for poorly-conditioned ones, and costs nothing.

**3. True sequential / information-filter update (research extension; big per-object win).**
For an object with thousands of old obs and a few new ones, (1)+(2) still re-integrate all old obs. A sequential update instead uses the prior covariance as a prior (information matrix `Lambda = C_prev^-1`) and processes **only the new observations**, never re-touching the old ones -- how production OD (JPL/MPC) actually runs.
- *Needs:* a C++ prior-information term in the LM core (add `Lambda` to the normal equations; residuals/partials for new obs only), obs-level diffing (which obs are new), rejection with a prior, and validation against a full re-fit + a nonlinearity fallback gate (fall back to full refit when the update moves the state too far).
- *Caveat:* it's a linearized update -- valid for routine tracking, not for an orbit-changing recovery.

## Prototype evidence

[...]

PR #420, opened 2026-07-04
Incremental fits: obs fingerprint + skip-unchanged (Phase A of #419)

matthewholman:
## Summary

Phase A of incremental / steady-state orbit determination (issue #419, levers 1 + 2). A catalog re-run now fits only the objects whose observations changed since the prior catalog; unchanged objects are carried forward verbatim. This is the throughput mechanism behind the steady-state MPC-catalog demo. Lever 3 (the true sequential / information-filter update, C++) will follow as a separate PR.

## What changed

- **Obs fingerprint in the fit record.** `_get_result_dtypes` gains `obs_hash` + `nobs_fit`, appended after the conditional non-grav columns so the existing positional output tuples are unaffected. Every fit now carries a deterministic fingerprint of the observation set it was built from (`""` on failed/invalid fits, so they are always retried).
- **`_obs_fingerprint()`** — SHA-1 over the fit-relevant columns (`_FINGERPRINT_COLUMNS`), computed on the **raw, pre-debias** observations and order-independent (row-strings are sorted before hashing). Deterministic across runs and processes.
- **`orbitfit(skip_unchanged=True)`** — before any ephemeris/observatory setup, `_partition_unchanged()` drops objects whose fingerprint matches a converged prior fit and carries their stored result forward, so a skipped object costs one hash rather than a fit. Changed objects fall through and are warm-started from the prior state via the existing `initial_guess` path (lever 2, already present).
- A per-object skip gate in `_orbitfit` backstops lower-level callers.
- Falls back to fitting when the prior catalog lacks the fingerprint columns (older catalogs), so it is backward compatible.

## Validation

- Changing one observation of one object in a 6-object set cuts the re-run **7.76s → 2.18s (~3.6×)** here; at a realistic 1-5% nightly change rate the saving is ~20-100×.
- The 4 unchanged objects are **byte-identical** to their prior fits; warm-start reproduces the cold answer to 2.7e-15.

## Tests

10 new tests in `tests/layup/test_incremental_fit.py`: fingerprint determinism / order-independence / sensitivity, carry-forward byte-identity, changed-obs refit, warm-start equivalence, and legacy-catalog fallback. `test_orbit_fit.py` and `test_unpack.py` still pass (the schema change is additive).

Part of #419.

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -1,3 +1,4 @@
+import hashlib
 import logging
 import os
 import re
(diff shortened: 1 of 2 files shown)


kjnapier (approved):
Nice

PR #421, opened 2026-07-05
Sequential / information-filter orbit update (Lever 3 of #419)

matthewholman:
## Summary

Lever 3 of incremental orbit determination (issue #419): a **true sequential / information-filter update**. Refine a prior orbit fit using **only new observations**, folding the previously-fit observations in through their Gaussian summary (the prior state + 6x6 covariance). To the extent the prior is locally Gaussian this equals a full batch refit over all observations, at the cost of integrating only the new observations — the pattern operational OD at JPL/MPC uses to keep a large catalog current.

Stacks conceptually on #420 (Phase A: skip-unchanged + warm-start) but is **independent** — this PR is off `main` and touches the fit core, not the Phase A code.

## What changed

**C++ (`src/lib/orbit_fit/orbit_fit.cpp`)**
- `compute_dX` gains an optional prior-information term: `C += Λ₀`, `grad += Λ₀·(x − x₀)` with `Λ₀ = P₀⁻¹`. Because `B`/`r` are assembled over the **new obs only**, `C = BᵀWB + Λ₀` is exactly the batch normal matrix of the full (old-summarised + new) problem, and the existing `C.inverse()` yields the posterior covariance. **A null prior reduces to ordinary least squares — the existing fit path is byte-unchanged** (the 22 base-fit tests still pass).
- `orbit_fit` snapshots the prior mean `x₀`, forms `(x − x₀)` each iteration, and folds the prior quadratic into the LM gain ratio (the reported chi-square stays data-only). With a prior the `npar` parameters are constrained by the prior, not consumed from the new obs, so the reduced chi-square is assessed over the new-obs rows.
- New `run_sequential_update` entry point + pybind binding: takes a prior `FitResult` and the new observations, forms `Λ₀` via a Cholesky solve, pins the epoch to the prior's, and returns the updated state + posterior covariance. **SRIF-ready boundary** — the prior is passed as a covariance, so a future square-root-information-filter variant can carry the Cholesky factor through the same signature.

**Python (`src/layup/orbitfit.py`)**
- `sequential_update()` driver: builds `Observation`s for the new obs (optical astrometry + streak), runs the C++ update, and applies a **nonlinearity gate** — the Mahalanobis size of the update in prior standard deviations. A too-large move means the linearization of the summarised old obs is untrustworthy, so it falls back to a full refit over `all_data` (or flags the result, flag=8, when no full set is supplied).

## Validation

On a well-observed object (587 obs), the sequential update reproduces a full batch refit to **state rel ~1e-13 (machine precision)** and covariance ~1e-6, and the agreement **tightens monotonically as the prior tightens** — the signature of a correct information filter whose only error is the linearization of the summarised observations. 4 tests in `test_sequential_update.py` (matches-full-refit, covariance tightening, gate fallback, gate flag); 22 base-fit tests confirm the ordinary fit path is unchanged.

## Follow-ups (separate PRs)

- **Outlier rejection on the new observations** in the sequential path (the full-refit fallback already carries the JPL/OrbFit rejection).
- **Steady-state catalog integration** — obs-level diffing and the per-object loop tying #420's skip/warm-start to this sequential update.
- **Broad-regime validation** — NEO / MBA / TNO / hyperbolic priors, and the conditioning breakdown that would motivate the SRIF upgrade.
- Non-grav priors, radar/occultation in the sequential path, and differing-epoch priors (state-transition propagation of `P₀`).

Part of #419.

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -19,6 +19,7 @@ from layup.routines import (
     run_bk_iod,
     run_bk_native_fit,
     run_from_vector_with_initial_guess,
+    run_sequential_update,
 )
 
 try:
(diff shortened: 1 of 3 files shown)


kjnapier (approved):
Very cool

PR #422, opened 2026-07-05
Incremental steady-state fit driver (capstone of #419)

### LINKED ISSUE #419: Incremental / steady-state catalog updates (skip-unchanged, warm-start, sequential update)

matthewholman:
## Summary

The capstone of incremental orbit determination (issue #419): `incremental_orbitfit`, a single steady-state maintenance pass over a batch of objects that ties the three levers together. For each object it routes:

- **skip** — the observation set is unchanged since `prior_catalog` (matching fingerprint): carry the prior fit forward verbatim, no fit.
- **sequential update** — observations were only appended (`prior_obs` given and every prior observation still present): update via `sequential_update` over the **new observations only**. Its nonlinearity gate falls back to a full refit when the update is too large.
- **full refit** — observations removed or changed, or no per-object `prior_obs`: refit over all current obs, warm-started from the prior state.
- **cold** — no prior for the object: cold IOD fit.

Returns the updated result catalog (same schema as `orbitfit` output, carrying the `obs_hash`/`nobs_fit` fingerprint so it seeds the next cycle) plus a routing tally.

**Obs-level diffing** (`_obs_row_keys` / `_append_only_new_obs`) reuses #420's per-row fingerprint strings, so a row's diff key equals its contribution to the object's fingerprint — append-only vs modified is detected exactly.

## ⚠️ Stacking — review order

This is one feature in three parts and **stacks on #420 and #421**:

- **#420** — obs fingerprint + `orbitfit(skip_unchanged=True)` (the skip route + schema).
- **#421** — the sequential / information-filter update (`run_sequential_update` + `sequential_update`).

This branch merges both, so the diff below **includes** them. Merge #420 and #421 first; this PR's net-new code is a single commit, **`091f804`** (`incremental_orbitfit` + `tests/layup/test_incremental_driver.py`) — review that commit directly, and I'll rebase once the two land so the diff reduces to just the driver.

## Validation

- **5 route tests** in `test_incremental_driver.py` (skip byte-identity, sequential-matches-full-refit, modified-obs→full, cold, no-prior-obs→full). Full suite of #420 + #421 + this + base fit tests: **31 passed**.
- **Real MPC data** (`mpc_demo/cycle_rehearsal_seq.py`, a steady-state cycle on numbered asteroids): the sequential route matches a full refit to **state rel ~1e-8 / KEP a ~1e-9**, at **6.5–13.7× less wall-clock on ~1000-obs arcs** (only the new obs are integrated; the win approaches the integration-work ratio as the arc lengthens). The nonlinearity gate never tripped on well-observed objects with small nightly additions — its job is the pathological cases.

Combined with the skip route, the steady-state loop is: 95–99% of objects skip (zero cost), changed objects sequential-update at ~7–13× less work, and only non-append or nonlinear changes fall back to a full refit — the operational maintenance loop JPL/MPC run.

## Follow-ups (unchanged from #421)

Outlier rejection on the new observations in the sequential path; non-grav priors; radar/occultation in the sequential path; differing-epoch priors. A CLI wrapper for `incremental_orbitfit` can follow once the API settles.

Part of #419.

Closes #419

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -1493,6 +1493,193 @@ def sequential_update(
     return seq
 
 
+def _obs_row_keys(data, column_names):
+    """Per-observation identity keys over the fit-relevant columns (issue #419).
+
+    Each key is the same per-row string that ``_obs_fingerprint`` hashes, so a
+    row's key equals its contribution to the object's fingerprint. Used to diff a
+    current observation set against the one a prior fit was built from.
+    """
+    cols = [c for c in _FINGERPRINT_COLUMNS if c in column_names]
+    return ["\x1f".join(_fmt_fingerprint_value(d[c]) for c in cols) for d in data]
+
+
+def _append_only_new_obs(current, prior_obs):
+    """Return the rows of ``current`` absent from ``prior_obs`` if the change is
+    append-only, else ``None``.
+
+    Append-only means every observation the prior was fit from is still present
+    (none removed or modified) -- the case the sequential update handles exactly.
+    If any prior observation is gone, the summarised old-obs information no longer
+    matches the current set, so the object must be fully re-fit and this returns
+    ``None``.
+    """
+    cur_keys = _obs_row_keys(current, current.dtype.names)
+    prior_keys = set(_obs_row_keys(prior_obs, prior_obs.dtype.names))
+    if not prior_keys.issubset(cur_keys):
+        return None  # an old observation was removed or changed -> not append-only
+    mask = np.array([k not in prior_keys for k in cur_keys], dtype=bool)
+    return current[mask]
+
+
+def _group_by_id(data, primary_id_column_name):
+    if data is None:
+        return {}
+    ids = data[primary_id_column_name]
+    return {oid: data[ids == oid] for oid in np.unique(ids)}
+
+
+def _fitresult_to_row(fit, obj_id, obs_hash, nobs_fit, dtypes):
+    """Pack a FitResult (from the sequential update or its refit fallback) into one
+    result-catalog row carrying the current-obs fingerprint, so the row is
+    identical in shape to ``orbitfit`` output and can seed the next cycle."""
+    success = fit.flag == 0
+    cov = tuple(fit.cov[i] for i in range(36)) if success else (np.nan,) * 36
+    row = (
+        (obj_id, (fit.csq if success else np.nan), fit.ndof)
+        + (tuple(fit.state[i] for i in range(6)) if success else (np.nan,) * 6)
+        + (
+            (fit.epoch - 2400000.5) if success else np.nan,
+            fit.niter,
+            fit.method,
+            fit.flag,
+            "BCART_EQ" if success else "NONE",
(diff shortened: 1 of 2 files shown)


kjnapier (approved):
Looks good to me

PR #423, opened 2026-07-05
Fix debias: years-since-J2000 epoch + nested healpix ordering

matthewholman:
## Summary

Star-catalog debiasing was applying **~10 arcsec** corrections instead of the correct **sub-arcsec** catalog biases. Two bugs in `debias()`:

1. **Epoch reference** — the proper-motion term used the raw JD (`dt = jd/365.25` ≈ **6700 yr**) instead of years since J2000 (`dt = (jd − 2451545)/365.25`), extrapolating the proper motion over millennia.
2. **Healpix ordering** — the Farnocchia/Eggl bias tables are stored **NESTED**, but the pixel lookup used **RING** (`nest=False`), pulling the bias from the wrong sky pixel.

## Impact

With debiasing **on**, every catalogued observation was shifted ~10″, so the reject/refit loop culled ~half of them as gross outliers — fits tripped `flag=2` (reduced χ² in the hundreds) and accuracy degraded. Found at **MPC full-catalog scale on USDF** (82% of fits flag=2). After the fix, debias-on rejects the same ~1% as debias-off and *improves* accuracy, as debiasing should:

| object | broken debias-on | **fixed debias-on** | debias-off |
|---|---|---|---|
| 20000 | 472/945 rejected | **2/945**, rel\|da\| 4.0e-6 | 8.9e-6 |
| 50000 | 383/766 rej, rel\|da\| 7.8e-5 | **6/766**, 4.8e-6 | 5.7e-6 |
| 30000 | 2140/4280 rejected | 344/4280, **8.6e-9** | 1.19e-8 |

## Tests

`test_debias` only asserted the correction was nonzero (not its magnitude), so it missed this. Added:
- a **synthetic one-pixel** regression test (no data files) that pins *both* the J2000 epoch and the nested lookup — 1″/yr of proper motion sampled 10 yr after J2000 must yield exactly 10″;
- a real-table check that a modern-epoch correction is sub-arcsec.

All 10 debias tests pass.

### DIFF
--- a/src/layup/utilities/debiasing.py
+++ b/src/layup/utilities/debiasing.py
@@ -91,8 +91,10 @@ def debias(ra, dec, epoch_jd_tdb, catalog, bias_dict, nside=256):
     if catalog_key is None or catalog_key not in bias_dict:
         return ra, dec
 
-    # find pixel from RADEC
-    idx = hp.ang2pix(nside, ra, dec, nest=False, lonlat=True)
+    # find pixel from RADEC. The Farnocchia et al. (2015) / hires-2018 bias tables
+    # are stored in NESTED healpix ordering, so a RING lookup pulls the bias from
+    # the wrong sky pixel.
+    idx = hp.ang2pix(nside, ra, dec, nest=True, lonlat=True)
 
     # Retrieve the offsets and proper motions from the bias dictionary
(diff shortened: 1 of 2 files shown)


kjnapier (approved):
Good catch

PR #424, opened 2026-07-05
Persistent disk cache for space-observatory Horizons lookups

matthewholman:
## Summary

Space-based observatories (WISE, HST, JWST, Spitzer, ...) have no fixed parallax constants, so layup resolves their geocentric observer state from **JPL Horizons per (naif_id, epoch)** — cached only *in-memory per `LayupObservatory` instance* (added in #377). At MPC full-catalog scale, every parallel worker re-queries Horizons cold, and JPL rate-limits to **HTTP 503** (~0.2% of numbered objects crash on it during the USDF run).

This adds a **persistent on-disk cache** so each `(spacecraft, epoch)` is fetched from Horizons **once, ever**, then served from disk across objects, processes, and runs.

## What changed

- `query_horizons_geocentric(naif_id, jds, cache_dir=...)` checks `<cache_dir>/horizons/<naif_id>/<jd:.6f>.npz` per epoch, queries Horizons only for the **misses**, and stores results atomically (`tempfile.mkstemp` + `os.replace`; distinct epochs are distinct files → lock-free and safe on a parallel/shared filesystem). States are geometric (`VEC_CORR='NONE'`), hence deterministic and exact to cache. Cache read/write failures (missing **or partial/corrupt** file) fall through to a refetch and never break a fit.
- `LayupObservatory` now stores its `cache_dir` and threads it into `_fetch_space_observatory_states`, so the observatory path — and the full-catalog run — cache under the same cache directory automatically.
- **Default off** for direct callers with no `cache_dir` and no `LAYUP_HORIZONS_CACHE` env, preserving existing behavior. Opt in via `cache_dir` or the env var; `cache_dir=False` / `LAYUP_HORIZONS_CACHE=0` disables.

## Tests

6 new (network mocked at `_query_chunk`): write-then-serve-from-disk, partial-overlap-fetches-only-misses, per-spacecraft isolation, disabled-via-arg, disabled-via-env, corrupt-file-refetch. The existing space-observatory + data-processing suites still pass (68 total).

## Provenance

Found and validated at MPC full-catalog scale on Rubin/USDF (the 503s were tripping ~0.2% of the numbered run). The USDF session independently built a harness-side monkeypatch cache and reconciled it to this layout (same `.6f` key, same atomic write, same corrupt-read → refetch), so cache already warmed there is reusable as-is; once this merges the monkeypatch is dropped and predict/orbitfit/nightly all benefit.

Extends #377 / issue #55. SPK kernels for the 9 spacecraft (network-free) are a sensible follow-up.

### DIFF
--- a/src/layup/utilities/data_processing_utilities.py
+++ b/src/layup/utilities/data_processing_utilities.py
@@ -355,6 +355,10 @@ class LayupObservatory(SorchaObservatory):
         # Get Layup configs
         config = LayupConfigs()
 
+        # Kept so space-observatory Horizons lookups land their persistent
+        # (naif_id, jd) -> state cache under the same cache directory.
+        self.cache_dir = cache_dir
+
         # Furnish the spiceypy kernels
         layup_furnish_spiceypy(cache_dir)
 
@@ -629,9 +633,11 @@ class LayupObservatory(SorchaObservatory):
         """Geocentric states of a spacecraft at the given JD(TDB) epochs.
 
         Thin wrapper around :func:`query_horizons_geocentric` -- a seam that
-        tests monkeypatch so they never touch the network.
+        tests monkeypatch so they never touch the network. The persistent
+        (naif_id, jd) -> state cache is keyed under this observatory's cache
+        directory, so it is shared across objects, processes, and runs.
         """
-        return query_horizons_geocentric(naif_id, jd_tdb_list)
+        return query_horizons_geocentric(naif_id, jd_tdb_list, cache_dir=self.cache_dir)
 
     def _populate_space_observatory(self, obscode, et, obscode_cache_key):
         """Resolve a space-based observatory's geocentric state via JPL Horizons.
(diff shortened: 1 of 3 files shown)


mschwamb:
I can't parse this English "Found and validated at MPC full-catalog scale on Rubin/USDF (the 503s were tripping ~0.2% of the numbered run). The USDF session independently built a harness-side monkeypatch cache and reconciled it to this layout (same .6f key, same atomic write, same corrupt-read → refetch), so cache already warmed there is reusable as-is; once this merges the monkeypatch is dropped and predict/orbitfit/nightly all benefit." 

what is mean by USDF? Is this to run on the internal Rubin compute?

matthewholman:
Sorry, USDF is the Rubin cluster, where I have been testing fits of the full MPC catalog.  503s are errors from the JPL server.

matthewholman:
Monkeypatch = temporary workaround

kjnapier (approved):
Looks okay to me. I don't know if there will be any negative consequences, but I guess we will find out

PR #411, opened 2026-07-02
Obs80Reader: dispatch two-line records by type + guard blank lines (Closes #402, #407)

matthewholman:
## Summary

Two `Obs80DataReader` robustness bugs the MPC full-catalogue fitting surfaced.

### #402 — roving-observer (`V`/`v`) records

The reader routed **every** two-line record (`S`/`R`/`V`) through the *satellite* parser. But a roving-observer `v` line carries **geodetic longitude/latitude (deg) + altitude (m)** on WGS84, not a geocentric position — so it mis-read the fields and crashed. Now it **dispatches on the first line's note2**:

- **`S`** → geocentric ICRF position (km/AU), unchanged
- **`V`** → geodetic **WGS84** longitude / latitude / altitude. (Capturing the position + its frame here; the geodetic→geocentric-ICRF conversion belongs to the observatory — **#282**, whose WGS84 half this feeds.)
- **`R`** → radar; raise a clear error (radar is ingested via ADES delay/doppler, not the obs80 reader)

### #407 — blank / short lines

`two_line_row_start` and `_is_header_row` indexed columns (`line[14]`, `line[3]`) with no length check, so a blank or truncated line raised `IndexError` and aborted the entire read. Both are now length-guarded, and the read loop skips blank/short lines.

## Tests

New fixture `tests/data/obs80_two_line_records.txt` (single optical + satellite + roving + interspersed blank lines) plus tests:
- roving parses to `WGS84` with the correct lon/lat/alt
- radar raises a clear error
- blank lines are skipped and the mixed file reads 3 records cleanly
- satellite parsing unchanged (existing test still passes)

All 10 `test_Obs80Reader` tests pass; validated end-to-end on real (433) satellite+roving data.

Closes #402. Closes #407.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #425, opened 2026-07-06
Fix Obs80 reader emitting positionless orphan observations (deleted obs + S/s mis-pairing)

### LINKED ISSUE #402: Obs80DataReader crashes on roving-observer (V/v) records; radar (R/r) two-line path also mishandled
### Summary

`Obs80DataReader` crashes on **roving-observer** (`V`/`v`) two-line records, and mishandles the **radar** (`R`/`r`) two-line path, because the second line of *every* two-line record is parsed as if it were a **satellite** position line.

Satellite (`S`/`s`) records themselves are fine (validated against TESS/WISE geocentric positions). The bug is that satellite is not the only two-line format.

### Root cause

`two_line_row_start()` correctly flags all three two-line lead codes:

```python
def two_line_row_start(line):
    note2 = line[14]
    return note2 == "S" or note2 == "R" or note2 == "V"
```

but `convert_obs80(..., second_line=...)` has only the **satellite** layout — it reads a unit flag at cols 33-34 and geocentric X/Y/Z (km/AU):

```python
unit_flag = second_line[32:34].strip()
if unit_flag in ["1", "2"]:
    obs_geo_x = float(second_line[34] + second_line[35:45].strip())
    obs_geo_y = float(second_line[46] + second_line[47:57].strip())   # <-- dies on 'v' lines
    obs_geo_z = float(second_line[58] + second_line[59:69].strip())
```

A roving-observer `v` line does **not** carry geocentric X/Y/Z — it carries **geodetic longitude / latitude / altitude** in a different layout, so the columns above read the latitude field and raise, e.g.:

```
ValueError: could not convert string to float: '55.6977   1'
```

### Reproduction

Any dataset containing roving-observer observations. Concrete example — asteroid **(433) Eros** obs80 has 832 `V`/`v` pairs:

```
00433        KV2026 02 04.687512 03 14 21.014 +23 59 46.68  ...  X~9c60270   <- 'V' obs line
00433        Kv2026 02 04.687512 1  37.8872    55.6977   110.0   ...  X~9c60270   <- 'v' line
                                    ^lon 37.8872  ^lat 55.6977  ^alt 110.0 m
```

Reading this file with `Obs80DataReader(...).read_rows()` (or fitting it with `orbitfit`) raises `ValueError` and aborts the whole object.

### Impact

- Any object/file with roving-observer astrometry fails to read/fit outright.
- Radar (`R`/`r`) obs80 two-line records take the same satellite path (untested; radar is otherwise supported via `delay`/`doppler` columns, but not via obs80 ingestion).
- Common in real MPC archives (amateur roving stations, occultation campaigns).

### Proposed fix

Dispatch the second-line parser on the note2 code of the **first** line instead of assuming satellite:

- `S` -> current geocentric X/Y/Z parser (unchanged).
- `V` -> parse geodetic longitude/latitude/altitude and convert to a geocentric observer position (same target representation as the satellite branch).
- `R` -> radar second-line handling (or explicitly route to the radar ingestion path).

Until then, roving/radar records must be pre-filtered out.

_Found during Phase 0a of the MPC full-catalog fitting demo; companion to #401._

### LINKED ISSUE #407: Obs80DataReader crashes (IndexError) on blank/short lines -- no length guard in two_line_row_start
### Summary

`Obs80DataReader` raises `IndexError: string index out of range` on a **blank or short line** in an obs80 file. `two_line_row_start` indexes column 15 unconditionally:

```python
def two_line_row_start(line):
    note2 = line[14]          # <-- IndexError when len(line) < 15 (blank/short line)
    return note2 == "S" or note2 == "R" or note2 == "V"
```

The read loop calls this on every line (`_read_rows_internal`), so a single blank line (a trailing newline, a separator, or a truncated record) aborts the entire read.

### Reproduction

```
IndexError: string index out of range
  Obs80Reader.py:14 in two_line_row_start   ->  note2 = line[14]
  Obs80Reader.py:234 in _read_rows_internal ->  if two_line_row_start(curr_line):
```

Reading any obs80 file that contains a blank/short line triggers it. (Surfaced in the MPC full-catalogue probe: an object observed *only* by a space telescope, whose records were all 2-line satellite pairs — after filtering, the file reduced to a lone blank line and the reader died.)

### Root cause

Neither `two_line_row_start` nor the read loop guards against lines shorter than the obs80 record width. obs80 records are 80 columns; anything shorter is not a record and should be skipped, not indexed.

### Proposed fix

Skip blank/short lines in the read loop (and guard `two_line_row_start`):

```python
def two_line_row_start(line):
    return len(line) >= 15 and line[14] in ("S", "R", "V")
```

and/or, in `_read_rows_internal`, `continue` on any line whose stripped length is below the minimum obs80 record length before processing it.

### Impact

Any obs80 file with a stray blank/short line crashes the whole read. Low-risk, well-contained fix. Companion to #402 (also `Obs80DataReader` robustness on real archive data).

_Found during Phase 0c of the MPC full-catalogue fitting probe._

matthewholman:
## Summary

The obs80 reader let two kinds of malformed / non-observation lines reach the fitter as **positionless records** (`sys=''`, `pos=nan`). For a space-based obscode (e.g. WISE / `C51`) such a row carries no ADES observer position and, downstream, falls back to a **per-row JPL Horizons lookup** — which 503-throttled under the high parallelism of the full MPC-catalog fit (surfaced on object `j4767`).

Two related bugs, both fixed here:

1. **Deleted / replaced observations (MPC note 2 code `X` / `x`) were read as normal observations.** `j4767`'s WISE block contains one `x` line at the same timestamp as a *kept* satellite obs (a superseded measurement of the same exposure); `tests/data/03666.txt` contains a 1938 `X` photographic plate. These are now skipped.

2. **Two-line records were paired purely by position.** Satellite `S` / radar `R` / roving `V` first lines were merged with whatever line followed, with no check that the follower was the matching lower-case `s` / `r` / `v` continuation. A desynchronised file (a duplicated or orphaned continuation line, or a first line whose continuation is missing) could mis-pair or emit a positionless standalone.

## Change

The three previously independent pairing loops (`get_row_count`, `read_rows`, `read_objects` / `_build_id_map`) are consolidated into a single `_iter_records` generator that is now the sole source of truth for how raw lines group into records. It:

- skips deleted observations (`X` / `x`),
- validates each continuation line against its first line (lower-case note 2 + matching designation cols 1–14 + matching obscode cols 78–80),
- drops orphan / mismatched continuations and continuation-less first lines instead of mis-pairing,

so every read path agrees on the record set (`len(read_rows()) == get_row_count()`, and `read_objects` stays index-aligned with `_build_id_map`).

## Validation

Verified on the real `j4767` bytes from the catalog run:

| reader | rows | C51 rows | C51 positionless |
|--------|------|----------|------------------|
| before | 452  | 13       | **1** (→ Horizons) |
| after  | 451  | 12       | **0** (all `ICRF_KM`) |

Tests added:
- a **real-data fixture** (`tests/data/j4767_wise_excerpt.txt`, j4767's WISE/C51 block) asserting all 12 satellite records carry their `ICRF_KM` position and the deleted `x` line is dropped;
- synthetic **deleted**, **duplicate-continuation**, **leading-orphan**, **missing-continuation**, and **count/read/objects-consistency** cases (all four desync cases fail on the old reader, pass on the new).

All pre-existing `Obs80Reader` tests pass. `tests/data/03666.txt`'s expected row count is updated `4313 → 4312` — the old value had counted the deleted 1938 plate as a real observation.

---
### Update — subsumes #411
Folded in #411 (issue/402-407 Obs80Reader robustness), now closed as superseded. This PR keeps #411's `convert_obs80` **by-type two-line dispatch** — `S` satellite ICRF, `V` roving-observer WGS84 geodetic (previously mis-read as a satellite position, #402/#282), `R` clear 'ingest via ADES' error — plus its `_is_header_row`/#407 blank-line guards and its roving/radar/blank-line tests + `obs80_two_line_records.txt` fixture. Combined with the deleted-`X`/`x` skip and S/s orphan fix here, `_iter_records` is now the single source of truth for record grouping across all read paths. Closes #402, #407. 18 Obs80Reader tests pass.

Closes #402
Closes #407


kjnapier (approved):
Looks good

PR #426, opened 2026-07-06
Make write_hdf5 idempotent; add append_hdf5 for chunked accumulation

matthewholman:
## Summary

`write_hdf5` used `store.append`, so a second write to an existing file — a re-run of `orbitfit`, or re-merging a catalog onto an existing `.h5` — **silently duplicated every row**. This surfaced re-merging the full MPC catalog, where it doubled `numbered.h5` to ~1.79M rows (every object twice).

The append behavior was load-bearing, though: `orbitfit` writes its output one chunk at a time and relied on `append` to accumulate the chunks into one file. So the fix separates the two intents rather than just switching to overwrite.

## Change (Option C)

- **`write_hdf5` now overwrites** (opens the store with `mode="w"`) → repeated writes are idempotent. This is the right default for the single-write callers: `comet`, `convert`, `unpack`, and the catalog finalize/merge step.
- **New `append_hdf5`** keeps the append behavior for accumulation.
- **`orbitfit`'s chunked writer** overwrites on the first write to each output file and appends the remaining chunks (via a small `_emit` helper). The helper also gives the CSV path the same idempotent-first-write behavior, so a re-run of `orbitfit` no longer doubles its output in either format.

## Tests

- `test_write_hdf5_is_idempotent` — a second `write_hdf5` to the same file does not duplicate rows.
- `test_append_hdf5_accumulates` — `write_hdf5` then `append_hdf5` accumulates chunked output.
- The existing accumulation assertion in `test_write_hdf5` is updated to the `write_hdf5` + `append_hdf5` pattern.

Found during the full MPC-catalog run (USDF); relevant to the incremental-fit cycle, where re-writing an existing catalog is the normal case.

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -52,7 +52,7 @@ from layup.utilities.file_io import (
     HDF5DataReader,
     Obs80DataReader,
 )
-from layup.utilities.file_io.file_output import write_csv, write_hdf5
+from layup.utilities.file_io.file_output import append_hdf5, write_csv, write_hdf5
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


kjnapier (approved):
Looks good

PR #428, opened 2026-07-06
Make predict_sequence march (integrate once, interpolate per detection)

matthewholman:
## Summary

`predict_sequence` re-integrated each orbit from its epoch **for every requested time**. It looped detections calling `predict_from_fit_result` → `predict`, and `predict` does `reb_simulation_create … assist_free … reb_simulation_free` per call — so predicting an orbit to N times ran N full integrations from the epoch, each with its own IAS15 ramp-up.

`residuals_at_state` (#406) already solved exactly this for the fit/rejection path with a sorted **forward + backward single pass** that reuses one sim and lets `integrate_or_interpolate` interpolate at each detection. `predict_sequence` simply never got the same treatment. This PR gives it that treatment.

## Change

- **`predict_sequence` now marches.** It splits detections into those after the epoch (marched forward) and before it (marched backward), each sorted, and runs each pass against a single sim — so the trajectory is integrated **once per direction** and `integrate_light_time → assist_integrate_or_interpolate` interpolates at each detection, instead of a fresh integration per detection. Mirrors `residuals_at_state`/`create_sequences`.
- **Refactored** the per-detection geometry + covariance out of `predict()` into a reusable `compute_single_predict(ephem, ax, var, …)` that operates on an existing sim. The bound single-observation `predict()` keeps its own sim create/free and is behavior-unchanged.
- Because each detection carries its own observer position, a single `predict_sequence` call may now **mix observatories** (each detection is light-corrected against its own station).

## Correctness

`integrate_or_interpolate` makes the integration independent of the query times (it interpolates at each), so the marching result is **bit-identical** to the per-detection path:

```
max |ΔRA|  = 0.000e+00 deg
max |ΔDec| = 0.000e+00 deg
max ellipse rel diff = 0.000e+00
```

(verified across times both before and after the epoch, i.e. both passes). New test `test_predict_sequence_marches_equivalently` asserts a one-call multi-time prediction equals per-time predictions to integrator precision.

## Impact

Any multi-time prediction benefits — ephemeris generation, the `predict` CLI over a time grid, and covariance-aware prediction of a catalog to many epochs — turning N integrations per orbit into ~1. No API change (`predict` / `predict_sequence` signatures and outputs are unchanged).

### DIFF
--- a/benchmarks/run_benchmarks.py
+++ b/benchmarks/run_benchmarks.py
@@ -87,14 +87,15 @@ def bench_fit(obs_file="holman_data_working.csv"):
 
 
 def bench_residuals(obs_file="holman_data_working.csv", reps=3):
-    """Speedup of ``residuals_at_state`` over ``predict_sequence`` for a per-
-    observation residual + mapped-covariance sweep (the pass that drives
-    Carpino/OrbFit-style outlier rejection).
+    """``residuals_at_state`` vs ``predict_sequence`` for a per-observation
+    residual + mapped-covariance sweep (the pass that drives Carpino/OrbFit-style
+    outlier rejection).
 
-    ``predict_sequence`` integrates fresh from the epoch to each observation (N
-    integrations); ``residuals_at_state`` reuses the fit's own single forward +
-    backward pass. Reports ``residuals_at_state`` throughput (obs/s) and the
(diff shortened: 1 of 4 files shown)


kjnapier (approved):
Good

PR #429, opened 2026-07-06
predict(): accept one obscode per time (mixed-observatory prediction)

matthewholman:
## Summary

`predict()` predicted every object to a **single** observatory for all requested times. The rest of the chain is already per-time — `obscodes_to_barycentric` reads a per-row `stn`, `_predict` sets each `Observation`'s observer position, and `predict_sequence` light-corrects per detection — so the single-observatory assumption lived in exactly one line (`obs_data = [(obscode, t) for t in times_et]`).

This lets `obscode` be **either** a single string (used for every time, as before) **or** a sequence of length `len(times)` giving one observatory per time — a **mixed-observatory prediction in one call**. Each time is light-corrected against its own station, while each orbit is integrated once across the whole sorted set (with the marching `predict_sequence`).

## Backward compatibility (the important part)

A string `obscode` is broadcast to `[obscode] * len(times)`, which produces the **byte-for-byte identical** `obs_data` and therefore identical output. The test asserts this directly:

```python
predict(data, "X05", times) == predict(data, ["X05"] * len(times), times)   # exactly
```

Existing callers are unaffected — the `predict` CLI passes a single `station`, and interactive / broad-phase callers pass a string. A sequence whose length ≠ `len(times)` raises a clear `ValueError`.

## Tests

`test_predict_mixed_and_single_obscode`:
- string output is byte-identical to a same-code-per-time list (backward compat);
- each time of a mixed-obscode call matches a single-obscode prediction at that station (to 1e-11 deg);
- a length mismatch raises `ValueError`.

## Why it matters

Combined with the marching `predict_sequence` (#428), predicting a catalog to a whole set of `(observatory, time)` targets — e.g. an ITF attribution sweep — can now integrate each orbit **once across all observatories** and interpolate at every target, instead of grouping by observatory and re-integrating per group. This PR is independent of #428 (it is correct on its own; #428 just makes the multi-time case fast).

### DIFF
--- a/src/layup/predict.py
+++ b/src/layup/predict.py
@@ -410,8 +410,11 @@ def predict(
     ----------
     data : numpy structured array
         The data to be processed.
-    obscode : str
-        The observer code.
+    obscode : str or sequence of str
+        The observer code. A single string is used for every time. A sequence of
+        length ``len(times)`` gives one observatory per time -- a mixed-observatory
+        prediction in one call (each time is light-corrected against its own
+        station).
     times : list
         The times for the predictions, in jd_tdb.
     primary_id_column_name : str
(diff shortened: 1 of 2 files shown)


kjnapier (approved):
Nice

PR #430, opened 2026-07-07
Make the non-grav g(r) sublimation law configurable (comet fitting)

matthewholman:
## Summary

The non-gravitational fit **hardcoded the asteroidal inverse-square law** `g(r) = (r/r0)^-2` (`orbit_fit.cpp`), so only Yarkovsky-style A2 could be fit. **Comets** — whose acceleration follows a Marsden water-ice (or other volatile) sublimation law — could not. ASSIST already evaluates the full Marsden form `g(r) = alpha·(r/r0)^-nm·(1+(r/r0)^nn)^-nk` (`forces.c`); this PR just exposes its parameters.

## Change

Thread an optional `gofr = [alpha, nm, nn, nk, r0]` **alongside the existing `a123`** through the fit chain — `compute_residuals_sequence` → `compute_residuals` → `orbit_fit` → `run_from_vector_with_initial_guess` (new pybind arg `gofr`, default empty) — and up to Python `orbitfit(nongrav_gr=...)` and the `fit_nongrav="auto"` selection path.

A `null`/empty `gofr` uses the inverse-square law, so **every existing non-grav fit is unchanged**.

## Backward compatibility

The default (and an explicit `[1, 2, 1, 0, 1]`) reproduce the previous fit **byte-for-byte**:

```
default    a2 = -2.757526e-06   csq = 9.2796
asteroid   a2 = -2.757526e-06   csq = 9.2796   (identical)
water-ice  a2 = -2.900889e-04   csq = 9.2620   (g(r) is used: A and g(r) trade off)
```

## Tests

`test_orbitfit_nongrav_gr_default_and_configurable`: the default equals an explicit inverse-square law byte-for-byte and still recovers the true A2; a Marsden water-ice `g(r)` changes the fitted amplitude; a malformed `g(r)` raises. The existing non-grav suites pass (the `fit_nongrav="auto"` test fakes were updated to accept the new `gofr` kwarg).

## Motivation

Enables cometary non-grav fitting — the foundation for comet linkage (joint-fitting two apparitions with a shared g(r) sublimation law), and generally useful for any comet orbit fit.

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -218,18 +218,35 @@ def _nongrav_warranted(csq_gravity, res_ng, names, thresholds=_AUTO_DEFAULT_THRE
     )
 
 
-def _select_nongrav_auto(assist_ephem, res_grav, observations, thresholds=_AUTO_DEFAULT_THRESHOLDS):
+def _gofr_arg(nongrav_gr):
+    """Normalize a non-grav g(r) argument to the ``[alpha, nm, nn, nk, r0]`` list the
+    C++ fit expects (empty -> the default inverse-square law)."""
+    if nongrav_gr is None:
+        return []
+    gr = list(nongrav_gr)
+    if len(gr) != 5:
+        raise ValueError("nongrav_gr must be [alpha, nm, nn, nk, r0] (5 values) or None")
+    return [float(v) for v in gr]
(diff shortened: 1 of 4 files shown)


kjnapier (approved):
I think this makes sense

PR #431, opened 2026-07-07
Support supplied per-observation astrometric uncertainties

matthewholman:
## What

Adds a third astrometric-weighting mode to `orbitfit`: `weight_data="supplied"` reads the per-observation `rmsRA`/`rmsDec` columns (in arcseconds) and uses them directly as the observation uncertainties, instead of the built-in ~1-arcsecond default (`weight_data=False`) or the Veres et al. (2017) model (`weight_data=True`).

## Why

layup currently weights every observation at ~1 arcsecond by default, and the Veres 2017 model returns ~1 arcsecond for anything pre-CCD. That badly over-weights historical astrometry — 19th- and early-20th-century visual and photographic positions are really only good to several arcseconds, up to an arcminute. In the fit's reject/refit loop that then throws the orbit out entirely, so a century-old comet discovery arc simply will not fit. Letting the caller supply per-observation uncertainties — from ADES, or from an external model such as era-based historical weighting — fixes that without hard-coding any weighting policy into layup.

## Behaviour / compatibility

- `weight_data=False` (default) and `weight_data=True` (Veres 2017) are unchanged — byte-identical.
- `"supplied"` requires the `rmsRA`/`rmsDec` columns (raises a clear error otherwise); a NaN or non-positive value on a given row falls back to the default weighting for that row.

## Tests

Adds tests that a looser supplied sigma yields a proportionally looser position covariance (confirming the weights are actually applied) and that the missing-columns case raises. The existing `weight_data` regression tests pass.

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -1015,7 +1015,7 @@ def _orbitfit(
     initial_guess=None,
     bias_dict: dict = None,
     sort_array: bool = True,
-    weight_data: bool = False,
+    weight_data=False,  # bool (Veres 2017) or "supplied" (rmsRA/rmsDec columns)
     iod: str = "gauss",
     engine: str = "cartesian",
     fit_nongrav: bool = False,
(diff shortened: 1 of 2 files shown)


kjnapier (approved):
Excellent

PR #432, opened 2026-07-07
Fit piecewise per-apparition non-gravitational amplitudes (per_arc)

matthewholman:
## What

Adds an optional `per_arc` mode to the Cartesian non-gravitational fit. It shares one orbit (state) and one `g(r)` sublimation law across the fit, but lets the non-gravitational amplitudes A1/A2/A3 differ between the observations before the fit epoch and those after it.

## Why

A comet's non-gravitational acceleration is A·g(r). The `g(r)` shape is a property of the subliming ice, so it is shared over time; the amplitude A reflects the active surface, which can change from one apparition to the next (or differ between fragments of a split comet). Fitting two apparitions as one orbit while allowing the amplitude to differ lets you ask not just "is this the same comet?" but *how* it is the same — stable, re-activated, or split. Because the non-gravitational force is negligible far from the Sun, a piecewise-constant amplitude (one value near each apparition's perihelion) is a good approximation.

## How

Reuses the fit's existing forward/reverse integration from the fit epoch as the two arcs (so the fit epoch must sit between the two apparitions). Each pass carries its own amplitude triple and variational particles, remapped into disjoint columns of a `6 + 2*nactive` parameter block. `orbitfit(..., per_arc=True)` threads it through and adds opt-in `a{1,2,3}_arc2` result columns (the later arc); the base `a{1,2,3}` columns then hold the earlier arc.

## Behaviour / compatibility

- `per_arc=False` (the default) is byte-identical to the ordinary shared-amplitude fit.
- Tests: a piecewise-constant synthetic truth is recovered (equal amplitudes recover equal; distinct amplitudes recover distinct and clear 3 sigma); `per_arc=False` still fits a single shared amplitude; the driver adds the `a{1,2,3}_arc2` columns only when requested. Existing non-grav suites pass.

## Dependency

Stacked on #430 (the `g(r)`-configurable non-grav fit): the base of this PR is `feat/nongrav-gofr`, so it should be retargeted to `main` and merged after #430 lands.

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -252,14 +252,24 @@ def _select_nongrav_auto(
     return res_grav  # no non-grav model is warranted
 
 
-def _get_result_dtypes(primary_id_column_name: str, nongrav_names=()):
+def _get_result_dtypes(primary_id_column_name: str, nongrav_names=(), per_arc=False):
     """Helper function to create the result dtype with the correct primary ID column name.
 
     For each fitted non-gravitational parameter in ``nongrav_names`` (a subset of
     ``A1``/``A2``/``A3``), two columns are appended -- e.g. ``a2`` and ``a2_unc``
     (the value and its 1-sigma uncertainty, au/day^2). With no non-grav params the
     default 6-parameter output schema is unchanged.
+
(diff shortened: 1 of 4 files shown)


kjnapier (approved):
Wow.

PR #438, opened 2026-07-14
Default the IAS15 integrator to the modern (mode 2) adaptive step controller

matthewholman:
## Summary
Change the default of the IAS15 adaptive step-size controller from the legacy controller (mode 1) to the newer controller of Pham, Rein & Spiegel (2024) (mode 2). Concretely this flips a single global default in `orbit_fit.cpp` (`g_ias15_adaptive_mode`, `-1` -> `2`) and updates its comment. The knob itself and its per-simulation application were added in #324 and are unchanged.

## Motivation
`assist_attach()` sets IAS15 to the legacy adaptive controller, which drives the integration timestep toward zero during close encounters with the Earth. A single such integration can then fail to return, hanging the orbit fit and consuming the wall-clock budget. This was the entire "timeout / hang" class encountered while re-fitting the full Minor Planet Center (MPC) catalog: close-approaching near-Earth objects (NEOs) that never completed under the legacy controller.

The mode-2 controller steps through these encounters efficiently. Crucially, **correctness is independent of the controller** -- the step controller only decides whether and how quickly an integration completes, never the resulting orbit. In the full-catalog run, the 40 hardest close-NEO cases (perihelion 0.24-1.02 AU) that hung under the legacy controller all completed in under 0.5 s under mode 2, with fitted states identical (to numerical tolerance) across controllers. Hanno Rein suggested this controller in the #324 review as a cleaner alternative to the minimum-timestep floor.

## Change
- `src/lib/orbit_fit/orbit_fit.cpp`: `g_ias15_adaptive_mode` default `-1` -> `2` (+ comment).
- No API change. `set_ias15_adaptive_mode()` / `get_ias15_adaptive_mode()` still let a caller override (e.g. `-1` to restore the legacy behavior, `1` to force it explicitly).
- The fit path, both predict paths, and the IOD picker already call `apply_ias15_adaptive_mode()` after `assist_attach`, so all of them pick up the new default.

## Tests
- `test_ias15_settings.py`: updated the default-value test to assert the new default (2).
- Added `test_fit_is_mode_independent` (guarded by `requires_ephem`): a well-behaved object fit under the legacy controller (mode 1) and the modern controller (mode 2) returns the same state -- verified locally that the two fits agree bit-for-bit.

## Scope / caveats
- This addresses the integrator **hang** class only. It does not fix the separate 2020 CD3 case, whose fit fails from a numerical breakdown in the observation model / variational partials during a sub-0.0003 AU topocentric pass -- a distinct, lower-priority item.

## Related
- Builds on #324 (introduced the adaptive-mode knob and its per-simulation application).

### DIFF
--- a/src/lib/orbit_fit/orbit_fit.cpp
+++ b/src/lib/orbit_fit/orbit_fit.cpp
@@ -63,9 +63,12 @@ namespace orbit_fit
     // the legacy value, apply_ias15_adaptive_mode() must run *after*
     // assist_attach, not before like apply_ias15_min_dt.
     //
-    // Default -1 means "leave ASSIST's choice untouched" (no behavior
-    // change for existing callers).
-    static int g_ias15_adaptive_mode = -1;
+    // Default 2 selects the newer (Pham, Rein & Spiegel 2024) controller on
+    // every freshly-attached sim: it steps through close-Earth encounters
+    // gracefully, whereas the legacy controller (mode 1) drives dt -> 0 and can
+    // hang the fit, at no accuracy cost (correctness is mode-independent).
+    // Set to -1 to leave ASSIST's legacy choice untouched, or 1 to force it.
+    static int g_ias15_adaptive_mode = 2;
 
     inline void set_ias15_adaptive_mode(int m) { g_ias15_adaptive_mode = m; }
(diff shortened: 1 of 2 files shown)


kjnapier (approved):
looks good

PR #442, opened 2026-07-17
Added 80 column fix from Sorcha code

### LINKED ISSUE #433: orbitfit/convert/predict fail with SPICE TYPEMISMATCH when the data-cache path is long
When the data cache lives under a path longer than ~80 characters, every verb that furnishes SPICE kernels (`orbitfit`, `convert`, `predict`) hard-fails with:

```
SPICE(TYPEMISMATCH) ... PATH_VALUES ... not a string value
```

on the generated `meta_kernel.txt`.

**Root cause:** SPICE kernel-pool *string* values are capped at 80 characters. The generated `PATH_VALUES = ('<full cache path>')` is a single over-long string and is not split using SPICE's `+=` continuation idiom.

**Reproduce:** run any fit with the cache under a deep path (e.g. a container/CI scratch dir). The identical run under a short path (e.g. `/tmp/lc`) succeeds.

**Impact:** default `~/Library/Caches/layup` users are unaffected, but CI, containers, and deep scratch dirs fail — and with a cryptic SPICE error rather than a layup-level diagnostic.

**Suggested fix:** emit continued `PATH_VALUES` lines via the SPICE `+=` continuation, or write `PATH_VALUES` relative to a `PATH_SYMBOLS` base directory.

_Found during a clean-room, fresh-user install + test run-through (fresh venv + `--recursive` clone, following the README Quickstart), 2026-07-10._

Little-Ryugu:
Added _split_kernel_path_str and its unit test from Sorcha to avoid 80 column bug

Fixes #433  .

Describe your changes.

### DIFF
--- a/src/layup/utilities/bootstrap_utilties/create_meta_kernel.py
+++ b/src/layup/utilities/bootstrap_utilties/create_meta_kernel.py
@@ -27,8 +27,8 @@ from layup.utilities.layup_configs import AuxiliaryConfigs
 
 
 def _split_kernel_path_str(abspath: str, split=77):
-    """If abspath string is longer than 79 chars, split it up in the meta kernel by inserting "+' '".
-    79 is the default because the character limit in SPICE is 80 before the string needs split.
+    """If abspath string is longer than 77 chars, split it up in the meta kernel by inserting "+',\n'".
+    77 is the default because the character limit in SPICE is 80 before the string needs split.
 
     Parameters
     ----------
--- a/src/layup/utilities/bootstrap_utilties/create_meta_kernel.py
+++ b/src/layup/utilities/bootstrap_utilties/create_meta_kernel.py
@@ -26,6 +26,28 @@ from layup.utilities.layup_configs import AuxiliaryConfigs
 """
 
 
+def _split_kernel_path_str(abspath: str, split=77):
+    """If abspath string is longer than 79 chars, split it up in the meta kernel by inserting "+' '".
+    79 is the default because the character limit in SPICE is 80 before the string needs split.
+
+    Parameters
+    ----------
+    abspath: str
+        The filepath string that needs split.
+    Split : int
+        The maximum size each part of the filepath can be before it needs split.
+    Returns
+    ---------
+    abspath : str
+    The filepath string split into chunks of required size.
+    """
+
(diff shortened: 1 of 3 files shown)


matthewholman:
Nice work, Ryan — tested it end-to-end and it fixes the bug. A long (~165-char) cache path reproduced the `TYPEMISMATCH` on `main`. Confirmed your change lets a kernel actually load through the generated `meta_kernel.txt`. The comma in `+',\n'` is the real fix.

One non-blocking note: the unit test checks the split string but never calls `furnsh`, so it wouldn't catch a break in the actual SPICE load. Here's a short integration test that guards the real behavior — builds a meta_kernel under a deliberately deep path and loads a kernel through it:

```python
import os
import spiceypy
from layup.utilities.bootstrap_utilties.create_meta_kernel import _split_kernel_path_str


def test_meta_kernel_loads_under_long_path(tmp_path):
    # A cache path long enough to trip the SPICE 80-char string limit.
    deep = tmp_path
    while len(str(deep)) < 160:
        deep = deep / "12345678"
    deep.mkdir(parents=True, exist_ok=True)

    # A tiny kernel we can check actually loaded.
    (deep / "tiny.tls").write_text(
        "\\begindata\nDELTET/DELTA_T_A = 99.5\n\\begintext\n"
    )

    mk = deep / "meta_kernel.txt"
    mk.write_text(
        "\\begindata\n\n"
        f"PATH_VALUES = ('{_split_kernel_path_str(str(deep))}')\n\n"
        "PATH_SYMBOLS = ('A')\n\n"
        "KERNELS_TO_LOAD = ( '$A/tiny.tls' )\n\n"
        "\\begintext\n"
    )

    spiceypy.kclear()
    try:
        spiceypy.furnsh(str(mk))                     # would raise TYPEMISMATCH before the fix
        assert spiceypy.gdpool("DELTET/DELTA_T_A", 0, 1)[0] == 99.5
    finally:
        spiceypy.kclear()
```

(Also, docstring says default 79 but it's `split=77`.) Approve after that — thanks for the quick turnaround!

Little-Ryugu:
No worries. I fixed the docdtring for sorcha but forgot layup oops. Thats a very good point with the unit test, it caused us to miss the bug in Sorcha. I'll add that test to both repos too thank you!!

matthewholman (approved):
Looks good!

PR #443, opened 2026-07-21
Minor changes to installation instructions.

hannorein:
I've done a fresh installation of layup and all dependencies on linux. Everything works well. I've made some small edits to the installation instructions.

One thing I noticed: On my cluster, the home directory is on a different partition than the "big" data partition. I think this is a fairly common setup. Currently, if I create a virtual environment with layup on the data partition, then it still writes almost half the data to my home partition, i.e. ~1.6G of files to `~/.cache/layup`. In my opinion it would be nice to keep all of layup's files within the directory it is installed in. But I understand that the `~/.cache` approach is kind of a standard thing to do... So I don't know. Not a big issue, I guess.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -23,6 +23,12 @@ create and activate a new environment.
    >> conda create env -n <env_name> python=3.11
    >> conda activate <env_name>
 
+Alternatively, you can create a virtual environment with python's `venv` module.
+
+.. code-block:: console
+
+   >> python -m venv venv
+   >> source venv/bin/activate
 
 Once you have created a new environment, you can install this project for local
 development using the following commands:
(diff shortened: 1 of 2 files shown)


docs/index.rst
```diff
@@ -23,6 +23,12 @@ create and activate a new environment.
    >> conda create env -n <env_name> python=3.11
```
matthewholman:
```suggestion
   >> conda create -n <env_name> python=3.11
```
Minor pre-existing typo, just above your `venv` addition: `conda create env -n ...` fails, because `conda create` reads `env` as a package name (`PackagesNotFoundError: env`). The valid forms are `conda create -n <name> ...` or the separate file-based `conda env create -f <file>`.

mschwamb:
You can specify with the bootstrap utility where you want the files that normally go into ~/.cache and in all of the layup verbs there is an --ar which allows you to specify where the cache files are if they're not in the usual place. I think that does what you want @hannorein

mschwamb (approved):

PR #449, opened 2026-07-24
fixed CLI duplicate bug

### LINKED ISSUE #434: CLI logging is duplicated, and convert reads the CSV header twice
Two small output issues seen across verbs:

1. **Duplicate log lines.** Every verb prints its completion message twice, in two formats, e.g.
   `... layup.convert ... INFO - Data has been written to X.csv` **and** `INFO:layup.convert:Data has been written to X.csv`.
   This is consistent with a custom handler *plus* propagation to the root logger.
2. **Header read twice.** `convert` logs `Reading the first line ... as header` twice per run (~5 ms apart, same format), suggesting the reader opens/peeks the input file twice.

Both are cosmetic, but they make the CLI look chatty, and the double read may waste I/O on large inputs.

**Suggested fix:** set `logger.propagate = False` (or configure a single handler), and peek the header once.

_Found during a clean-room, fresh-user install + test run-through (fresh venv + `--recursive` clone, following the README Quickstart), 2026-07-10._

Little-Ryugu:
Fixes #434  .

CSVReader had a log output that none of the other logs had. This caused duplicate CLI prints when the sample reader and full reader were called. I've moved this log to convert.py so it only occurs once.

logger. propagate was set to False to avoid other duplicate logs, such as in the output file log.

- [x]  Have you written a unit test for any new functions?

### DIFF
--- a/src/layup/convert.py
+++ b/src/layup/convert.py
@@ -784,6 +784,7 @@ def convert_cli(
 
     sample_data = sample_reader.read_rows(block_start=0, block_size=1)
 
+    logger.info(f"Reading the first line of {input_file} as header:\n{sample_data.dtype.names}\n")
     # Check orbit format in the file
     input_format = get_format(sample_data)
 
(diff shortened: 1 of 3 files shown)


matthewholman (approved):
Looks good to me.

Issue #448, opened 2026-07-23
Make the default ephemeris/reference-data cache location configurable (currently ~/.cache, on the home partition)


matthewholman:
Raised by @hannorein in #443: layup writes ~1.6 GB of ephemeris/reference data to `~/.cache/layup` even when layup is installed on a separate data partition. On clusters where the home directory is a different (often smaller) partition than the data partition, this is awkward.

The default comes from `pooch.os_cache("layup")` (in `predict.py` and `orbitfit.py`). A `cache_dir` argument already overrides it per call, but the default falls back to the OS cache in the home directory.

Suggestion: make the default configurable — e.g. honor an environment variable (`LAYUP_CACHE_DIR`, or the standard `XDG_CACHE_HOME`) — so all of layup's data can be kept off the home partition without threading `cache_dir` through every call.

mschwamb:
We already have this. You just have to specify this on the command line flag --ar and run the bootstrap capability to with the flag on the command line to put the cache files whever ever you want. This all needs to be documented in the read the docs

hannorein:
@mschwamb That works. Thanks. I didn't know that!

PR #452, opened 2026-07-27
fixed cache dir bug

### LINKED ISSUE #451: Layup checking Sorcha cache for files when cache_dir is None in LayupObservatory.
When initialising the LayupObservatory class and the cache_dir is None, Layup will use the local user's Sorcha cache for aux files.

This is because the function layup_furnish_spicepy uses the Sorcha function `sorcha.ephemeris.simulation_setup.furnish_spiceypy `, which defaults to the Sorcha cache when  cache_dir is none:

https://github.com/Smithsonian/layup/blob/bec27479c926af40464eb10597761ed7715724b0/src/layup/utilities/data_processing_utilities.py#L347-L352

 https://github.com/Smithsonian/layup/blob/bec27479c926af40464eb10597761ed7715724b0/src/layup/utilities/data_processing_utilities.py#L354-L378

This resulted in a failed unit test for me locally, as my Sorcha cache didn't have up-to-date Obs codes (unit tests worked after Sorcha bootstrap was used). 

Both orbitfit and predict that use this class specify the cache_dir, so no issues in the main code, only unit tests that don't specify cache_dir can lead to this error. 

Changing the default from `None` to the local layup cache dir: `cache_dir = str(pooch.os_cache("layup"))` should fix this.

Little-Ryugu:
Fixes #451  .

Changed cache_dir to default to str(pooch.os_cache("layup")) instead of None in LayupObservatory. Stops layup_furnish_spiceypy from searching for files in the Sorcha created cache.


Little-Ryugu:
Don't accept/merge just yet. @mschwamb said there was a different bug before that caused a problem similar to this in orbitfit. I'm going to double-check everything first incase I missed something.

Little-Ryugu:
@mschwamb, the fix was in pr #225 but missed the case where cache_dir is left as None. The call for Layup_observatory in orbitfit and predict was changed from:
```
Layup_observatory = LayupObservatory()
```
To:
```
Layup_observatory = LayupObservatory(cache_dir=cache_dir) 
```
This fixed the original bug in the main code, but the unit tests still have the bug as they call ` Layup_observatory= LayupObservatory()`

There was another fix to a similar issue #286, but the fix was related to the observatory position and didn't change the unit tests calling the Sorcha cache.

mschwamb:
I think you might want to make "layup" a global variable at the top of the file so it's clear that this is hardcoded - easier to spot if it needs to be changed.

mschwamb (changes requested):
see other comment about changes

src/layup/utilities/data_processing_utilities.py
```diff
@@ -59,6 +59,10 @@
 
 logger = logging.getLogger(__name__)
 
+# default Cache directory name where the layup auxiliary files are stored when layup bootstrap is ran
+# eg on Mac ~/Library/Caches/layup
+cache_dir_name = "layup"
```
mschwamb: Can you make it all caps - so that the style is the same throughout layup see https://github.com/Smithsonian/layup/blob/ed1579add6615e7ec9e4e9ffae0dc1c7feed7afc/src/layup/constants.py#L21

mschwamb (changes requested):

src/layup/utilities/data_processing_utilities.py
Little-Ryugu: No problem will do.

mschwamb (approved):

matthewholman (changes requested):
**This doesn't fix #451 on the paths that matter.**

Changing the default only helps callers that omit the argument. All three production callers pass it explicitly, and their own defaults are `None`:

- `predict.py:438`, `orbitfit.py:1502`, `orbitfit.py:1555`

So `layup.predict(data, obscode, times)` still sends `None` through to `layup_furnish_spiceypy`, which hands it to Sorcha as `ar_data_file_path=None` — the search #451 reports. `tests/layup/test_predict.py:480` passes `cache_dir=None` explicitly too.

The evidence is one line above a caller:

```python
kernels_loc = str(pooch.os_cache("layup")) if cache_dir is None else str(cache_dir)   # orbitfit.py:1554
observatory = LayupObservatory(cache_dir=cache_dir)                                   # 1555
```

Also, a default argument is evaluated once at import, so this freezes the cache path at module load — awkward for tests and for #448.

Suggested instead:

```python
def __init__(self, cache_dir=None):
    if cache_dir is None:
        cache_dir = str(pooch.os_cache("layup"))
    self.cache_dir = cache_dir
    layup_furnish_spiceypy(cache_dir)
```

This covers explicit `None`, evaluates per call, matches the idiom used elsewhere (including line 434 of this file), and fixes `self.cache_dir`, which feeds the Horizons state cache and is currently stored as `None`.

A test asserting `predict(..., cache_dir=None)` furnishes from the layup cache would catch the regression.

mschwamb (changes requested):
See the detailed review from Matt

Little-Ryugu:
That's a good point, @matthewholman. My change also still leaves an issue if a user specifies cache_dir=None directly when pip installing Layup and using the predict or orbitfit function. I'll add that change now.

matthewholman (commented):
Fix is correct — normalising in the body covers the explicit-`cache_dir=None` path that all three production callers take, and `layup_furnish_spiceypy` now gets the resolved value. CI green.

One request before merge: a test for that path. This bug survived once already because the tests call `LayupObservatory()` bare, which was never the broken case.

```python
def test_layup_observatory_explicit_none_uses_layup_cache():
    obs = LayupObservatory(cache_dir=None)
    assert obs.cache_dir == str(pooch.os_cache("layup"))
```

Minor: `self.cache_dir = cache_dir` is now set twice, at lines 455 and 462.

@mschwamb, this is still yours to approve.

mschwamb:
@matthewholman  you should be the one to approve this as you've already done the deep dive on stuff I missed

matthewholman (approved):
This is great, @Little-Ryugu!

matthewholman:
Approved. @mschwamb your earlier change request is still recorded and is blocking the merge — could you clear it? Then @Little-Ryugu can merge.

mschwamb (approved):
Approved

PR #453, opened 2026-07-28
Mtot as primary, not Msun

### LINKED ISSUE #447: Comet original 1/a_0 should use barycentric elements with GM_total; code uses heliocentric with GM_sun
The original (and future) orbit of a long-period comet — its 1/a_0 — is conventionally computed from **barycentric** osculating elements, using GM_total (the total solar-system mass) as the gravitational parameter.

The current `comet.py` appears to use **heliocentric** elements with GM_sun: the osculating orbit is taken relative to the Sun (`sim.particles[-1].orbit(primary=ephem.get_particle("sun", ...))`), so the reported 1/a_0 and the `ao_barycentric` column are heliocentric despite the name.

**Key references**
- Marsden, Sekanina & Everhart (1978), AJ, 83, 64 — defines the original and future 1/a "referred to the barycenter of the solar system." https://ui.adsabs.harvard.edu/abs/1978AJ.....83...64M
- Królikowska (2014), A&A, 567, A126 — Warsaw Catalogue of cometary orbits; each comet is integrated to 250 au, at which point the reference frame is switched from heliocentric to barycentric. https://arxiv.org/abs/1407.0553
- Dybczyński & Królikowska (2020), A&A — The catalogue of cometary orbits and their dynamical evolution (CODE catalogue). https://arxiv.org/abs/2005.09698

awilson110:
Fixes #447  .

Replaced the primary to be Mtot instead of Msun


hannorein:
Hi @awilson110, this looks good. I think this should work.

matthewholman (commented):
Correct fix for #447 — `rebound.Particle(m=Mtot)` is the barycenter at the origin, so the reported elements are now genuinely barycentric (μ = GM_total, position relative to the SSB), and the `demo_comet_expected.csv` fixtures are regenerated to match.

One thing worth confirming: the pure-REBOUND fallback (`include_assist=False`, previously `sim.particles[0]`) now also uses the origin+`Mtot` primary. That's right as long as `simulation_convert_to_rebound` leaves the sim in the SSB frame — but that branch only fires for comets past the ASSIST timeframe, so the fixtures probably don't exercise it. A quick check would close it.

Minor: `tests/layup/test_layup_configs.py` drops one line near the top — intentional, or a stray edit?

awilson110:
> Correct fix for #447 — `rebound.Particle(m=Mtot)` is the barycenter at the origin, so the reported elements are now genuinely barycentric (μ = GM_total, position relative to the SSB), and the `demo_comet_expected.csv` fixtures are regenerated to match.
> 
> One thing worth confirming: the pure-REBOUND fallback (`include_assist=False`, previously `sim.particles[0]`) now also uses the origin+`Mtot` primary. That's right as long as `simulation_convert_to_rebound` leaves the sim in the SSB frame — but that branch only fires for comets past the ASSIST timeframe, so the fixtures probably don't exercise it. A quick check would close it.
> 

@hannorein would you be able to confirm that the function `simulation_convert_to_rebound` keeps the simulation in the barycentric frame? I'd imagine it does, but good to check

> Minor: `tests/layup/test_layup_configs.py` drops one line near the top — intentional, or a stray edit?

That was a mistake, I can remove it if needed

hannorein:
Yes, `simulation_convert_to_rebound` does not change the frame.

matthewholman (commented):
Looks good

awilson110:
@hannorein @matthewholman if we all agree the PR looks good, would one of you be able to approve it so I can merge?

matthewholman (commented):
This seems ok.  The tests all pass?

awilson110:
Yeah the tests all pass after I reran them a couple of times (the failure was from being unable to download the mpc files), the only block currently is approval

matthewholman (approved):
All set.

PR #456, opened 2026-08-11
updated SPICE files and unit tests

### LINKED ISSUE #455: kernel files need updating
earth_620120_250826.bpc --> earth_620120_260806.bpc
earth_2025_250826_2125_predict.bpc --> earth_2026_260806_2126_predict.bpc

yearly update of these kernel files - need our defaults updated and unit tests

Little-Ryugu:
Fixes #455  .

updated SPICE files and unit tests

### DIFF
--- a/src/layup/config_setups/default_config.ini
+++ b/src/layup/config_setups/default_config.ini
@@ -4,11 +4,11 @@
 planet_ephemeris = de440s.bsp
 planet_ephemeris_url = https://naif.jpl.nasa.gov/pub/naif/generic_kernels/spk/planets/de440s.bsp
 
-earth_predict = earth_2025_250826_2125_predict.bpc
-earth_predict_url = https://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/earth_2025_250826_2125_predict.bpc
+earth_predict = earth_2026_260806_2126_predict.bpc
+earth_predict_url = https://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/earth_2026_260806_2126_predict.bpc
 
-earth_historical = earth_620120_250826.bpc
-earth_historical_url = https://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/earth_620120_250826.bpc
+earth_historical = earth_620120_260806.bpc
+earth_historical_url = https://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/earth_620120_260806.bpc
 
 earth_high_precision = earth_latest_high_prec.bpc
 earth_high_precision_url = https://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/earth_latest_high_prec.bpc
--- a/src/layup/utilities/layup_configs.py
+++ b/src/layup/utilities/layup_configs.py
@@ -24,14 +24,14 @@ class AuxiliaryConfigs:
     planet_ephemeris_url: str = f"{naif_base_url}/spk/planets/de440s.bsp"
     """url for planet_ephemeris"""
 
-    earth_predict: str = "earth_2025_250826_2125_predict.bpc"
+    earth_predict: str = "earth_2026_260806_2126_predict.bpc"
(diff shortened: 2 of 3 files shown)


mschwamb (approved):
This is the same as the Sorcha updates which are working. This will be needed to get a full running version of layup

matthewholman:
The kernel filenames and URLs check out — both new files return 200 from NAIF, and both of the ones being replaced now return 404, so this is fixing a live breakage rather than routine maintenance: a fresh `layup bootstrap` on `main` cannot complete until it merges.

One thing to add while here. The workflow caches the downloaded data with

```yaml
key: layup-data-${{ runner.os }}-v1
```

and the comment above it says to bump the version suffix to force a refresh. Since this change replaces two of the cached files, it should go to `-v2`.

Worth doing for a second reason: that warm cache is why CI never noticed the old URLs had died. Runs kept restoring kernels that no longer exist upstream, so the suite stayed green while a new user's bootstrap would have failed. Bumping the key restores CI's ability to see this class of failure, and avoids the cache quietly accumulating both generations of kernels.

mschwamb (changes requested):
See @matthewholman's comment

matthewholman:
Recording the diagnosis here, since this will happen again every time NAIF retires a kernel.

**Symptom.** CI goes red across all platforms. A handful of tests fail — not the whole suite — in the modules that build a fresh `cache_dir`, and the run may be killed at the 20-minute cap.

**Cause.** NAIF publishes new Earth orientation kernels and removes the old ones. Our pinned defaults then 404. The replacement here is dated `260806`, so NAIF published it on 6 August and retired the previous pair within a few days.

**Why it is hard to see.** Two layers of masking:

1. The `actions/cache` step restores `~/.cache/layup` containing the *old* kernels, so every test that uses the default cache still finds its files and passes. Only tests that download into a fresh directory hit the 404. That is why a catalogue-wide breakage looks like three or four unrelated test failures.
2. Since 9 August the macOS legs failed earlier still, at *Install dependencies* (#457), and fail-fast cancelled the Linux legs before they got far enough to show anything. The last complete run was 8 August.

**The timeline is the clearest evidence.** `main` at the identical commit `ed1579a` was green 5–8 August and has failed every day since the 9th. No code changed; the data moved.

**What to do next time**

1. Check the URLs directly — `curl -sI <url>` on the two `*_url` values in `default_config.ini`. A 404 confirms it in seconds.
2. Update all three places, as this PR does: `config_setups/default_config.ini`, `utilities/layup_configs.py`, `tests/layup/test_layup_configs.py`.
3. Bump the cache key suffix. Without it the stale cache keeps hiding the problem.
4. Expect the first run after a key bump to be slow — a cold bootstrap pulls the whole auxiliary set, including `linux_p1550p2650.440` at roughly 600 MB. With `timeout-minutes: 20` that run may itself time out, so raise the limit in the same change rather than afterwards.

Merging this now. #457 has to land as well before macOS can get far enough to confirm it, so neither PR alone will produce a green run.

matthewholman:
@mschwamb the cache-key change is now #460, so this PR does not need to carry it.

Splitting it out for two reasons. It is a CI concern rather than a kernel-filename one, and it turned out not to be a one-line change: bumping the key makes the cache cold, and a cold bootstrap re-fetches the whole auxiliary set including `linux_p1550p2650.440` at ~600 MB. Against the current `timeout-minutes: 20` that first run would likely have timed out — so the key bump and a timeout raise have to land together, which is more than belongs in a kernel update.

That was my omission in asking for it here; apologies for the detour.

This PR is correct as it stands, and it fixes breakage that is live right now — both replaced kernels return 404 from NAIF, so a fresh `layup bootstrap` on `main` cannot complete. Could you clear the change request so it can go in?

Note that its own macOS leg will still fail until #459 lands (the AppleClang 21 link error, #457), and #459's tests cannot pass until this one lands. Neither is green alone; the pair is.

mschwamb (approved):

PR #459, opened 2026-08-11
Don't link the compiled extension against assist/rebound on macOS

matthewholman:
Draft, opened to get the macOS CI legs to answer a question this machine cannot: **does the build succeed under AppleClang 21?**

### The problem

`CMakeLists.txt` linked `_core` directly against assist's and rebound's compiled Python extension modules. On macOS such a module is either a dynamic library (`MH_DYLIB`, linkable) or a bundle (`MH_BUNDLE`, loadable only). Xcode 26 / AppleClang 21 produces bundles, so `ld` refuses:

```
ld: unsupported mach-o filetype (only MH_OBJECT and MH_DYLIB can be linked) in
'.../site-packages/libassist.cpython-313-darwin.so'
```

Nothing in layup changed — the runner image did. Neither package promises a link type, so this was always resting on an implementation detail.

### The change

On macOS, build the extension unlinked and let the rebound/assist symbols resolve from the flat namespace at load time; the link line already carries `-undefined dynamic_lookup`. Linux keeps linking with its RPATH, unchanged.

That alone is not enough, which is the part worth reviewing. Importing `assist` and `rebound` does **not** put their symbols in the flat namespace — both load their libraries through ctypes, whose default mode is `RTLD_LOCAL`, so `import layup.routines` dies with:

```
ImportError: dlopen(..._core.cpython-312-darwin.so, 0x0002):
symbol not found in flat namespace '_assist_attach'
```

So `layup/routines/__init__.py` reopens the libraries with `RTLD_GLOBAL` before importing the extension.

It deliberately reopens the paths the packages expose as `__libpath__`, not the copies layup installs beside the extension. I checked that this promotes the already-loaded image rather than mapping a second one — `assist_init` resolves to the same address through assist's own handle and through the `RTLD_GLOBAL` handle. A second copy would have given the C++ side its own library-global state, separate from the one Python-level `assist`/`rebound` calls use.

### Testing, and its limit

macOS 15.2 / AppleClang 16: full suite 460 passed, 1 skipped, both before and after. Verified that the unlinked extension fails to import without the promotion and succeeds with it.

**This machine cannot reproduce the failure** — AppleClang 16 links the old code fine. So the local runs show the new loading path works; they cannot show that the build is fixed. That is what the macOS legs here are for.

### Note for reviewers

If the macOS legs still fail, the fallback is to ask upstream to ship a real dynamic library alongside the extension module, which would let us go back to linking on every platform.

Fixes #457.

Issue #457, opened 2026-08-11
cpython issues on building package on mac


mschwamb:
Issues building on mac 

Building wheels for collected packages: layup, sorcha, sbpy
  Building editable for layup (pyproject.toml): started
  Building editable for layup (pyproject.toml): finished with status 'error'
  error: subprocess-exited-with-error
  
  × Building editable for layup (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [63 lines of output]
      WARNING: tool.scikit-build.metadata is deprecated; use the standard top-level [[tool.dynamic-metadata]] instead
      /private/var/folders/df/djsxfhc17x95674wsm_g8s980000gn/T/pip-build-env-ermpoiyo/overlay/lib/python3.13/site-packages/vcs_versioning/_backends/_git.py:431: UserWarning: "/Users/runner/work/layup/layup" is shallow and may cause errors
        pre_parse(wd)
      *** scikit-build-core 1.0.3 using CMake 4.4.0 (editable)
      *** Configuring CMake...
      loading initial cache file /var/folders/df/djsxfhc17x95674wsm_g8s980000gn/T/tmpb3s6hhb_/build/CMakeInit.txt
      -- The CXX compiler identification is AppleClang 21.0.0.21000101
      -- The C compiler identification is AppleClang 21.0.0.21000101
      -- Detecting CXX compiler ABI info
      -- Detecting CXX compiler ABI info - done
      -- Check for working CXX compiler: /usr/bin/clang++ - skipped
      -- Detecting CXX compile features
      -- Detecting CXX compile features - done
      -- Detecting C compiler ABI info
      -- Detecting C compiler ABI info - done
      -- Check for working C compiler: /usr/bin/clang - skipped
      -- Detecting C compile features
      -- Detecting C compile features - done
      Usage
      
        cmake [options] <path-to-source>
        cmake [options] <path-to-existing-build>
        cmake [options] -S <path-to-source> -B <path-to-build>
      
      Specify a source directory to (re-)generate a build system for it in the
      current working directory.  Specify an existing build directory to
      re-generate its build system.
      
      Run 'cmake --help' for more information.
      
      -- Found Python: /Library/Frameworks/Python.framework/Versions/3.13/bin/python (found version "3.13.14") found components: Interpreter Development.Module
      -- Performing Test HAS_FLTO_THIN
      -- Performing Test HAS_FLTO_THIN - Success
      -- Performing Test HAS_FLTO
[...]

PR #461, opened 2026-08-12
DIAGNOSTIC (do not merge): name the test that wedges the Ubuntu legs

matthewholman:
🔴 **Diagnostic only — do not merge.** Opened to make CI name a test it currently cannot. I will close it once we have the answer.

### What we are chasing

Since #456 and #459 landed, **both macOS legs on `main` pass** in ~13 minutes. Both **Ubuntu** legs stall dead — 9 minutes 41 seconds with no log output whatsoever — until the 20-minute cap kills them. Same behaviour on three legs across two days, so it is not a one-off.

### Why the normal run cannot tell us which test

- CI runs `pytest -n auto`, so progress output is in *completion* order across workers. A stalled worker cannot be tied back to a test. The two Ubuntu legs failing at different progress positions is scheduling noise, not two different tests. (I misread this earlier and named the wrong tests — disregard any names from before.)
- `pytest-timeout` is configured at 300 s and never fires. It uses `timeout_method = "signal"`, chosen deliberately because the thread method crashes xdist workers — and `SIGALRM` cannot interrupt a blocking C call. The stall runs 581 s, so whatever is stuck is not returning to the interpreter.

### What this branch changes

Linux only, one Python, and:

| flag | why |
|---|---|
| `-n0` | serial, so progress order is collection order |
| `-v` | one line per test, so the log names what was running when it wedged |
| `--timeout-method=thread` | *can* interrupt a blocking C call; dumps every thread's stack and kills the process. Rejected in `pyproject.toml` because it crashes xdist workers — with `-n0` there are none |
| `--timeout=180` | the stall exceeds 9 minutes, so 3 is ample and keeps the run short |

`timeout-minutes` goes to 45 because a serial run is several times slower than `-n auto`.

### Also still unexplained

There is a genuine test failure on the Ubuntu legs, separate from the hang, and it is unnamed for the same reason. This run should surface both.

Related: #457 (the macOS build fix that unblocked all this), #460 (cache key). Note #460's `timeout-minutes` 20 → 40 now looks like the wrong instinct — this is a wedge, not slowness, so a higher cap only doubles the cost of a hung run.

### DIFF
--- a/.github/workflows/testing-and-coverage.yml
+++ b/.github/workflows/testing-and-coverage.yml
@@ -14,8 +14,10 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        os: ['macos-latest','ubuntu-latest']
-        python-version: ['3.13', '3.14']
+        # DIAGNOSTIC BRANCH -- Linux only, one Python. macOS is green on main;
+        # this branch exists solely to name the test that wedges the Ubuntu legs.
+        os: ['ubuntu-latest']
+        python-version: ['3.13']
 
     runs-on: ${{ matrix.os }}
     # Fail a wedged leg fast instead of running to the 6-hour default. The unit
@@ -23,7 +25,9 @@ jobs:
     # is the recurring flaky-download hang (issue #388), which escapes
     # pytest-timeout when it happens in a spawned worker. Capping keeps reruns
     # cheap instead of burning ~2 hours per hang.
-    timeout-minutes: 20
+    # DIAGNOSTIC: serial runs are several times slower than -n auto, so the
+    # normal 20 would fire before we reach the hang.
+    timeout-minutes: 45
     steps:
     - uses: actions/checkout@v7
     - name: Set up - ${{ matrix.os }} - Python ${{ matrix.python-version }}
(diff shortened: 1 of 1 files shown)


matthewholman:
Answered. Closing.

**The serial run is green: 460 passed, 1 skipped, in 10m14s.** So there is no single hanging test — the wedge is a property of running in parallel on Linux, which is what this branch was built to find out.

That pointed straight at the cause: layup sizes its own process pool at `os.cpu_count()` whenever `num_workers` is negative, so under `pytest -n auto` each xdist worker spawns its own full pool — 4 workers × 4 children on a ~4-core runner, each re-importing layup and JAX under `spawn`. `tests/layup/conftest.py` already fixes the identical mistake one level down for OpenBLAS/OpenMP threads; the process level was never capped.

Fixed at the source in #462. The cache-key change is #460, with its `timeout-minutes` raise dropped — a higher ceiling does not help a wedge, it only doubles what each hung run costs.

The `--timeout-method=thread` change here turned out not to be the thing that cracked it, but it is worth remembering the trick: with `-n0` there are no xdist workers, so the objection recorded in `pyproject.toml` does not apply, and it is the only timeout method that can interrupt a blocking C call.

No part of this branch should be merged.

PR #464, opened 2026-08-15
Document LAYUP_NUM_WORKERS, and correct the --num-workers help

### LINKED ISSUE #463: add to documentation about overriding number of workers
Environment override for the automatic worker count, mirroring OMP_NUM_THREADS
# and friends: a caller that already knows how much of the machine it owns can
# say so without threading the value through every layup call.
_NUM_WORKERS_ENV = "LAYUP_NUM_WORKERS"

matthewholman:
Closes #463.

**The ask:** `LAYUP_NUM_WORKERS` (added in #462) existed only in a docstring, so a user had no way to find it. Adds a short README section and a `docs/parallelism.rst` page in the toctree.

**Also a defect #462 left behind:** all four CLI verbs — `orbitfit`, `convert`, `comet`, `predict` — still described `-1` as *"uses all available CPUs"*, which stopped being true. It now resolves to `$LAYUP_NUM_WORKERS` if set, else `1` when layup is already running inside another worker process, else the CPUs actually available (the affinity mask on Linux, so `taskset` and cgroup limits are respected).

Wrong help text is worse than missing documentation, so that part is a fix rather than an addition.

Text kept deliberately short in all three places, with the one thing a user actually needs to decide: set `LAYUP_NUM_WORKERS` when layup does not own the whole machine — your own process pool, or one of several jobs on a shared node — and note that it is separate from `OMP_NUM_THREADS`, which controls threads *within* a worker.

No behaviour change. CLI tests pass (36).

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,24 @@ method, non-gravitational parameters, parallel workers, …):
 layup orbitfit --help
 ```
 
+### Control how many CPUs layup uses
+
+`--num-workers` (CLI) and `num_workers=` (API) default to `-1`, meaning decide
+automatically: `$LAYUP_NUM_WORKERS` if set, otherwise 1 when layup is already
+running inside another worker process, otherwise the CPUs available to this
+process.
+
+Set `LAYUP_NUM_WORKERS` when layup does not own the whole machine — running it
+from your own process pool, or as one of several jobs on a shared node:
+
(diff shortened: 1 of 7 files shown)


Little-Ryugu (approved):
LGTM!

PR #468, opened 2026-08-24
Add the JOSS paper to the repo, and fix the install documentation

matthewholman:
Documentation and packaging only — **no changes to `src/` or `tests/`.** All items come from a mock JOSS review run against `main` today.

## 1. The paper is now in the repository

JOSS requires the paper and bibliography in the submitted repo. `git log --all --diff-filter=A -- '*paper.md'` returned nothing — they had only ever existed in a local directory and were never committed on any branch. Added as `paper/paper.md` + `paper/paper.bib` with the frontmatter updated to match.

`paper/README.md` records that **the bibliography now exists in two places** — here and in the AJ paper repository — and must be kept in step. Today's worked example: `bernstein2000`, the central method citation, carried a paraphrased title and the wrong given name for the second author ("Bhrigu" rather than "Bharat"). The DOI always resolved, so every automated check passed and nothing looked wrong.

## 2. The README named the wrong submodules

> "to download the required submodules, `assist`, `eigen`, and `rebound`."

`.gitmodules` declares **`include/eigen`** and **`include/autodiff`**. `assist` and `rebound` became pip wheels in #320, and **`autodiff` — a real submodule the build needs — was never mentioned anywhere**. A non-recursive clone fails with a bare `fatal error: 'Eigen/Dense' file not found`, with nothing pointing at the cause.

## 3. No prerequisites were stated

Added a requirements table: Python **≥3.11**, a **C++17** compiler, pip **≥21.3**, macOS/Linux only, ~3.2 GB disk.

Not hypothetical. The `python3` shipped with macOS is 3.9.6 with pip 21.2.4, and following the README verbatim gives:

```
ERROR: File "setup.py" or "setup.cfg" not found. Directory cannot be installed in
editable mode ... (A "pyproject.toml" file was found, but editable mode currently
requires a setuptools-based build.)
```

Nothing in that message says "your Python is too old". A reviewer doing a clean install hit exactly this. The README now names that failure and its fix.

Also: the two size figures disagreed — Setup said 3.2 GB, Quickstart said "a few hundred MB" for a download that is ~1 GB expanding to ~1.5 GB. Made consistent.

## 4. Added a "Running the tests" section

Noting that **`layup bootstrap` must run first**. Without it, a large fraction of the suite skips and the run reports success while validating none of the orbit fitting — 86 of 329 test functions are gated on the ephemeris.

## 5. Removed the template notebook

`docs/notebooks/intro_notebook.ipynb` was the unmodified LINCC template stub — *"Introducing Jupyter Notebooks in Sphinx"*, printing Sierpinsky triangles — and was listed **first** in the toctree, above both real tutorials. Verified the two remaining toctree targets still resolve.

## 6. Commented out the PyPI badge

The `layup` name on PyPI holds only **0.0.1, uploaded 2025-01-09 — four days before this repository's first commit**: a 1.3 kB wheel containing an empty `__init__.py`, with no dependencies. `pip install layup` succeeds and installs nothing, without error. The badge advertised that as though it were the package.

Commented rather than deleted, with the reason and the restore condition inline, pointing at #436.

## 7. Author names: Ruch and Holt

The JOSS paper had "Thomas Ruch" / "Carrie Holt"; the AJ paper had the initialled forms. Checked against Crossref rather than guessed:

- **Carrie E. Holt** — clear. Six of nine recent works, including **all four carrying her ORCID**.
- **Thomas R. Ruch** — thinner. Two publications, one each way; the initialled one is both more recent and from the *heliostack* collaboration that brings him onto this paper. ⚠️ **Worth confirming with him.**

Both ORCID *records* give the shorter forms, but that is a profile field rather than a byline. The two papers now carry identical author lists in identical order.

## Not addressed here

- `docs/index.rst` is still the LINCC template developer guide, so the published docs have no statement of need and no user install path. That needs prose, not a mechanical fix.
- The paper's quantitative claims (the MPC-catalog "one part in 10⁸" and incremental OD's "order of magnitude") still have no checkable artifact in the repository.
- The release path (#436) is untested; `publish-to-pypi.yml` has never run and has two identifiable faults.


mschwamb:
@matthewholman can you add the unit test the builds the JOSS paper. Helpful for co-authors to review it in PDF form before submission https://github.com/marketplace/actions/open-journals-pdf-generator

matthewholman:
Good idea — added in 68194f3.

`.github/workflows/draft-paper.yml` uses Open Journals' own draft action, so it is the same pipeline JOSS runs. It fires on any push or PR touching `paper/`, and can also be run on demand from the Actions tab (`workflow_dispatch`) to produce a PDF for circulation without editing anything. The PDF comes out as a build artifact named `paper` — download it from the run summary.

Deliberately not committing the PDF: it is derived, and a binary that changes on every edit makes for poor diffs. Say the word if you would rather have it in the tree for convenience.

Worth noting what it does and does not catch. It checks that the paper builds, the frontmatter parses, and every citation resolves against `paper.bib` — a failure here is a failure JOSS would hit. It does not check scientific content.

It would also have caught the class of problem we found earlier today: `bernstein2000` — the Bernstein–Khushalani citation, the paper's central method reference — had a paraphrased title and the wrong given name for the second author (*Bhrigu* rather than *Bharat*). The DOI always resolved, so every automated check passed and nothing looked wrong. Rendering the bibliography is what surfaces that.

mschwamb (approved):

Issue #445, opened 2026-07-22
BK dotted parameters mean time derivatives in bk_basis and scaled velocities in bk_iod; propose a single 1/gamma-scaled convention


matthewholman:
### Summary

The dotted Bernstein-Khushalani (BK) parameters `adot`, `bdot`, `gdot` mean two different things in two different files, and the two meet in one place. Nothing is producing wrong orbits today — the disagreement is suppressed to second order at the point of contact — but the protection is accidental and undocumented, and it will break silently if a reasonable future change is made.

This issue records the mismatch and proposes a single convention that fixes it and simplifies several formulas at the same time.

Related: #444 (a documentation error in the same header).

### Background, for anyone who has not been in this code

The BK parameterization describes an orbit in a frame tied to the observations rather than to the ecliptic. Layup builds an orthonormal frame `(a, b, n0)` where `n0` is the normalized mean of the observed lines of sight and `(a, b)` span the tangent plane. The object's position at the reference epoch is then described by two tangent-plane coordinates and one distance-like coordinate:

- `alpha` = x0/z0 and `beta` = y0/z0, the gnomonic tangent-plane coordinates
- `gamma` = 1/|r0|, one over the barycentric distance at the epoch

and three more parameters describe the motion. **It is those last three where the two conventions diverge.**

### The two conventions

**`bk_basis.cpp` / `bk_basis.h` — the dots are true time derivatives.**

The header states it (`bk_basis.h:12-13`, "(adot, bdot, gdot) are their time derivatives at the epoch") and both transforms implement it. `cartesian_to_bk` applies the quotient rule to `alpha = (rho_hat.a)/(rho_hat.n0)`:

```cpp
adot = (rho_dot.dot(fid.a) - alpha * rho_dot_n0) / u;
```

and `bk_to_cartesian` uses the chain rule, which only makes sense if the dots are derivatives:

```cpp
// d(rho_hat)/dt = adot*rho_hat_alpha + bdot*rho_hat_beta
// d(1/gamma)/dt = -gdot/gamma^2
```

**`bk_iod.cpp` — the dots are BK's scaled velocity components.**

The linear initial-orbit-determination model treats the tangent-plane coordinate as linear in time with `adot` as the slope (`bk_iod.cpp:22-23`):

```
x_obs_i = alpha + adot * t_i - gamma * X_i
y_obs_i = beta  + bdot * t_i - gamma * Y_i
```

[...]

PR #469, opened 2026-08-24
Restore the perspective denominator in BK-IOD (#446)

### LINKED ISSUE #446: BK-IOD drops the perspective denominator 1/(1 - gamma*ze); it is the dominant seed error (40% in the main belt)
`bk_iod.cpp`'s linear model drops the perspective denominator, and that — not any convention question — is the dominant error in the BK-IOD seed. Measured on the 98-case diagnostic scan (`diagnostic/scan/truth/`, 7 populations x 14 arc lengths), it is a 1.2% effect at 42 au and a 40% effect in the main belt, and it is what makes the main-belt seed fail outright.

Split out of #445, where it turned up while measuring something else.

## The omission

For straight-line motion the gnomonic observable satisfies exactly

```
x_obs = (alpha + adot*t - gamma*X) / (1 + gdot*t - gamma*ze)
y_obs = (beta  + bdot*t - gamma*Y) / (1 + gdot*t - gamma*ze)
```

with `X = r_obs . a`, `Y = r_obs . b`, `ze = r_obs . n0`. The code (`bk_iod.cpp:141-157`) fits the numerator alone and sets the denominator to 1, which is what makes the system linear in `(alpha, beta, gamma, adot, bdot)` and solvable in closed form. The header comment at `bk_iod.cpp:34-38` already identifies this as the regime limit; this issue is about fixing it rather than documenting it.

## How large it is

Maximum `|gamma * ze|` per population, i.e. the size of the term being set to zero:

| population | `gamma * ze` |
|---|---|
| sednoid 80 au | 1.2e-02 |
| scattered 70 au | 1.4e-02 |
| classical 42 au | 2.3e-02 |
| centaur 25 au | 3.8e-02 |
| centaur 15 au | 6.4e-02 |
| main belt 3.5 au | 2.8e-01 |
| main belt 2.5 au | 4.0e-01 |

This is essentially 1 au / \|r\|, so it does not become small until the outer solar system. The resulting seed error, median heliocentric distance error of `run_bk_iod` over each population's arc grid:

| population | median \|dr\|/r |
|---|---|
| scattered 70 au | 2.2% |
| sednoid 80 au | 2.3% |
| classical 42 au | 4.6% |
| centaur 25 au | 5.3% |
| centaur 15 au | 10.8% |
| main belt 3.5 au | 102% |
| main belt 2.5 au | 228% |

For comparison, every other discrepancy found in #445 is 6 to 160 times smaller than this term in every population. The header's own guidance ("main belt and closer: BK is essentially outside its regime of validity") is a consequence of this one approximation, not of anything intrinsic to the parameterization.

## Proposed remedy

The denominator does not have to be dropped — it only has to be known. The standard fix is to iterate:

1. Solve the linear system as now, giving `p0 = (alpha, beta, gamma, adot, bdot)` with `gdot = 0`.
2. Form `d_i = 1 + gdot*t_i - gamma*ze_i` from the current estimate.
3. Re-solve the same weighted system with each row scaled by `d_i` — equivalently, fit `d_i * x_obs_i` against the unchanged numerator basis. Still linear, still closed form, same `colPivHouseholderQr` solve.
4. Repeat 2-3 a small fixed number of times.

Each pass is one more 5x5 solve on an already-assembled design matrix, so the cost is negligible next to the LM fit this seeds. Two or three passes should be plenty given `gamma*ze <= 0.4`; the iteration is a contraction whenever `|gamma*ze| < 1`, which fails only for an object inside roughly 1 au.

The weights need care: scaling row `i` by `d_i` scales its residual too, so the weight should become `w_i / d_i^2` to keep the chi-square meaning unchanged. Worth getting right, since the returned `cov` is used as the seed covariance.

## What to measure

Same harness as #445 — `tools/bk_iod_sweep.py` over `diagnostic/scan/truth/`:

- median and worst-case `|dr|/r` per population per arc length, before and after;
- whether the main-belt cases become usable at arcs beyond 1 day (the header table says they currently fail catastrophically at 2 days and up);
- iteration count to convergence of the denominator loop, and confirmation that it does not diverge for the closest population;
- downstream effect on the LM: does a better seed reduce `run_bk_native_fit` iterations or failure rate.

## Relationship to #444 and #445

Independent. #445 is a parameterization-consistency change that by construction cannot move a fitted result; this one changes the seed's accuracy and should. They touch the same file but not the same lines, and either can land first. If #445 lands first, the identity above is already the form the code works in, which makes this change smaller.

### LINKED ISSUE #465: Unusually long orbitfit execution time for certain MBAs
LINUX ONLY

When running orbitfit for the MBA 1119 T-1 (using only its Rubin DP2 r band observations; if needs be I can send these to anyone who is a rubin data rights holder), it stalls for an unusually long time. Total execution takes ~6 minutes. This object is a normal MBA on a very circular orbit (e=0.04), there are 14 observations being used with an arc of over 1 month (2025-07-18 to 2025-08-27). During execution, 3 errors appear:

```Error! The requested time is outside the coverage provided by the ephemeris file.```
This error repeated tens of times

```Error: cannot interpolate beyond timestep bounds (h=-nan).```
This error repeats 4 times, where h is a different value each time

```Warning! At least 10 predictor corrector loops in IAS15 did not converge. This is typically an indication of the timestep being too large.```
This error appears once.

However, these errors are not recorded in the .log or .err files. I believe these errors come from ASSIST. This object does not noticeably stall on MacOS, and fits without error.

MACOS AND LINUX

When running orbitfit for the MBA 2015 PE134 (again only its Rubin DP2 r band observations), it also stalls but for different times on Mac and Linux. On Mac it takes 10 minutes to execute, on Linux it takes 40 minutes, and in both cases the fitting fails (flag 3). However, running with the IOD `auto` takes twice as long but does successfully fit. Again this object is an MBA with a slightly elliptical (e=0.11) orbit, with 14 observations spanning more than 1 month (2025-07-01 to 2025-08-15). 

No errors appear for this object. 

For both of these objects, I would have expected them to be fit a lot faster, or error out faster if they can't be fit. 

Some environment info (for both OS's):

OS = MacOS Tahoe 26.6.2; Ubuntu 22.04.5

```python=3.14.3
layup=0.1.dev722+gc860acffe (Ubuntu) 0.1.dev708+gbec27479c.d20260727 (MacOS)
sorcha=1.2.1
assist=1.2.3
rebound=4.5.0

### PR DESCRIPTION
**Pull request by matthewholman**
Closes #446. Two commits: the fix, and a test-fixture repair that had to come first.

## The fix

The exact gnomonic relation carries a denominator that `bk_iod.cpp` dropped:

```
x_obs = (alpha + adot*t - gamma*X) / (1 + gdot*t - gamma*ze)
```

The omitted factor is `|gamma*ze| ~ 1 au / r_helio` — negligible for TNOs, **0.29 at 3.5 au and 0.40 at 2.5 au**, and per #446 the dominant seed error in the inner solar system.

**It does not make the system nonlinear.** Multiplying through and collecting the unknowns:

```
x_obs = alpha + adot*t - gamma*(X - x_obs*ze) - gdot*(x_obs*t)
```

Every unknown is still first order, because `x_obs` is *data* — so `x_obs*ze` is a known coefficient. The fix is one changed coefficient per row:

```cpp
A(2*i, 2) = -X_i;                    // before
A(2*i, 2) = -(X_i - x_obs * ze);     // after
```

#446 proposed iterating to handle the denominator. I prototyped that first ([comment](https://github.com/Smithsonian/layup/issues/446#issuecomment-5394809429)) and it works, but **diverges above `|gamma*ze| ~ 0.29`** — i.e. across the whole main belt, precisely where it was wanted. The closed form has no such failure mode.

### Measured, through `run_bk_iod`

Synthetic 10-day arcs, before → after:

| population | before | after |
|---|---|---|
| main belt 2.5 au | 53.8% | **8.4%** |
| main belt 3.5 au | 44.7% | **4.3%** |
| centaur 15 au | 15.5% | **0.15%** |
| classical 42 au | 6.3% | **0.00%** |
| sednoid 80 au | 3.4% | **0.01%** |

The main belt does not reach sub-percent and should not be expected to: what remains once the perspective term is handled is the straight-line no-gravity assumption itself, a real limit at 2.5 au over 10 days. But a few percent is a seed the LM can start from, where 45–54% is not.

### Also here

A **weight-refinement pass**. Multiplying by the denominator scales each residual by `d_i`, so the implied weighting is `w_i*d_i^2`; re-solving with `w_i/d_i^2` restores the intended chi-square. It moves the point estimate by ~1e-4 relative — immaterial — but the returned **covariance is used as the seed covariance downstream**, which is where it matters. Skipped if any `d_i` is non-positive.

### `gdot` is deliberately left pinned

It enters the relation linearly through `-(x_obs*t)`, so freeing it is one extra column and improves every population again (main belt 2.5 au 8.4% → 5.5%). **Blocked on #445**: this file's dots are BK's scaled velocities while `bk_basis`'s are true time derivatives, and that mismatch is invisible *only* because `gdot` is zero. Freeing it is exactly the "reasonable future change" #445 warns will make it break silently. Documented in the header.

## The tests actually guard the fix

Five parametrized cases from 2.5 to 80 au, plus a covariance sanity check. I reverted only the C++ — keeping the tests — rebuilt, and confirmed **all five fail with exactly the pre-fix errors** (0.5384, 0.4467, 0.1549, 0.0627, 0.0344), each naming the cause in its assertion message.

They use a synthetic two-body arc built in-process, so they run everywhere rather than being gated on the ephemeris.

## Second commit: the truth set shipped 10 of 98 cases

`_bk_guards.py` has always documented "98 ASSIST-integrated cases" and `test_bk_everywhere`'s sweep helper says "all 98", but `tests/data/bk_scan_truth.json` held **10**. Any test naming a case outside that subset failed with a bare `KeyError` rather than skipping — which is how `test_iod_auto.py::test_iod_auto_recovers_when_gauss_fails` failed the moment it was run with `LAYUP_SLOW_TESTS=1`, on patched and unpatched builds alike.

Regenerated from the diagnostic scan with the same trimming, 13 kB → 138 kB. **Verified faithful**: all 10 previously-shipped cases round-trip byte-identical, so this only adds coverage.

(Found in a mock JOSS review, which noted a reader of that guard would substantially overestimate BK coverage.)

## ⚠️ This probably does not close #465

Worth stating plainly so nobody assumes otherwise:

- **#465's failing path is `iod='gauss'`**, which never reaches BK-IOD. The stall there is the picker running phantom *Gauss* roots through LM.
- The grind now looks **platform-linked**. Adam marks 1119 T-1 "LINUX ONLY" and fitting fine on macOS; `test_iod_auto.py:119` names Linux nondeterminism explicitly; and with both blockers above cleared, it still will not reproduce on macOS — the test skips because **Gauss succeeds** there, so the fallback is never reached.

A candidate axis is the `if(APPLE)`/`else()` split in `CMakeLists.txt`: Linux links `libassist`/`librebound` directly, macOS `dlopen`s them with `RTLD_GLOBAL` (#459). Different code path in the library that is grinding, and one no Mac can exercise.

The discriminating experiment — a seed-quality → integrator-work curve on Linux, with seeds perturbed both inward and outward — is written up for a USDF session.

## Checks

79 passed, 1 skipped across `test_bk_iod`, `test_bk_basis`, `test_bk_fit`, `test_bk_everywhere`, `test_iod_picker`, `test_iod_auto`. Black clean. Built and measured in a clean venv against this tree, not the local dev venv.


kjnapier (approved):
Looks okay to me

Issue #471, opened 2026-08-24
PyPI publishing will fail: direct-URL sorcha dependency, and publish workflow does not init submodules


matthewholman:
Two faults that will make the first PyPI upload fail. Neither is tracked anywhere — they surfaced in `RELEASE_CHECKLIST.md` and in a mock JOSS review, and were being informally filed under #436, which does not actually cover them (#436 is the stale 0.0.1 stub and a bootstrap leftover).

Both confirmed on `main` @ `eba34b2`.

**Not blocking the JOSS submission.** The checklist's critical path is finalize → tag → GitHub release → Zenodo DOI → submit, and PyPI is explicitly parallel to it. This is about not discovering these two on the day someone tries to publish.

## 1. The `sorcha` dependency is a direct URL, which PyPI rejects

`pyproject.toml:24`:

```toml
"sorcha @ git+https://github.com/dirac-institute/sorcha.git",
```

PyPI refuses any distribution whose metadata carries a direct-URL `Requires-Dist`. The upload fails at validation, after the build.

sorcha is on PyPI (1.2.1 latest), so this becomes `sorcha>=1.2.0`. Evidence the released version works: 1.2.0 is installed in a local dev environment and layup imports and fits against it, and @awilson110 reports running 1.2.1 with layup on both macOS and Ubuntu in #465.

⚠️ It still wants a clean-environment verification before landing — layup currently tracks sorcha `main`, which may be ahead of the release in ways the test suite would notice. That check is the actual work here; the edit is one line.

While in there: **`pytest` is listed in `[project].dependencies`** as well as in the `dev` extra, so every user installs a test framework at runtime. Worth dropping from the runtime list.

## 2. `publish-to-pypi.yml` has never run, and will fail on the submodules

The workflow has **0 runs**. At line 36 it calls:

```yaml
      run: python -m build
```

with no `git submodule update --init` beforehand. `CMakeLists.txt` references `include/eigen` and `include/autodiff`, both git submodules, so the C++ extension build fails with a bare missing-header error:

```
fatal error: 'Eigen/Dense' file not found
```

(That exact failure was reproduced by following the docs' non-recursive install path, so it is not hypothetical.)

Fix: `submodules: recursive` on the `actions/checkout` step, or an explicit init before the build.

## Also worth deciding here

[...]

PR #472, opened 2026-08-24
Remove the extracted archive after bootstrap untars it (#436)

matthewholman:
Addresses the second item of #436. The first item — the stale PyPI 0.0.1 stub — is downstream of the first real release and stays open; the two faults that will actually make that release fail are now filed separately as #471.

## The problem

`layup bootstrap` untars `debias_hires2018.tgz` (~156 MB) and leaves the archive sitting beside the extracted contents. Nothing reads it again, so it roughly doubles that part of the cache — on a package that already asks the user for about 3.2 GB.

## The fix

Remove it after extraction, best-effort:

```python
pooch.Untar(extract_dir=".").__call__(fname, action, pup)
try:
    os.remove(fname)
except OSError as exc:
    print(f"Could not remove the extracted archive {fname}: {exc}")
```

The `try` is deliberate. A read-only cache, or a platform still holding the handle, should cost disk space — not abort a multi-hundred-megabyte download that has already succeeded.

## Tests

Two, both network-free: they build a small tarball in a `tmp_path` and call the same `_decompress` hook pooch invokes.

- `test_decompress_removes_the_tar_archive` — asserts the archive is gone **and** the extracted contents survive. Verified to fail on the unpatched code with `AssertionError: the .tgz archive should be removed after extraction`.
- `test_decompress_survives_an_unremovable_archive` — monkeypatches `os.remove` to raise, and asserts extraction still completes. This pins the best-effort behaviour rather than leaving it to be inferred from the `try`.

### DIFF
--- a/src/layup/utilities/bootstrap_utilties/download_utilities.py
+++ b/src/layup/utilities/bootstrap_utilties/download_utilities.py
@@ -159,6 +159,15 @@ def _decompress(fname: str, action: str, pup: pooch.Pooch) -> None:  # pragma: n
     if os.path.splitext(fname)[-1] in tar_extentions:
         print(f"Trying to decompress tar file: {fname}")
         pooch.Untar(extract_dir=".").__call__(fname, action, pup)
+        # Remove the archive once its contents are extracted (issue #436).  The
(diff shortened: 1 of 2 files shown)


kjnapier (approved):
Should be fine

PR #474, opened 2026-08-24
JOSS paper: five factual corrections before submission

matthewholman:
Five factual corrections to `paper/paper.md` before submission. **One has been in the paper for a month.**

Branches off #473, so it includes that change too — merge #473 first, or merge this and close that.

## 🔴 The Bernstein-Khushalani basis does not do energy-constrained 5-parameter fits

```diff
- a Bernstein-Khushalani basis that also supports energy-constrained 5-parameter fits.
- Both share the same internal numerical-integration and observation-modeling framework
+ a Bernstein-Khushalani basis. Both have six parameters and share the same internal
+ numerical-integration and observation-modeling framework
```

The BK fit is a **full six-parameter fit with a bound-orbit energy *prior* on `gdot`** — `bk_fit.cpp:65-69` sets `P_prior(5,5) = 1/sigma_gdot^2`, and all six parameters are still estimated. Nothing is constrained away. The genuine five-parameter object is BK-IOD, which the paper already describes separately, so as written the two passages said "5-parameter" about different things.

**This was corrected in the AJ paper on 2026-07-22 and never propagated back.** The wording now matches it.

## Four smaller ones

**Sign error.** One *minimizes* chi-square but *maximizes* log-likelihood; the sentence said "minimizing the chi-square or log-likelihood function". Now "minimizing the chi-square (equivalently, maximizing the log-likelihood)".

**`giorgini1996` was cited as evidence it cannot support.** It appeared as backing for "Neither OrbFit nor OpenOrb matched JPL Horizons in detailed comparisons" — but it is a DPS abstract describing the *Horizons service itself* and contains no such comparison. Only `chernyavskaya2021` supports that claim. Moved to where Horizons is named.

**An overreach a reviewer will push back on.** "None is written for Python-based, LSST-scale pipelines" — OpenOrb ships `pyoorb`. Softened to "none is designed as a Python-native library for LSST-scale batch processing", which is the defensible version and what the rest of the paragraph actually argues.

**"ASSIST includes all the terms in the equations of motion"** is overbroad. Now names them: planetary and major-asteroid perturbations, general-relativistic corrections, solar oblateness.

## Recorded, not done — these need prose

Four items are noted in a comment block in the source rather than acted on, since they are writing rather than correction:

1. **Length**: ~1700 body words against JOSS's 250–1000 guidance. The trim is the Summary/Functionality overlap — both list the ingest formats, both list the two parameterizations, both introduce IOD (defining the acronym twice), both state that every fit returns a covariance.
2. **`adam_core` / THOR is absent from the state of the field** — the closest competitor to the niche claimed here, and two of its principals are co-authors.
3. **Two quantitative claims a reviewer cannot check**: the catalog re-fit's "one part in 10^8", and incremental OD's "order of magnitude less time" (no benchmark exists; `benchmarks/README.md` lists the matching one as planned).
4. **Three bibliography entries cited by neither paper** — `veres2017` and the two Anthropic ones.

## Checks

Frontmatter parses, all 27 citations resolve, and the `Draft JOSS paper` workflow builds the PDF on this PR.


matthewholman:
Added one more, after comparing the two papers sentence by sentence — prompted by the BK 5-parameter claim having drifted between them for a month.

**Murtagh's acknowledgement was incomplete here.** The AJ paper credits the DfE Northern Ireland studentship *and* travel support from the STFC for UK participation in LSST through grant ST/S006206/1; this paper had only the first half. Now matches exactly.

The same comparison found a gap in the other direction, which is **not** in this PR because acknowledgements are the author's own text: this paper acknowledges Wilson's studentship under STFC grant UKRI177, and the AJ paper has nothing for him. @awilson110 — worth adding yours to `sections/16_acknowledgments.tex` in the Overleaf project.

The two AI Usage Disclosures also disagree and are likewise left alone here. This paper discloses code conversion from Rust to C++ and names ChatGPT alongside Claude; the AJ version omits both. Whichever is accurate, they should match before either is submitted.

Other than those, the papers now agree everywhere they share prose — the remaining differences are LaTeX artifacts and punctuation.

kjnapier (commented):
No notes from me

kjnapier (approved):

PR #476, opened 2026-08-24
Adopt the 1/gamma-scaled BK dot convention (#445, #444)

### LINKED ISSUE #444: Comment in bk_basis.h says gamma is inverse heliocentric distance; the code uses barycentric
### Summary

The documentation comment on `BKState` contradicts itself, and the half that is wrong is the half a reader is most likely to copy. This is a documentation bug only — the code is self-consistent and no numerical result is affected.

In `src/lib/orbit_fit/bk_basis.h`, the comment block describing the Bernstein-Khushalani (BK) parameters says:

```cpp
// Bernstein-Khushalani parameters at epoch.  Origin: barycenter.     <-- line 9
// (alpha, beta) are gnomonic tangent-plane coordinates of the
// line-of-sight direction rho_hat at a fiducial direction n0,
// gamma = 1/|r_helio|, and (adot, bdot, gdot) are their time          <-- line 12
// derivatives at the epoch.
```

Line 9 says the origin is the solar system barycenter. Line 12 says gamma is one over the *heliocentric* distance (the distance from the Sun). Those cannot both be true.

### What the code actually does

`cartesian_to_bk` in `src/lib/orbit_fit/bk_basis.cpp` takes the state it is handed and forms gamma from its position vector directly:

```cpp
const Eigen::Vector3d r = cart.head<3>();
const double r_norm = r.norm();
const double gamma = 1.0 / r_norm;
```

The state passed in is barycentric, as line 9 states. So **gamma is the inverse barycentric distance**, and line 12 is the part that is wrong.

### Why it is worth fixing

The two distances differ by the Sun–barycenter separation, which reaches roughly 0.007 au. That is negligible for a trans-Neptunian object at 40 au (a few parts in 10,000), but it approaches a percent for an object near 1 au — small, yet large enough to matter to anyone who takes the header at its word and mixes the two in downstream analysis.

The more immediate cost is that this comment is the natural place to look to find out what the BK parameters mean. It was nearly copied into the methods paper as the definition of gamma, which is how it came to light.

### Suggested fix

Change line 12 so that it agrees with line 9 — for example, `gamma = 1/|r|`, the inverse barycentric distance. Renaming anything is optional; correcting the comment is the substance.

This appears to be the only misleading occurrence. The other uses of `r_helio` in the tree (order-of-magnitude comments in `bk_iod.cpp`, and `min_r_helio_AU` in the initial-orbit-determination root picker) look like genuine heliocentric quantities and are unaffected.

### Related observation, not necessarily a bug

Flagging this while the area is open rather than filing it separately. The bound-orbit energy prior on `gdot` is handed `MU_SUN` — the heliocentric gravitational parameter GM_sun = k², from `constants.py` — and applies it to the barycentric gamma above (`orbitfit.py`, in `run_bk_native_fit`). In other words, the prior evaluates the heliocentric two-body energy at the barycentric distance.

This is the standard approximation for this parameterization and is almost certainly harmless, since the Sun dominates the mass. But it is an approximation, and a one-line comment saying so would spare the next reader the same double-take. Happy to open it separately if it is worth tracking on its own.

### LINKED ISSUE #445: BK dotted parameters mean time derivatives in bk_basis and scaled velocities in bk_iod; propose a single 1/gamma-scaled convention

matthewholman:
Closes #445. Closes #444.

The Bernstein-Khushalani dots were the plain time derivatives of the coordinates. They are now the velocity components in the orthonormal fiducial basis, scaled by gamma:

```
adot = γ (v · a),   bdot = γ (v · b),   gdot = γ (v · n₀)
```

This makes `gdot` the line-of-sight velocity — the parameter angles-only astrometry constrains least, and the one the bound-orbit energy prior exists to regularize. Under the old convention it was the barycentric radial rate, a different quantity except when n₀ ∥ r̂.

Three things fall out, and the diff is net −60 lines in `bk_basis.cpp`:

- the velocity becomes diagonal in the fiducial basis;
- `dcart_dbk` loses its entire second-derivative block, since v no longer depends on α or β;
- `sigma_gdot_sq` collapses to `2μγ³ − adot² − bdot²`, with no (α, β) dependence.

`bk_fit.cpp` needed no change — it works through the transforms generically.

**BK-IOD.** The exact relation carries a factor `s = √(1+α²+β²)` on three columns. Undoing it needs no iteration: scaling a column and dividing the matching parameter by the same factor is an exact least-squares identity, so α and β come out already correct and one division finishes it. Seed error over the 98-case diagnostic set improves from a median of 0.0126 to 0.0091, better in 59 of 84.

**#444** is fixed in the same comment block: γ was documented as `1/|r_helio|` two lines below a struct that says barycenter.

**Regression.** Fitted orbits move by at most 1.6e-5 au against a formal σ of 577 au, and covariances are unchanged to 1.6e-6 relative. Confirmed across 307 real short arcs — see the comment below.

**One test threshold moved.** `test_bk_native_fit_recovers_known_state` required `csq < 1e-12`, now `1e-6`. Its truth states are circular at epoch, so the old `gdot = −γ²(r̂·v)` vanished identically and the energy prior contributed nothing; the threshold was calibrated on that accident. The new `gdot` does not vanish, and `csq` floors at the prior evaluated at truth — which matches the converged value to seven digits, so the fit still lands on truth. The state-recovery assertion is unchanged and is what pins it.

**Found but not fixed here:** `converged()` (`orbit_fit.cpp:815`) is an absolute test applied to parameters spanning seven decades, so it is not invariant under reparameterization. Filed as #477; it is shared with the Cartesian fit, so it does not belong in this PR.

### DIFF
--- a/src/lib/orbit_fit/bk_basis.cpp
+++ b/src/lib/orbit_fit/bk_basis.cpp
@@ -97,27 +97,20 @@ namespace orbit_fit
     {
         const RhoFrame f = compute_rho_frame(bk.alpha, bk.beta, fid);
         const double inv_g = 1.0 / bk.gamma;
-        const double inv_g2 = inv_g * inv_g;
 
         const Eigen::Vector3d r = inv_g * f.rho_hat;
 
-        // v is the exact time-derivative of r = rho_hat / gamma.  Using the
-        // chain rule d(rho_hat)/dt = adot*rho_hat_alpha + bdot*rho_hat_beta
-        // and d(1/gamma)/dt = -gdot/gamma^2:
+        // Under the 1/gamma-scaled convention (issue #445) the dots ARE the
+        // velocity components in the orthonormal fiducial basis, scaled by gamma:
+        //   adot = gamma (v . a),  bdot = gamma (v . b),  gdot = gamma (v . n0)
+        // so the inverse is three terms in an orthonormal basis:
+        //   v = (1/gamma) (adot a + bdot b + gdot n0)
         //
-        //   v = (1/gamma) d(rho_hat)/dt + d(1/gamma)/dt * rho_hat
-        //     = (1/gamma)(adot*rho_hat_alpha + bdot*rho_hat_beta)
-        //       - (gdot/gamma^2) rho_hat
-        //
-        // Dimensional check (lengths [L], time [T]): gamma ~ 1/[L], so
-        // 1/gamma ~ [L]; rho_hat_alpha/beta are dimensionless and adot,bdot
-        // ~ 1/[T], so term 1 ~ [L]/[T].  gdot ~ 1/([L][T]) and 1/gamma^2 ~
-        // [L]^2, so term 2 ~ [L]/[T] as well -- both are velocities.  The
-        // magnitudes look unusual only because rho_hat_alpha/beta are
-        // deliberately not unit length (see RhoFrame above), not because the
-        // units are inconsistent.
-        const Eigen::Vector3d v = inv_g * (bk.adot * f.rho_hat_alpha + bk.bdot * f.rho_hat_beta)
-                                  - bk.gdot * inv_g2 * f.rho_hat;
+        // The gnomonic tangent vectors rho_hat_alpha, rho_hat_beta -- which are
+        // deliberately not unit length and were a recurring source of confusion --
+        // drop out of the velocity path entirely.  All three dots now share units
+        // of inverse time.
+        const Eigen::Vector3d v = inv_g * (bk.adot * fid.a + bk.bdot * fid.b + bk.gdot * fid.n0);
 
         Eigen::Matrix<double, 6, 1> cart;
         cart << r, v;
@@ -139,17 +132,12 @@ namespace orbit_fit
         const double alpha = rho_hat.dot(fid.a) / u;
         const double beta = rho_hat.dot(fid.b) / u;
 
-        // gdot = d/dt (1/|r|) = -(r . v) / |r|^3 = -gamma^2 * (rho_hat . v)
-        const double gdot = -gamma * gamma * rho_hat.dot(v);
-
-        // d(rho_hat)/dt = gamma * (v - (v . rho_hat) * rho_hat)  -- v's component perp to rho_hat,
-        // scaled to a sphere tangent vector.  alpha-dot, beta-dot come from
-        // applying the quotient rule to alpha = (rho_hat . a) / (rho_hat . n0)
-        // and similarly for beta.
-        const Eigen::Vector3d rho_dot = gamma * (v - v.dot(rho_hat) * rho_hat);
-        const double rho_dot_n0 = rho_dot.dot(fid.n0);
-        const double adot = (rho_dot.dot(fid.a) - alpha * rho_dot_n0) / u;
-        const double bdot = (rho_dot.dot(fid.b) - beta * rho_dot_n0) / u;
+        // Under the 1/gamma-scaled convention (issue #445) the dots are simply the
+        // velocity components in the orthonormal fiducial basis, scaled by gamma.
+        // No quotient rule, no tangent-vector norms.
+        const double adot = gamma * v.dot(fid.a);
+        const double bdot = gamma * v.dot(fid.b);
+        const double gdot = gamma * v.dot(fid.n0);
 
         BKState bk;
         bk.alpha = alpha;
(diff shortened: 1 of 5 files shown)


matthewholman:
Filed the convergence-test fragility noted above as #477, so it does not get lost with this PR. It is deliberately not fixed here.

matthewholman:
Checked this against 307 real short arcs, not just the two test cases: sane-fit counts are identical between `main` and this branch in every regime, for both IOD settings.

|  | gauss main / #476 | auto main / #476 |
|---|---|---|
| 3-day (85) | 35 / 35 | 36 / 36 |
| 7-day (106) | 60 / 60 | 63 / 63 |
| 14-day (116) | 89 / 89 | 91 / 91 |

The gauss column is the control — it does not touch the BK basis, so its being identical also confirms the four merges this branch is behind `main` have no effect on fitting. No timing cost either: 6 s vs 5 s on the 3-day group.

Arcs were cut from the cached MPC corpus, 4-40 observations each. "Sane" means `flag == 0` and bound and `0.5 < a < 100` au and under 3x circular.

This is evidence the change is safe, not that it helps. On this population the fit reaches the same place from either seed; the seed-accuracy improvement in the description is the evidence for the other half.

bernardinelli:
The math looks correct to me, and the explanations makes sense! Since there have been successful tests, I think we're good to go!

bernardinelli (approved):

PR #466, opened 2026-08-24
Add the four community files required for the JOSS release

matthewholman:
Adds the four community files that `RELEASE_CHECKLIST.md` lists as Phase 0 items and that the JOSS submission bot checks for. `LICENSE` (MIT) was already present; these were the remaining four.

Closes the community-file half of the pre-release checklist. No code changes.

### What is here

| file | notes |
|---|---|
| `CODE_OF_CONDUCT.md` | Contributor Covenant 2.1, verbatim. Enforcement contact set to mholman@cfa.harvard.edu. |
| `CITATION.cff` | All 18 authors with ORCIDs, taken from the AJ paper author block. YAML-validated. |
| `CHANGELOG.md` | Keep a Changelog format, with an `[Unreleased]` section. |
| `CONTRIBUTING.md` | How to report a bug, install for development, and open a pull request. |

### Two things left deliberately incomplete

**`CITATION.cff` has no `version`, `date-released` or `doi`.** These cannot be filled until the release is tagged and Zenodo mints the DOI — checklist Phase 4 wires them back afterwards. The TODO is marked in the file.

**`CHANGELOG.md` describes the initial release by capability, not by history.** There are 235 merged pull requests behind v1.0; enumerating them would not help anyone. At tag time, promote `[Unreleased]` to `[1.0.0] - <date>`.

### Worth a look when reviewing

- **The author list in `CITATION.cff`** — please check your own entry. It is taken from the AJ paper, so anyone who has corrected their ORCID or spelling there should be correct here, but it is worth each author's eye.
- **`CONTRIBUTING.md` asks for evidence with numerical changes** — a finite-difference comparison, a Horizons cross-check, or a Monte-Carlo test. That paragraph is there because the test suite passes happily against a subtly wrong Jacobian, which is how several real bugs reached `main`. Reword it if it reads as too strong a demand.
- **The Code of Conduct contact.** A personal address is the simplest thing that works, but if the project would rather route reports somewhere shared or institutional, now is the moment to change it.

### Not in this PR

The `sorcha @ git+https://...` dependency (#436) still needs fixing before PyPI will accept a build — PyPI forbids direct-URL dependencies. sorcha 1.2.0 and 1.2.1 are both on PyPI and both appear to work with layup, so the change is likely one line, but it deserves a clean-environment verification rather than being folded in here.

Issue #440, opened 2026-07-16
review/update of read the docs documentation


mschwamb:
review/update of read the docs documentation

matthewholman:
Reopening: this was shut automatically when #478 merged, and it should not have been.

**Why it happened.** The commit message on `4e1b50f5` contains a sentence of the form *"this does not <closing-keyword> #440, which is a broader review"*. GitHub's parser reads the closing keyword immediately before the issue number and ignores the negation wrapped around it, so the sentence saying the issue should stay open is what shut it. The pull request *description* had the same phrasing and was reworded before merge — the commit message was not, and keywords are honoured there too.

Worth knowing generally: writing *about* a closing keyword next to an issue number is enough to trigger it, in a PR body or a commit message.

**What is genuinely done, and what is not.**

#478 was a first pass: it replaced the LINCC template boilerplate with a real landing page — what Layup is, installation, a runnable quickstart, and navigation — and moved the contributor bootstrap to its own page. #479 separately fixed the docstrings that were rendering broken in the API reference (18 docutils warnings to 0).

Still open under this issue:

- **The autoapi API reference has ~84 `python_import_resolution` warnings.** Separate from the docstring formatting #479 fixed; these are unresolved cross-module imports and they degrade the generated pages.
- **The notebooks have not been reviewed.** `docs/notebooks/orbit_fitting_api.ipynb` and the pre-executed visualizer demo predate most of this year's changes.
- **The install instructions must change when the first release ships.** The landing page currently warns readers *off* `pip install layup`, because that command installs a 1.3 kB placeholder. That warning becomes wrong the moment a real version is published, and the two need to move together.
- **Log files land in the working directory with no way to redirect them.** Raised in #481 and not addressed by the change that shut it; worth deciding alongside `LAYUP_CACHE_DIR` (#448, PR #484) rather than inventing a second mechanism.

Whoever picks this up should treat the landing page as done and start from the list above.

PR #478, opened 2026-08-24
Give the documentation a landing page (#440)

matthewholman:
Toward #440 — a first pass, not the whole issue. **#440 deliberately stays open**; see the last section. (Worded carefully on purpose: GitHub's auto-close parser reads a closing keyword sitting immediately before an issue number and ignores any negation wrapped around it, so the obvious phrasing would have shut the issue on merge.)

## The problem

The published front page at [layup.readthedocs.io](https://layup.readthedocs.io/en/latest/)
is still the LINCC project template's boilerplate. Its only heading is
**"Dev Guide - Getting Started"** and its first instruction is `conda create`.

It never says what Layup is, how a user installs it, or shows it doing
anything. A reader arriving from the paper or from the repository lands on
contributor bootstrap instructions for a package whose purpose is never stated.

This matters now because it is one of the things a JOSS reviewer checks
directly: whether the documentation states the need, gives installation
instructions, and includes example usage.

## The fix is mostly assembly, not writing

Everything else in the docs already exists and builds: the autoapi API
reference, the notebooks, the parallelism page. And the missing content was
already written — the README has a requirements table and a three-part
quickstart, and the JOSS paper has a statement of need that has been through
review. None of it was ever wired into the landing page.

So this PR assembles a real one:

- **what Layup is**, and the four observation types it ingests, condensed from
  the paper's summary;
- **installation**, with the requirements table from the README;
- **a quickstart** — `layup bootstrap`, then the bundled (3666) Holman demo
  through `orbitfit`, `convert` and `predict`;
- **where to go next** — the API notebook, parallelism, the API reference, and
  the developer guide.

The developer setup that used to *be* the front page is not deleted, just moved
to `docs/dev_guide.rst` and labelled for contributors, which is what it always
was.

## Two things worth your attention

**1. The install section deliberately does not say `pip install layup`,** and
carries a warning box saying so.

That command currently *succeeds* and installs a 1.3 kB placeholder: the
`layup` name on PyPI holds version 0.0.1, uploaded 2025-01-09 — four days
before this repository's first commit — containing an empty `__init__.py` with
no dependencies. A reader following it would get a silent no-op, then find that
nothing works, and reasonably conclude the package is broken. That is worse
than giving no instructions at all.

This is the same fact behind the PyPI badge being commented out of the README.
The section should change to `pip install layup` **in the same PR that
publishes the first real release** (#436 / #471), not before and not after.

**2. The Sphinx dependencies are in `docs/requirements.txt`, not the `dev`
extra.** `pip install -e .'[dev]'` does not let you build the documentation,
though the old front page implied it did (it told you to install pandoc so that
notebook rendering "works as expected"). The new developer guide says where
they actually live.

## Testing

Built locally with Sphinx. Both new pages render with no warnings of their own,
and the toctree resolves. The notebooks need `pandoc`, which is not installed
on this machine, so that part of the build was excluded from the check — it is
untouched by this change, and CI builds it.

## Deliberately left out

- **Community-file links.** `CONTRIBUTING.md`, `CODE_OF_CONDUCT.md` and
  `SUPPORT.md` do not exist on `main` yet; they arrive with #466. An RST
  comment marks the spot so the link is a one-line follow-up rather than a
  rediscovery.
- **A citation.** The "Citing Layup" section is a placeholder pointing at the
  first release, since there is no DOI yet.

## Why #440 stays open

#440 is "review/update of read the docs documentation" — broader than the front
page. At least one thing surfaced while building that belongs to it rather than
here: the autoapi-generated API reference emits a substantial number of
docstring formatting errors (unexpected section titles, undefined substitutions
such as `|r|` being read as a substitution reference, definition lists ending
without a blank line). Those are real defects in the rendered API pages, but
they are fixes to docstrings across the source tree, not to the landing page,
and they should not hold up the front door.

### DIFF
--- a/docs/dev_guide.rst
+++ b/docs/dev_guide.rst
@@ -1,0 +1,65 @@
+Developer guide
+===============
+
+This page is for people working *on* Layup. If you only want to use it, start
+at the :doc:`home page <index>` instead.
+
+Setting up an environment
+-------------------------
+
+Before installing any dependencies or writing code, create a virtual
+environment. LINCC-Frameworks engineers primarily use ``conda``:
+
+.. code-block:: console
+
+   >> conda create -n <env_name> python=3.11
+   >> conda activate <env_name>
+
+Alternatively, use Python's ``venv`` module:
+
+.. code-block:: console
+
+   >> python -m venv venv
+   >> source venv/bin/activate
+
+Once you have an environment, install the project for local development:
+
+.. code-block:: console
+
+   >> pip install -e .'[dev]'
+   >> pre-commit install
+   >> conda install pandoc
+
+Notes:
+
+1. The single quotes around ``'[dev]'`` may not be required for your operating
+   system.
+2. ``pre-commit install`` initializes pre-commit for this local repository, so
+   that a set of checks runs before each commit completes. For more
+   information, see the Python Project Template documentation on
+   `pre-commit <https://lincc-ppt.readthedocs.io/en/latest/practices/precommit.html>`_.
+3. Installing ``pandoc`` lets you verify that automatic rendering of Jupyter
+   notebooks into documentation works as expected. See the Python Project
+   Template documentation on
+   `Sphinx and Python Notebooks <https://lincc-ppt.readthedocs.io/en/latest/practices/sphinx.html#python-notebooks>`_.
+
+Running the tests
+-----------------
+
+.. code-block:: console
+
+   >> pytest
+
+Building the documentation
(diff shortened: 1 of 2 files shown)


mschwamb:
It's fine. Reads as though an AI wrote it.

matthewholman:
Yes, that's true.  I swapped in the JOSS summary paragraph.  If you have time to edit, that would be great!

mschwamb (approved):

PR #480, opened 2026-08-24
Add .zenodo.json so the archive credits the paper's authors

matthewholman:
Adds `.zenodo.json` so that when a release is eventually archived, the Zenodo record
credits the paper's authors.

## Why this is needed

Left to itself, Zenodo builds the archive record's author list from the repository's
**contributor statistics** — whoever has commits, in commit order. For Layup that is not
the eighteen-author list on the JOSS paper, and **the archive's author list is what people
cite**.

Fixing it after the fact means hand-editing the Zenodo record, which is easy to forget and
easy to get wrong under deadline. Fixing it here means the record is right the first time.

## What is in the file

All eighteen authors in the paper's order, with ORCIDs and affiliations, **generated from
`paper/paper.md`'s own front matter rather than transcribed** — the order was then checked
back against the paper programmatically. Adam Wilson has no ORCID in the paper, so the
field is omitted for him rather than guessed.

Two other fields are set deliberately:

- **`license`.** Zenodo defaults to `cc-by` for anything that is not a dataset, which
  would misdescribe an MIT-licensed package. The value `"mit"` is the license ID from
  Zenodo's own vocabulary API (`zenodo.org/api/vocabularies/licenses/mit`) — checked, not
  guessed. It is lowercase, and it is not an SPDX identifier, so `"MIT"` would be wrong.
- **`upload_type: software`**, so the record is not filed as a publication.

## What this does NOT do

Nothing takes effect until **both** of these happen:

1. the Zenodo–GitHub integration is switched on for `Smithsonian/layup`, and
2. a release is published *after* that.

Only releases created after the toggle is enabled are archived at all. Worth knowing when
that gets scheduled: because `Smithsonian/layup` is organization-owned, a Smithsonian org
owner may need to approve OAuth access for the Zenodo application first, which is the one
part of the sequence that depends on somebody outside this repository.

## Not on the JOSS submission critical path

Flagging this because it is easy to assume otherwise: **the DOI is not required to submit
to JOSS.** Per the JOSS documentation, the tagged release and archive come *after* review
— "upon successful completion of the review, authors will make a tagged release of the
software, and deposit a copy of the repository with a data-archiving service such as
Zenodo or figshare, get a DOI for the archive, and update the review issue thread with the
version number and DOI."

Submission itself needs the open repository with an OSI-approved license, `paper.md`, and
compliance with the review criteria.

## Testing

`.zenodo.json` parses as valid JSON; the creator list was verified to match `paper.md`'s
author order exactly, including the non-ASCII name and the three-part name. No code
changes, and the file is at the repository root so it does not enter the wheel.


mschwamb (changes requested):
Some affiliations don't match the JOSS paper.

matthewholman:
Holding this until #488 lands, and then regenerating.

You are right that the affiliations do not match, and the reason is worth recording so it does not recur: `.zenodo.json` was **generated from `paper/paper.md`** rather than typed, but it was generated on 08-25 — before #488 renumbered the affiliations into first-reference order and corrected several entries. So it is a faithful copy of a file that has since moved.

Regenerating now would only re-introduce the same drift when #488 merges, so the sensible order is #488 first, then regenerate from the merged `paper.md` and re-check that every author resolves to the same institutions on both sides.

One thing this suggests for later: the two files will drift again the next time the author block changes. A test that parses both and compares author-to-institution mappings would catch it automatically — it is the same check I ran manually on #488 that found the Moeyens index. Worth a small follow-up issue if you think it is warranted.

matthewholman:
@mschwamb, could you check now?

.zenodo.json
```diff
@@ -0,0 +1,109 @@
...
+      "orcid": "0000-0001-7335-1715"
+    },
+    {
+      "name": "Ruch, Thomas R.",
+      "affiliation": "University of Michigan, Ann Arbor, MI 48109, USA",
+      "orcid": "0000-0003-0403-0891"
+    },
+    {
+      "name": "Holt, Carrie E.",
+      "affiliation": "Las Cumbres Observatory, 6740 Cortona Drive, Suite 102, Goleta, CA 93117, USA",
```
mschwamb: Carrie needs to have LSST-DA Catalyst Postdoctoral Fellow listed somewhere in the affiliation I believe as per their rules.

matthewholman:
Can you check if this seems okay, @mschwamb ?

matthewholman:
All set, @mschwamb ?

mschwamb (approved):
Looks good to me.

Issue #481, opened 2026-08-24
The `log` verb ships as a user-facing command, and every command litters the working directory


matthewholman:
Found during a clean-room fresh-user run-through on `main` @ `719761c`, 2026-08-24 — fresh clone, fresh venv, following only the published instructions.

## What a user sees

`log` is advertised in `layup --help` alongside `orbitfit`, `convert` and `predict`. Running it gives:

```
$ layup log
2026-08-24 19:41:12,450 - layup.log_cmdline - INFO - About to call `log_cli`.
2026-08-24 19:41:12,450 - layup.log - INFO - In `log_cli` function.
2026-08-24 19:41:12,450 - layup.log - INFO - In `log` function.
2026-08-24 19:41:15,129 - layup.log - INFO - In `log_by_id` function.
2026-08-24 19:41:16,920 - layup.log - ERROR - I think you're great.
2026-08-24 19:41:16,920 - layup.log_cmdline - INFO - All done.
```

It generates a random array, pushes it through the parallel machinery, and prints placeholder messages — including **`I think you're great.` at ERROR level**, which reads as a fault to anyone who runs it. It takes about four seconds and produces no output file.

## This is not a bug in `log`, it is a verb that was never meant to ship

The module says so itself, in its first line (`src/layup/log.py`):

```python
"""This module is meant to be an example for how to logging works in Layup.
It has a corresponding layup_cmdline verb, `log` as well."""
```

and again inside it:

```python
# These imports are needed only for the demo
```

So this is developer scaffolding — a worked example of how to log from a module and fan work across workers — exposed on the user-facing command line.

## Why it is worth fixing before the JOSS submission

"Have the functional claims of the software been confirmed?" is a JOSS review checklist item, and the reviewer's natural way to check is to run each advertised verb. `log` is the one that will look broken: it announces an ERROR, does nothing a user asked for, and produces nothing. It costs nothing to explain in review, but it is a poor first impression from a package whose other nine verbs all do real work.

## Options, roughly in order of how much they change

1. **Remove the CLI verb, keep the module.** `src/layup/log.py` stays as the documented example for contributors; it just stops being reachable from `layup <verb>`. Smallest change, and it keeps the teaching value the module was written for.
2. **Keep it but hide it** from the verb list, so it is available for debugging without being advertised.
3. **Leave it and change the messages**, so at least nothing is logged at ERROR. Cheapest, but it still leaves a verb that does nothing for a user.

[...]

PR #483, opened 2026-08-24
Only create the .err log file if something is logged to it (#481)

matthewholman:
Addresses the **second** half of #481 — the empty `.err` files. The first half, whether the `log` verb should remain on the user-facing CLI, is a decision about the CLI surface and is with the team, so nothing here touches it.

## The problem

Following the documented quickstart left an empty `layup-<timestamp>.err` beside every command:

```
demo_orbit_kep.csv                 layup-2026-08-24-19-40-18.err
demo_orbitfit_output.csv           layup-2026-08-24-19-40-18.log
holman_data_working.csv            layup-2026-08-24-19-40-33.err
layup-2026-08-24-19-39-18.err      layup-2026-08-24-19-40-33.log
layup-2026-08-24-19-39-18.log      layup-2026-08-24-19-40-38.err
my_predictions.csv                 layup-2026-08-24-19-40-38.log
```

Eight log files against four real outputs, and **four of the eight `.err` files were empty**. A user who did everything correctly gets an error log for every step, which invites the thought that something went wrong.

## The fix

`logging.FileHandler` opens its file eagerly, so the `.err` was created whether or not anything was ever written to it. `delay=True` defers the open until a record is actually emitted — which is precisely the intended semantics: the file appears if and only if something was logged at ERROR or above.

One line, plus a comment and a docstring note.

The `.log` handler is deliberately left eager: it always receives INFO records, so delaying it would change nothing.

## Tests

Adds `tests/layup/test_layup_logging.py` — the first tests to cover the logger at all:

- a clean run (INFO + WARNING) leaves **no** `.err`;
- an ERROR still produces the file, and it carries only ERROR and above.

**Both were confirmed to fail with `delay=True` removed**, rather than assumed to be testing something. That check felt worth doing explicitly given #467 recently had to arm four assertions that were passing without asserting anything.

The tests share a fixture that strips handlers from the top-level `layup` logger. `LayupLogger` configures a shared logger, so without it a second instantiation in the same process stacks handlers on the first and the tests observe each other's output.

## How this was found

A clean-room fresh-user run-through on `main` @ `719761c` — fresh clone, fresh venv, following only the published instructions. Worth recording that the rest of that run passed: clone, `pip install -e .`, `bootstrap`, `demo prepare`/`howto`, `orbitfit`, `convert` and `predict` all succeeded, and the fitted elements for (3666) Holman match JPL SBDB to 8e-7 in *a*, 1e-4 in *e* and 9e-6 in *i*.

## Not addressed here

There is still no way to redirect the log directory — the files land in the current working directory unconditionally. That probably belongs with #448's cache-location setting rather than getting its own mechanism, so it is deliberately left out.

### DIFF
--- a/src/layup/utilities/layup_logging.py
+++ b/src/layup/utilities/layup_logging.py
@@ -11,6 +11,9 @@ class LayupLogger:
     3) layup-<datetime>.err depending on the log level. See the `_prepare_logger`
     method for details about which levels are sent to which handlers.
 
+    The .err file is created lazily: it appears only if something is actually
+    logged at ERROR or above, so a clean run does not leave an empty one.
+
     LayupLogger is intended to be used in one of two ways in general. Either
     instantiated within the `execute()` function in one of the layup_cmdline verbs
     or as a context manager when calling the API directly.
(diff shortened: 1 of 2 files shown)


Little-Ryugu (approved):
LGTM!

PR #487, opened 2026-08-25
removed "log" cmdline verb

### LINKED ISSUE #481: The `log` verb ships as a user-facing command, and every command litters the working directory

Little-Ryugu:
Fixes #481  .

Removed the example verb log from the toml, and its corresponding files


matthewholman:
Thanks for picking this up so quickly.

Removing `src/layup/log.py` along with the CLI wiring goes a step further than the options I listed in the issue — I had suggested keeping the module as the contributor example it was written to be. Removing both is a reasonable call and arguably the cleaner one, since a documentation-only module that nothing imports tends to rot. Flagging it only so it reads as a decision rather than something that slipped through: the file's docstring said it existed to show how logging works from a module, so if that guidance is still wanted it now lives nowhere.

**One thing that will be lost when this merges, and it is easy to keep.**

The closing keyword in the description means issue 481 shuts on merge, and that issue has two halves:

1. **The `log` verb.** This PR. Done.
2. **Every command writes its log files into the current working directory**, with no flag, config key or environment variable to put them anywhere else. Following the README quickstart end to end left eight log files against four real outputs.

The second half is untouched here. Half of it — the empty `.err` beside every successful run — was fixed separately in #483, but the "you cannot choose where they go" part is still open, and shutting issue 481 buries it.

Two options, either fine:

- soften the closing keyword to a plain reference ("Toward 481"), and shut the issue by hand once the log-directory question is settled; or
- let it close, and open a small follow-up for the log directory alone.

I would lean to the second, because there is now a natural home for it: **#484** makes the *data* cache location configurable through `LAYUP_CACHE_DIR`, and the log directory is the same class of question. Deciding both together is better than inventing a second mechanism later. (There is a note to that effect on #448 as well.)

**A small thing on the checklist.** "Have you written a unit test for any new functions?" is ticked, but the diff is four files of pure deletion with no test changes. That is fine for a removal — the real check is that CI passes and `layup --help` no longer advertises the verb — but the tick slightly overstates what was done, and someone reading the history later may go looking for a test that does not exist.

Worth confirming while you are in there, since `main.py` discovers verbs by scanning `PATH` for `layup-<verb>` executables: after this merges, an environment that still has an **old** `layup-log` script on `PATH` from a previous install will keep advertising the verb, and running it will fail rather than say "unknown verb". Not something this PR can fix, and not a reason to hold it — just the kind of thing that generates a confused bug report a month from now.

Little-Ryugu:
No worries.
I think it would be better to close issue #481 and open a new one solely about creating a cmdline arg that allows a user to choose the location of the generated log files.

mschwamb:
I think the code is the contributor example. I think the CLI is pretty easy to understand easily from the bootstrap and demo that I can we could just get rid of the log command. This CLI was adapted from the Sorcha one by a first year grad student (thanks @Little-Ryugu) so I think it's okay to completely remove.

mschwamb (approved):

mschwamb:
@Little-Ryugu can you put in a ticket for addressing the log file if it's not already sorted in another PR?

PR #488, opened 2026-08-25
WIP: proposed JOSS paper updates

mschwamb:
proposed JOSS updates

- Updates affiliations so they are listed in numeric order as other papers have standard that affiliations are numbered in order they are first referenced. 
- Updated acknowledgements to include NASA ADS replacement as I'm using it to check bibliography. MPC has a longer acknowledgement. Added in DP2 acknowledgement as we've used it for testing, and DP2 data rights statement this is required. Added in acknowledgments for J. Murtagh, and C. H that were in the overleaf and not transferred over. Updated SSSC Sprint acknowledgement, and removed  other  Sorcha funding acknowledgements that transferred over.


matthewholman:
Checked the affiliation renumbering mechanically: parsed the author and affiliation blocks before and after, resolved every author's indices to institution *names* on both sides, and compared. All 15 affiliations survive, all 18 authors survive in order, and **15 of 18 map to exactly the same institutions**. Three differ.

## One is an error

**Joachim Moeyens** — indices `"8,3"` became `"8,13"`.

| | resolves to |
|---|---|
| before | Asteroid Institute **+ DiRAC/UW** |
| after | **Department of Physics, University of Toronto** + Asteroid Institute |

The Asteroid Institute correctly moved from 8 to 13, but the old `8` was left in place and now resolves to Toronto Physics. So he gains a Toronto affiliation and loses DiRAC.

Confirmed with M.J.H.: he is at the Asteroid Institute and DiRAC/UW. The AJ paper agrees — `\astinst` and `\uw`, no Toronto.

**Suggested:** `affiliation: "8,13"` → `affiliation: "13,3"`.

## One is a question, not an error

**Wilson Beebe** — `"3"` became `"3,9"`, gaining LINCC alongside DiRAC.

That is a change of *fact* rather than of numbering, so it should not ride along with a renumbering pass unnoticed. It looks deliberate — Oldag, West and Chandler all carry `3,9` — but the **AJ paper has him at DiRAC only**, so the two papers now disagree. Whichever is right, they should match.

## One I flagged and withdrew

**Joseph Murtagh** — `"2"` became `"3"`, QUB to DiRAC/UW. I initially read this as the same stale-index slip. It is correct: he has graduated and moved to UW. The AJ paper already reflects it, including a DiRAC Postdoctoral Fellow altaffiliation.

## Two smaller things

**The grant number.** This PR touches the acknowledgements, which is where a discrepancy already sits: this paper says A. Wilson's studentship is **UKRI177**, and the AJ paper's Overleaf branch said **UKRI1776**. They cannot both be right, and since the AJ text was carried over from here, if UKRI1776 is correct then this paper is wrong too. There is a note about it in the AJ paper's `sections/16_acknowledgments.tex`. Worth settling while this file is open.

**The DP2 acknowledgement and data-rights statement.** Checked the current position: [the DP2 documentation](https://dp2.lsst.io/index.html) states "Only Rubin data rights holders may have an account in the Rubin Science Platform (RSP) and access to Data Preview 2", and [the community thread on public release](https://community.lsst.org/t/public-data-release-of-dp2-data-idacs-or-else/12263) confirms the standard two-year proprietary period, with Rubin itself not planning to provide public access afterwards. So the data-rights statement is the right thing to include. Worth being explicit in the wording that DP2 was used for *testing*, since a reader who cannot obtain DP2 cannot reproduce anything that rests on it.

## Method note

The check is mechanical and worth repeating after any further edit to the author block — a renumbering that is right in the list can still leave a stale index on an individual author, which is exactly what happened to Moeyens. Comparing resolved institution *names* rather than index numbers is what catches it.

matthewholman:
Following up on the Beebe question above: **confirmed correct by M.J.H.** — DiRAC and LINCC is right, so `"3"` -> `"3,9"` should stand. That was the one of the three I could not settle from the sources.

This paper was the one that was right and the AJ paper was out of step, so I have corrected the AJ paper rather than this one. Oldag, West, Beebe and Chandler now all carry DiRAC and LINCC in both.

That leaves **only Moeyens** to fix here: `affiliation: "8,13"` -> `"13,3"`, since the stale `8` now resolves to Toronto Physics rather than the Asteroid Institute. M.J.H. confirms he is at the Asteroid Institute and DiRAC/UW.

Everything else in the renumbering checks out.

mschwamb:
Thanks. I'll make those changes. I'm still in the process of making suggested edits and double checking changes, I'll let you know when it's ready for review.

mschwamb:
Ready for review

mschwamb:
One more change. I missed the other part to join in because of the claude comments. That's fixed now.

matthewholman:
FYI replaced UKRI177 with UKRI176 in the AJ paper.

matthewholman (approved):
LGTM!

PR #494, opened 2026-08-27
Update paper.md

mschwamb:
Minor fixes. 
fix pandey reference so it renders and add a space between 'REBOUND' and reference parenthesis


matthewholman (approved):
Looks okay.

PR #496, opened 2026-08-27
Stop re-downloading the debiasing tarball every bootstrap (#482)

### LINKED ISSUE #482: bootstrap leftover survives in caches created before #472
Follow-on from #472, found during a clean-room fresh-user run-through on `main` @ `719761c` (which **includes** the merged #472), 2026-08-24.

## The gap

#472 deletes `debias_hires2018.tgz` **after the untar**. That is the right fix, and it works — for anyone bootstrapping from now on.

It does nothing for a cache that was already populated. `layup bootstrap` on an existing cache takes the "everything is already here" path:

```
$ layup bootstrap
INFO - Checking cache for existing files.
All expected files were found in the local cache: /Users/…/Library/Caches/layup/
```

It never re-extracts, so it never reaches the new cleanup, and the archive stays:

```
$ ls -la ~/Library/Caches/layup/debias_hires2018.tgz
-rw-r--r--  1 user  staff  156 MB  debias_hires2018.tgz
```

So **every existing user keeps the 156 MB leftover indefinitely**, on a package that already asks for about 3.2 GB. The people most affected are the ones who have been using `layup` longest.

## Why this is worth a few lines rather than a shrug

The obvious counter-argument is that anyone can delete it by hand. But nobody knows it is there — it lives in a platform-specific cache directory (`pooch.os_cache("layup")`), it is not mentioned in the documentation, and the only reason it is known at all is that a run-through went looking. A user who never re-bootstraps will never recover the space.

## Suggested fix

Have `bootstrap` remove a stale `debias_hires2018.tgz` **on the already-cached path too**, not only after an untar — i.e. if the extracted contents are present and the archive is also present, delete the archive. That is a small addition to the branch #472 already touched, and it makes the fix retroactive without a re-download.

Worth logging when it happens, so the reclaimed space is visible rather than mysterious.

## Bearing on #436

#436's second item ("bootstrap leftover") is satisfied by #472 **for new installs only**. If #436 is closed on the strength of #472 alone, this case goes with it — which is the reason for filing separately rather than reopening that thread.

(#436's first item, the stale PyPI `0.0.1` stub, remains downstream of the first real release and is unaffected by any of this.)

matthewholman:
`data_file_list` names `debias_hires2018.tgz`, and `_check_for_existing_files`
walks that list. #472 deletes the tarball as soon as it is unpacked, so the check
was looking for exactly the file #472 removes. Every bootstrap since reported it
missing, re-downloaded ~156 MB from JPL, unpacked it, deleted it, and did the
same again next time.

Extracted archives are now skipped in the check. They stay in the pooch registry
so they can still be fetched; they are just no longer expected to survive. The
extracted payload, `bias.dat`, is still checked, so an incomplete cache is still
caught.

Also sweeps a leftover archive when the cache is otherwise complete, which covers
caches populated before #472 — the case #482 was filed for.

Closes #482.


Little-Ryugu (approved):
LGTM! 
Just highlighting as a comment. 
Do we need the function _remove_extracted_archives, as the code now already removes extracted archives during bootstrap #472? The fix is for previous versions of Layup, but does that matter for the pre-v1.0 versions?

matthewholman:
Two separate halves. The check-skip is the live bug: `data_file_list` names the tarball #472 deletes, so every bootstrap on `main` re-downloads 156 MB. The sweep is the lesser half -- it only helps caches already on disk from before #472, and can go once v1.0 ships.

Little-Ryugu:
No worries, just wanted to check. pr is all good to merge.

Issue #493, opened 2026-08-26
orbitfit should report what was determined, not just whether it converged


matthewholman:
`flag == 0` means the optimizer converged. It does not mean an orbit was determined, and on short arcs those differ badly: 37% of converged short-arc fits come back physically impossible, 56% at three days (#485).

The information needed to tell them apart is already computed and returned. On the failing cases the one-sigma velocity exceeds the velocity itself. Every downstream consumer has ended up writing its own screen — the full-catalogue run had to drop 37,735 rows on a covariance test after they passed `flag == 0`.

The deeper issue is that reduced chi-square is the quality gate, and it is *best* exactly when the arc is least informative, so it cannot detect this class of failure.

Suggested: `orbitfit` should surface a verdict about determinacy alongside the state — at minimum a flag value for "converged but under-determined", derived from the covariance it already has. Then `flag == 0` can mean what callers assume it means.

Related: #485 reports this as a bug; this is the contract question underneath it.

mschwamb:
I don't think this is an enhancement. I think this is something we need in some form now. I would assume that flag=0 means that the orbit is somewhat physical. So if we know the orbit is not physical in any way possible than I think that should be flagged as 1 or 2 to identify it

matthewholman:
Here are the Layup orbitfit `flag` values, as currently set in the code (2026-08-27).
Not documented anywhere in docs/.

We could add one that means "converged but orbit is impossible/not feasible.

-1  Not attempted.  Placeholder written for rows that were never fit.
    (orbitfit.py:681)

 0  Converged and accepted.
    (orbit_fit.cpp:1163, bk_fit.cpp:190, bk_iod.cpp:293)

 1  Did not converge.  The initial value, left in place when nothing
    better is reached.
    (orbit_fit.cpp:1046, bk_fit.cpp:30, bk_iod.cpp:93)

 2  Converged, but chi-square per degree of freedom is above threshold.
    A quality gate applied to a fit that did converge.  See issue #367.
    (orbit_fit.cpp:1181, bk_fit.cpp:216)

 3  IOD candidates existed but none converged on the primary interval.
    The least-bad attempt, by chi-square, is returned so the caller has
    something to inspect.
    (orbitfit.py:981)

 4  The primary interval converged, but the incremental build-up to the
    full observation set failed partway.  Set at the segment where it
    broke.
    (orbitfit.py:1000)

 5  No solution at all.  Either the IOD returned no candidates and we are
[...]

Issue #485, opened 2026-08-25
orbitfit reports flag=0 for a 10-minute arc while returning an unbound orbit at 970x circular velocity


matthewholman:
`orbitfit` reports `flag = 0` — success — for a very short arc while returning an orbit that is not physically possible. Anything downstream that filters on `flag == 0`, which is the documented way to select good fits, takes it.

## Reproducing it, with data that ships with layup

No special data needed. Take the bundled demo dataset and keep one 10-minute window from a single night:

```python
import csv, datetime
rows = [r for r in csv.DictReader(open("holman_data_working.csv")) if r["obsTime"]]
for r in rows:
    r["_t"] = datetime.datetime.fromisoformat(r["obsTime"].replace("Z", ""))
rows.sort(key=lambda r: r["_t"])
t0 = datetime.datetime.fromisoformat("2024-07-15T06:23:43.700")
win = [r for r in rows if 0 <= (r["_t"] - t0).total_seconds() <= 900]   # 11 obs, 10.2 min
hdr = [k for k in rows[0] if not k.startswith("_")]
with open("holman_shortarc.csv", "w", newline="") as f:
    w = csv.DictWriter(f, hdr); w.writeheader()
    for r in win: w.writerow({k: r[k] for k in hdr})
```

then

```
layup demo prepare orbitfit
layup orbitfit holman_shortarc.csv ADES_csv -o shortarc_out
```

## What comes back

For (3666) Holman — a main-belt asteroid with a well known orbit — from 11 real observations spanning **10.2 minutes**:

| | value |
|---|---|
| `flag` | **0 (converged)** |
| reduced chi-square | **0.009** |
| \|r\| | 3.29 au |
| **\|v\| / v_circular** | **971** |
| a | **−3.5e-06 au** |
| energy | **unbound** |

The fitted speed is roughly **970 times the circular velocity** at that distance. The orbit is hyperbolic with a semimajor axis of a few times 10⁻⁶ au. The object is an ordinary MBA.

**Both engines do this.** `--engine bk_native` returns the same `flag = 0` and the same 971×, so the bound-orbit energy prior on `gdot` does not prevent it.

## Why the existing quality gate cannot catch this

[...]

PR #497, opened 2026-08-27
Report a converged but physically impossible orbit as flag 9 (#493)

matthewholman:
Adds flag 9: converged, but the orbit is not physically possible.

The criterion is hyperbolic excess speed, not boundedness. Layup is expected to
fit interstellar objects, and those are unbound and fast -- 3I/ATLAS arrives at
about 59 km/s -- so an unbound orbit is never on its own grounds for rejection.
Default threshold 200 km/s (`orbitfit.MAX_EXCESS_SPEED_KM_S`, 0 disables it).

Measured on 196 short arcs cut from MPC histories, with a long-arc fit of the
same object as truth: the gate rejected **no correct fit**, and every fit it
rejected was wrong. It is a floor, not a filter -- it catches the unambiguous
cases and cannot reach fits that are plausible and still wrong.

Applied to the result the caller receives, not inside the C++ fit: the driver
runs one fit per IOD candidate and reads a non-zero flag as "candidate failed",
so a per-candidate verdict is retried away and reported as flag 3.

Like flag 2, a gated fit reports state and covariance as NaN.

Two things found on the way, both filed separately rather than fixed here: 59%
of flag=0 short-arc fits have the wrong semimajor axis (#485 says 37%, measured
against a physicality proxy rather than truth), and `orbitfit.py` overwrites the
C++ flag on two paths.

Toward #493.

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -34,6 +34,56 @@ except ImportError:  # extension not rebuilt yet
 # _MU_SUN (= heliocentric GM = k^2) is used by the BK-native fit for the
 # bound-orbit energy prior on gdot; SPEED_OF_LIGHT (au/day) by the radar ingest.
 from layup.constants import MU_SUN as _MU_SUN, SPEED_OF_LIGHT
+
+# Flag for a fit that converged but describes an orbit no real object could
+# occupy (issue #493). It sits with flag 2 (chi-square too large) and flag 6
+# (degenerate covariance) as a fit that converged and was then rejected by a
+# gate, rather than with the flags for fits that never converged.
+FLAG_IMPLAUSIBLE_ORBIT = 9
+
+# Hyperbolic excess speed, in km/s, above which a converged fit is reported as
+# physically impossible.
+#
+# A short arc can converge with an excellent reduced chi-square onto a state no
+# real object could occupy, because the arc does not constrain the velocity. The
+# chi-square gate cannot catch this: it is anti-correlated with the failure,
+# since the less the object moves across the arc the better the fit (issue #485).
+#
+# Excess speed is the one statement available without assuming the object is
+# bound. Layup is expected to fit genuine interstellar objects, and those are
+# unbound and fast -- 3I/ATLAS arrives at about 59 km/s -- so boundedness itself
+# is never grounds for rejection and only a speed far above any plausible arrival
+# speed is evidence of a bad fit rather than an unusual object. The default is
+# about 3.4x the fastest interstellar object yet observed. Measured against
+# long-arc truth orbits it rejected no correct fit, and every fit it rejected was
+# wrong. Set to 0 to disable the check.
+MAX_EXCESS_SPEED_KM_S = 200.0
+
+_KM_S_IN_AU_PER_DAY = 86400.0 / 149597870.7
+
+
+def _implausible_excess_speed(state) -> bool:
+    """True when a converged state's hyperbolic excess speed is impossibly large.
+
+    ``state`` is the barycentric equatorial state (au, au/day). Barycentric rather
+    than heliocentric, and the Sun's mass rather than the whole solar system's,
+    are both far below the threshold that matters here.
+    """
+    if not MAX_EXCESS_SPEED_KM_S > 0:
+        return False
+    pos = np.asarray(state[:3], dtype=float)
+    vel = np.asarray(state[3:6], dtype=float)
+    if not (np.all(np.isfinite(pos)) and np.all(np.isfinite(vel))):
+        return False  # a NaN state is already reported by the convergence flag
+    r = float(np.linalg.norm(pos))
+    if not r > 0:
+        return False
+    energy = 0.5 * float(np.dot(vel, vel)) - _MU_SUN / r
+    if not energy > 0:
+        return False  # bound, or exactly parabolic
+    return np.sqrt(2.0 * energy) > MAX_EXCESS_SPEED_KM_S * _KM_S_IN_AU_PER_DAY
+
+
 from layup.convert import convert
 from layup.iod import filter_candidates_by_residual, get_iod, iod_methods
 
@@ -1331,6 +1381,14 @@ def _orbitfit(
                 res = res_ng
             else:
                 logger.debug("Non-grav refinement did not converge; reporting non-grav params as NaN.")
+
+        # Physical-plausibility gate (issue #493). Applied here, to the result the
+        # caller is about to receive, rather than inside the C++ fit: the driver
+        # above runs one fit per IOD candidate and reads a non-zero flag as "this
+        # candidate failed", so a verdict set per-candidate is retried away and
+        # reported as flag 3 or 4.
+        if res.flag == 0 and _implausible_excess_speed(res.state):
+            res.flag = FLAG_IMPLAUSIBLE_ORBIT
 
         # Populate our output structured array with the orbit fit results
         success = res.flag == 0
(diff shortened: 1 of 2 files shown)


src/layup/orbitfit.py
```diff
@@ -1332,6 +1382,14 @@ def _orbitfit(
             else:
                 logger.debug("Non-grav refinement did not converge; reporting non-grav params as NaN.")
 
+        # Physical-plausibility gate (issue #493). Applied here, to the result the
```
mschwamb: Let's not put issues into the doc strings This does not help maintainable.

tests/layup/test_implausible_orbit_flag.py
```diff
@@ -0,0 +1,139 @@
+"""flag = 9: converged, but the orbit is not physically possible (issue #493).
+
+A short arc can converge with an excellent reduced chi-square onto a state no
+real object could occupy, because the arc does not constrain the velocity. The
+chi-square gate cannot catch this -- it is anti-correlated with the failure,
+since the less the object moves the better the fit (issue #485).
```
mschwamb: no issues in doc strings.

src/layup/orbitfit.py
```diff
@@ -34,6 +34,56 @@
...
+# since the less the object moves across the arc the better the fit (issue #485).
+#
+# Excess speed is the one statement available without assuming the object is
+# bound. Layup is expected to fit genuine interstellar objects, and those are
+# unbound and fast -- 3I/ATLAS arrives at about 59 km/s -- so boundedness itself
+# is never grounds for rejection and only a speed far above any plausible arrival
+# speed is evidence of a bad fit rather than an unusual object. The default is
+# about 3.4x the fastest interstellar object yet observed. Measured against
+# long-arc truth orbits it rejected no correct fit, and every fit it rejected was
+# wrong. Set to 0 to disable the check.
```
mschwamb: Do we need all this text as a doc string/comment that duplicates what is in function. I think a shorter summary here would be useful.

tests/layup/test_implausible_orbit_flag.py
```diff
@@ -0,0 +1,139 @@
...
+evidence of a bad fit rather than an unusual object.
+"""
+
+import numpy as np
+import pytest
+
+import layup.orbitfit as orbitfit_module
+from layup.orbitfit import FLAG_IMPLAUSIBLE_ORBIT
+
+GM_SUN = 2.9591220828559115e-4  # au^3/day^2
```
mschwamb: why are we not using constants.py to store these values?

tests/layup/test_implausible_orbit_flag.py
```diff
@@ -0,0 +1,139 @@
...
+KM_S_IN_AU_DAY = 86400.0 / 149597870.7
```
mschwamb: why are we not using constants.py to store these values?

tests/layup/test_implausible_orbit_flag.py
```diff
@@ -0,0 +1,139 @@
...
+    assert get_max_v_inf() == 0.0
+
+
+# ---------------------------------------------------------------------------
+# End-to-end, through a real fit. 3I/ATLAS is the natural fixture: it is a
+# genuine interstellar object, so it is unbound and fast enough to sit on the
+# interesting side of the gate, and layup already ships its discovery-arc
+# astrometry. Moving the threshold across its excess speed must move the flag,
+# which tests the gate itself rather than the accessor.
+# ---------------------------------------------------------------------------
```
mschwamb: A natural fixture? It doesn't matter whether layup ships the observations in a demo or something else. This is not relevant to a code comment about what the code is doing. Tests should be appropriate whether or not the data is available already or not. I'm not sure which is the fastest. I'd either remove or revise this entire comment.

src/layup/orbitfit.py
```diff
@@ -34,6 +34,56 @@
...
+MAX_EXCESS_SPEED_KM_S = 200.0
+
+_KM_S_IN_AU_PER_DAY = 86400.0 / 149597870.7
```
mschwamb: why are we not using constants.py to store these constants in?

tests/layup/test_implausible_orbit_flag.py
```diff
@@ -0,0 +1,139 @@
...
+@pytest.fixture(autouse=True)
+def _restore_threshold():
+    """The threshold is module-level state; put it back after each test."""
+    original = orbitfit_module.MAX_EXCESS_SPEED_KM_S
+    yield
+    orbitfit_module.MAX_EXCESS_SPEED_KM_S = original
+
+
+def set_max_v_inf(km_s):
+    orbitfit_module.MAX_EXCESS_SPEED_KM_S = km_s
```
mschwamb: why are we changing the values of constants? I get wary when tests does this. They should be able to set with what's already in the constant set in the code. A unit test can simply be written without having to edit all of this?

src/layup/orbitfit.py
```diff
@@ -34,6 +34,56 @@
...
+
+
+def _implausible_excess_speed(state) -> bool:
+    """True when a converged state's hyperbolic excess speed is impossibly large.
+
+    ``state`` is the barycentric equatorial state (au, au/day). Barycentric rather
+    than heliocentric, and the Sun's mass rather than the whole solar system's,
+    are both far below the threshold that matters here.
+    """
+    if not MAX_EXCESS_SPEED_KM_S > 0:
```
mschwamb: I think this is really bad to have 0 be turning off this check - just setting the implausible speed to very high would do the same thing keep the same logic or set some kind of flag to turn this check off

src/layup/orbitfit.py
```diff
@@ -34,6 +34,56 @@
...
+_KM_S_IN_AU_PER_DAY = 86400.0 / 149597870.7
+
+
+def _implausible_excess_speed(state) -> bool:
+    """True when a converged state's hyperbolic excess speed is impossibly large.
+
+    ``state`` is the barycentric equatorial state (au, au/day). Barycentric rather
+    than heliocentric, and the Sun's mass rather than the whole solar system's,
+    are both far below the threshold that matters here.
+    """
```
mschwamb, line 69: doc stings re not in the same format as elsewhere in layup

tests/layup/test_implausible_orbit_flag.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def get_max_v_inf():
+    return orbitfit_module.MAX_EXCESS_SPEED_KM_S
+
+
+def test_default_threshold_clears_the_fastest_known_interstellar_object():
+    """The default must not reject a real interstellar object.
+
+    Layup's stated advantage over OrbFit and OpenOrb is that it handles the
```
mschwamb: Can we  not comment about other people's packages in comments?

src/layup/orbitfit.py
```diff
@@ -34,6 +34,56 @@
 # _MU_SUN (= heliocentric GM = k^2) is used by the BK-native fit for the
 # bound-orbit energy prior on gdot; SPEED_OF_LIGHT (au/day) by the radar ingest.
 from layup.constants import MU_SUN as _MU_SUN, SPEED_OF_LIGHT
+
+# Flag for a fit that converged but describes an orbit no real object could
+# occupy (issue #493). It sits with flag 2 (chi-square too large) and flag 6
+# (degenerate covariance) as a fit that converged and was then rejected by a
+# gate, rather than with the flags for fits that never converged.
+FLAG_IMPLAUSIBLE_ORBIT = 9
```
mschwamb: where are the other flags defined. I think all flags should be easily findable with their numbers. Maybe we need another file that stores those constants?

mschwamb (changes requested):
Feedback for consideration

matthewholman:
I believe your concerns have been addressed, @mschwamb

mschwamb (approved):

PR #501, opened 2026-08-28
Use updated LINCC Frameworks acknowledgement

mschwamb:
Use updated LINCC Frameworks acknowledgement. Confirmed by Rachel Mandelbaum one of the co-leads of LINCC Frameworks


awilson110 (approved):
Looks good to me

PR #503, opened 2026-08-28
Logging config and cmd-line additions.

### LINKED ISSUE #42: Add back in logging to config file reader/parse class
Add back in logging to config file reader/data class once logging (but have it direct to stderr by default) has been revamped for the file readers data class by @drewoldag

### LINKED ISSUE #490: Creating a log file in working directory when running layup verbs.
When any layup verb is run, a log file is created in the working directory where the command was run.

I'm happy to do this now. I'll add a new command-line arg for the log path for every layup verb that creates a log file. 

 #481 mentioned this briefly, and this issue was made so this part can be highlighted clearly.

Little-Ryugu:
Fixes #42 
Added back logging to layup_configs. if the default configs are used the log will print ("Using default config parameters.")
if the cmdline arg -c is used then a copy of that config file is printed into the logs (similar to sorcha)
```
# Save a raw copy of the configuration to the logs as a backup.
            with open(config_file_location, "r") as file:
                logger.debug(f"Copy of configuration file {config_file_location}:\n{file.read()}")
```
Fixes #490  .

Added the following command line args for verbs comet, convert, orbitfit, unpack and predict. 
allowing for a user to decide where layup outputs and logs will be generated.
```
optional.add_argument(
        "-t",
        "--stem",
        help="output file name stem.",
        dest="t",
        type=str,
        default="comet_output",
        required=False,
    )

    optional.add_argument(
        "-o",
        "--outfile",
        help="Path to store output and logs.",
        type=str,
        dest="o",
        default="./",
        required=False,
    )
```
Also made it so that the log filename contains the verb that was ran to make it easier to differenate different log files:
` log_file_base_name = f"layup-{verb}-{timestamp}"`
i.e log file names will now look like this:
layup-comet-2026-08-28-14-43-25.log

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ layup demo howto orbitfit
 asteroid (3666) Holman, in ADES CSV form — to the current directory, and `howto`
 prints the ready-to-run command. Fit it with:
 ```
-layup orbitfit holman_data_working.csv ADES_csv -o demo_orbitfit_output
+layup orbitfit holman_data_working.csv ADES_csv -t demo_orbitfit_output
 ```
 This writes the best-fit barycentric Cartesian orbit and its covariance to
 `demo_orbitfit_output.csv`. Supported input formats are `MPC80col`, `ADES_csv`, `ADES_psv`,
@@ -111,12 +111,12 @@ This writes the best-fit barycentric Cartesian orbit and its covariance to
 
 Convert the result to another orbit representation (Cometary, Keplerian, …):
 ```
-layup convert demo_orbitfit_output.csv KEP -o demo_orbit_kep
+layup convert demo_orbitfit_output.csv KEP -t demo_orbit_kep
 ```
 
 Predict future on-sky positions, with uncertainties, for an observatory:
 ```
-layup predict demo_orbitfit_output.csv --days 30 --station X05 -o my_predictions
+layup predict demo_orbitfit_output.csv --days 30 --station X05 -t my_predictions
 ```
 
 Every verb takes `--help` for its full set of options (engine choice, IOD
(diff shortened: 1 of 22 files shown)


matthewholman:
Two on the flag changes.

`predict -t` was `--timestep` and is now `--stem`. A stem is any string, so `-t 30m` stays valid input -- the timestep silently reverts to the 1h default. Dropping the short form entirely would make the old spelling error instead.

`--ts` also reads like an abbreviation of nothing. `--timestep` on its own seems better.

mschwamb:
I think for predict we can use -s --stem so that seems reasonable for that one. Different for different verbs but that's probably okay

mschwamb (changes requested):
See comment about predict

matthewholman:
`-s` is taken in `predict` too -- it's `--start-date`. Same failure mode: `-s 2025 MAY 18` would become a stem and the start date would silently fall back to the current UTC date.

`predict` already uses `-ar -c -ch -d -e -f -i -n -o -osd -pid -s -sg -st -t`. Simplest may be to give `--stem` no short form there.

mschwamb:
@Little-Ryugu 

Something like this should work without needing to define the shorten the command line argument. 

`optional.add_argument('--stem',   dest="stem", type=str, help=output file name stem',   default="comet_output",required=False) `

mschwamb:
I think Matt wants it to be stem across all of the verbs

Little-Ryugu:
I've changed it now so that the cmdline arg can be `-t` or `--stem` for the verbs (`comet`,` convert`,` orbitfit`, `unpack`), except for `predict` that can only use `--stem` as `-t` is the timestep in `predict`. 

Would it make sense to keep it consistent by removing `-t` and making `--stem` the only cmd-line arg for the output file name stem for all the verbs?

Little-Ryugu:
> I think Matt wants it to be stem across all of the verbs

Sorry missed that when I asked there. I'll change them now

mschwamb (approved):
Looks good to me and is working for me. But @matthewholman should review as well before merge.

mschwamb:
Good to merge

Issue #500, opened 2026-08-27
Three tests run whichever `layup` is first on PATH, not the one under test


matthewholman:
`test_predict.py:223`, `test_predict.py` (get_onsky_data) and
`test_comet.py:216` invoke the CLI as a subprocess:

```python
result = subprocess.run(["layup", "predict", str(input_file), ...])
assert result.returncode == 0
```

`"layup"` resolves against PATH, so the test exercises whatever installation
comes first -- not necessarily the working tree. On this machine that is a conda
env with a broken assist/rebound link, and all three fail with
`assert 1 == 0` and an unrelated `OSError: dlopen ... librebound...so`. Putting
the intended environment's `bin` first makes all three pass, with no code change.

The failure mode is the bad one: it looks like a code regression, and the real
message is buried in captured stderr. It can also pass for the wrong reason,
testing a stale install while the working tree is broken.

Suggest `[sys.executable, "-m", ...]`, or resolving the entry point relative to
`sys.executable`, so the subprocess inherits the interpreter under test.

Issue #499, opened 2026-08-27
Document that flag under-reports gate rejections; triage on the failed_* columns


matthewholman:
`orbitfit.py:981` and `:1000` assign the flag unconditionally:

```python
x = min(candidates, key=lambda c: c.csq)
x.flag = 3          # line 981
...
if x.flag != 0:
    x.flag = 4      # line 1000
```

Whatever the C++ returned is discarded. A candidate that converged and was then
rejected on chi-square per degree of freedom (flag 2) is reported as 3 or 4,
i.e. as a fit that never converged. The reason it was rejected does not survive.

So flag 2 is rarer in real output than the code suggests, and anything triaging
on it under-counts.

Same mechanism bites any new post-convergence verdict: it is why #497 applies
its check to the final result rather than per candidate.

matthewholman:
#513 changed what this issue is about.

The reason is no longer lost: a candidate that converged and was then rejected is reported in `failed_csq` / `failed_cov`. But `flag` is still overwritten at the lines above, deliberately — keeping it unchanged is what made #513 non-breaking, so `flag == 0` and the existing output format still mean what they did.

So the under-count is now intended behaviour rather than a defect, and what is left is saying so: `flag` marks where a fit stopped, the `failed_*` columns say what rejected it. That belongs with #495.

matthewholman:
Done in #495, merged. `docs/fit_flags.rst` documents the six outcome columns and carries the warning that `flag == 2` and `flag == 6` under-report rejections, with `failed_csq` and `failed_cov` as the columns to count on.

PR #504, opened 2026-08-28
test: invoke layup CLI subprocesses under the interpreter under test

Manny7717:
## Problem

`test_predict_output` (and the `-sg` / `-osd` variants) and `test_comet_output`
invoke the CLI as a subprocess with a bare `"layup"` argv entry:

```python
result = subprocess.run(["layup", "predict", str(input_file), ...])
```

`"layup"` is resolved against `PATH`, so these tests exercise whichever
installation of layup comes first on the machine — not necessarily the working
tree under test. On a machine where a different environment (e.g. a conda env
with a broken `assist`/`rebound` link) shadows the intended one, all three
tests fail with `assert 1 == 0` and an unrelated `OSError: dlopen ...
librebound...` buried in captured stderr. The failure looks like a code
regression when it is purely an environment-resolution problem — and the tests
can also pass for the wrong reason, exercising a stale install while the
working tree is broken.

## Fix

Replace the bare `"layup"` argv entries with the interpreter under test,
invoking the same entry-point modules directly:

```python
result = subprocess.run(
    [sys.executable, "-m", "layup_cmdline.predict", str(input_file), ...]
)
```

and similarly `sys.executable -m layup_cmdline.comet` in `test_comet.py`.

The console scripts `layup-predict = "layup_cmdline.predict:main"` and
`layup-comet = "layup_cmdline.comet:main"` wrap exactly those `main()`
functions, and both modules already carry `if __name__ == "__main__": main()`,
so `python -m` runs the identical entry point. The subprocess now inherits the
interpreter running the test suite (the environment where layup is installed
for development), no `PATH` lookup involved — the working tree under test is
always what gets exercised.

All four call sites are changed (the issue listed three; the `-sg` variant in
`test_predict.py` had the same problem).

## Tests

- `tests/layup/test_predict.py` — `test_predict_output`,
  `test_predict_output` with `-sg`, `test_get_onsky_data_output`
- `tests/layup/test_comet.py` — `test_comet_output`

These are the subprocess tests that previously depended on `PATH` resolution.
Verified locally with the repository built as `pip install -e ".[dev]"`
(editable install in a fresh venv) and `layup bootstrap` data present:

```
tests/layup/test_predict.py::test_predict_output PASSED
tests/layup/test_predict.py::test_get_onsky_data_output PASSED
tests/layup/test_comet.py::test_comet_output PASSED
3 passed in 191.98s
```

`black --check` clean on both files.

**Regression proof (PATH dependence):** with `PATH=/usr/bin:/bin` (no venv bin
on PATH), the old bare-`"layup"` invocation raises
`FileNotFoundError: 'layup'` — the subprocess-dependent tests would fail
outright (or, with a shadowing install, fail with the misleading
`librebound`/`libassist` `OSError` described in the issue). The new
`sys.executable -m layup_cmdline.predict --help` invocation under the same
restricted PATH exits 0 and prints the expected usage, showing the tests now
exercise the interpreter under test regardless of `PATH`.

## Limitations

- This changes only the test invocation, not the CLI itself. If layup is not
  installed in the environment running the tests, the subprocess fails to
  import `layup_cmdline.predict` — but that environment could not have run the
  tests at all before, since the in-process imports already require the
  package.
- The `layup` dispatcher (`layup_cmdline.main`, the `layup <verb>` wrapper)
  still resolves `layup-<verb>` scripts via `PATH` internally; these tests
  bypass the dispatcher by invoking the verb modules directly, which is also
  one less process layer.

### DIFF
--- a/tests/layup/test_comet.py
+++ b/tests/layup/test_comet.py
@@ -1,5 +1,6 @@
 import argparse
 import os
+import sys
 
 import pytest
 import rebound
@@ -214,7 +215,17 @@ def test_comet_output(tmpdir):
     # The demo comet fixture is keyed by ObjID; comet's -pid now defaults to
     # provID (CLI-consistency), so pass -pid ObjID explicitly.
     result = subprocess.run(
-        ["layup", "comet", str(input_file), "-f", "-o", str(temp_out_file), "-pid", "ObjID"]
+        [
+            sys.executable,
+            "-m",
+            "layup_cmdline.comet",
+            str(input_file),
+            "-f",
+            "-o",
+            str(temp_out_file),
+            "-pid",
+            "ObjID",
+        ]
     )
 
     assert result.returncode == 0
--- a/tests/layup/test_predict.py
+++ b/tests/layup/test_predict.py
@@ -1,5 +1,6 @@
 import os
 import subprocess
+import sys
 from pathlib import Path
 
 import numpy as np
@@ -221,7 +222,17 @@ def test_predict_output(tmpdir):
     temp_out_file = f"test_output_{input_file.stem}"
 
     result = subprocess.run(
-        ["layup", "predict", str(input_file), "-f", "-o", str(temp_out_file), "-s", start]
+        [
+            sys.executable,
+            "-m",
+            "layup_cmdline.predict",
+            str(input_file),
+            "-f",
+            "-o",
+            str(temp_out_file),
+            "-s",
+            start,
+        ]
     )
 
     assert result.returncode == 0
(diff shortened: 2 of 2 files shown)


matthewholman:
Thanks for these. We'll implement both from the issues rather than merge -- we keep changes to work authored within the project. #499 and #500 stay open.

Manny7717:
Thanks for the clear note, @matthewholman — understood that layup keeps feature work authored in-house. Closing this PR since you're implementing #500 internally. Appreciate the quick response!

PR #505, opened 2026-08-28
fix(orbitfit): preserve post-convergence gate verdicts in fit flags

### LINKED ISSUE #499: Document that flag under-reports gate rejections; triage on the failed_* columns

Manny7717:
Closes #499.

## Problem

`do_fit` overwrites the fitter's flag unconditionally at two points:

- `src/layup/orbitfit.py:981` — the "no root converged" path forced the least-bad candidate to `flag = 3`;
- `src/layup/orbitfit.py:1000` — the incremental build-up path forced the breaking segment to `flag = 4`.

A candidate that **converged and was then rejected by a quality gate** — `flag = 2` (chi-square per degree of freedom above threshold) or `flag = 6` (degenerate covariance), i.e. the "Converged, but rejected by a gate" group in the `docs/fit_flags.rst` taxonomy — was therefore reported as a fit that never converged, and the reason it was rejected did not survive. Anything triaging on `flag == 2` under-counts real cases.

## Fix

Only a plain non-convergence verdict (`flag = 1`, the fitter's initial value) is replaced by the driver's location markers (`3` = no IOD root converged on the primary interval, `4` = incremental build-up failed partway). Post-convergence gate verdicts (`2`, `6`) are preserved at both sites — the least-bad candidate and the breaking segment are exactly where that reason matters.

No downstream code depends on the exact `3`/`4` values (verified: they are terminal verdicts), so behavior for genuinely non-converged fits is unchanged.

## Tests

5 new tests in `tests/layup/test_iod_picker.py` (stub `_run_fit`/ephemeris, no kernels or C++ fitter needed):

- `test_no_root_converged_preserves_chi2_gate_verdict` — flag 2 survives the "no root converged" path (fails on old code)
- `test_no_root_converged_preserves_degenerate_verdict` — flag 6 survives (fails on old code)
- `test_no_root_converged_still_marks_plain_nonconvergence` — flag 1 still becomes 3
- `test_incremental_buildup_preserves_chi2_gate_verdict` — flag 2 survives the incremental path (fails on old code)
- `test_incremental_buildup_still_marks_plain_nonconvergence` — flag 1 still becomes 4

Verified: the 3 preservation tests fail against the pre-fix code and pass with it; `test_iod_picker.py` (16), `test_iod_auto.py` (4), `test_incremental_driver.py` + `test_ias15_settings.py` (18) all pass. `black --check` clean (line-length 110).

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -978,7 +978,14 @@ def do_fit(
         logger.debug(
             f"Primary interval: no root converged " f"(best csq={x.csq:.3g}, n_roots={len(candidates)})"
         )
-        x.flag = 3
+        if x.flag == 1:
+            # Only a plain "did not converge" verdict becomes the driver's
+            # "no root converged" marker (3). A post-convergence gate verdict
+            # from the fitter (2 = chi-square per dof above threshold, 6 =
+            # degenerate covariance) is more informative than 3 and must
+            # survive: the least-bad candidate is exactly where that reason
+            # matters (issue #499).
+            x.flag = 3
         return x
 
     # Attempt to fit all the data, given the fit of the primary interval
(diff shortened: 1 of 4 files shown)


matthewholman:
Thanks for these. We'll implement both from the issues rather than merge -- we keep changes to work authored within the project. #499 and #500 stay open.

Manny7717:
Thanks for the clear note, @matthewholman — understood that layup keeps feature work authored in-house. Closing this PR since you're implementing #499 internally. Appreciate the quick response!

PR #506, opened 2026-08-29
Dispatch layup verbs from installed metadata, not PATH (#500)

### LINKED ISSUE #500: Three tests run whichever `layup` is first on PATH, not the one under test

matthewholman:
`layup <verb>` chose which program to run by `PATH` order.

`find_layup_verbs()` walked every `PATH` directory for executables named
`layup-*`, and the dispatcher ran `layup-<verb>` as a bare name, so resolution
was left to the OS.

Two consequences. With two layup installations on one machine, `layup comet`
from one environment can execute the other's `layup-comet` -- that is how this
surfaced, and it is what made the three CLI tests in #500 fail. And any
executable named `layup-<verb>` in a writable directory earlier on `PATH` runs
in preference to the real one, with the user's privileges; an empty `PATH`
entry means the working directory, and shared systems often carry
group-writable bin directories.

Verbs now come from the installed distribution's entry points and run in this
process, so nothing is resolved by name.

Also pins the three CLI subprocess tests to this environment's console script.
With the conda installation first on `PATH` -- the configuration that was
failing -- all three now pass. Full suite: 488 passed, 1 skipped.

Closes #500. Suggest v1.0: this is a correctness problem in the shipped CLI
rather than the test hygiene the issue describes.

### DIFF
--- a/src/layup_cmdline/main.py
+++ b/src/layup_cmdline/main.py
@@ -8,12 +8,11 @@ from importlib.metadata import distribution
 
 
 def find_layup_verbs():
-    """The verbs this installation provides, as a dict of name -> entry point.
+    """Return the verbs this installation provides, as a dict mapping verb name
+    to entry point.
 
-    Read from the installed package's own metadata. Do not go back to searching
-    PATH for executables named layup-<verb>: that ran whichever layup came first
-    on PATH, which on a machine with more than one installation was often not
-    the one the user meant.
+    The names come from the installed package's own metadata, so they are the
+    verbs belonging to this layup, not whichever ones happen to be first on PATH.
     """
     verbs = {}
     for ep in distribution("layup").entry_points:
--- a/src/layup_cmdline/main.py
+++ b/src/layup_cmdline/main.py
@@ -1,34 +1,25 @@
 import argparse
 import sys
+from importlib.metadata import distribution
 
 #
 # Generic verb dispatcher code
 #
 
 
-def _verb_entry_points():
-    """The ``layup-*`` console scripts *this* installation declares, by verb.
+def find_layup_verbs():
+    """The verbs this installation provides, as a dict of name -> entry point.
 
-    Taken from the installed distribution's own metadata rather than by
-    searching ``PATH``. Searching ``PATH`` picked up any executable named
-    ``layup-<verb>`` anywhere on it, so with two layup installations on one
-    machine the dispatcher could run the other one's verb -- and an executable
-    dropped in any writable directory earlier on ``PATH`` (an empty ``PATH``
-    entry means the working directory) would be run in preference to the real
-    one, with the user's privileges.
+    Read from the installed package's own metadata. Do not go back to searching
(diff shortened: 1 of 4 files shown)


src/layup_cmdline/main.py
```diff
@@ -1,23 +1,34 @@
...
+    """The ``layup-*`` console scripts *this* installation declares, by verb.
+
+    Taken from the installed distribution's own metadata rather than by
+    searching ``PATH``. Searching ``PATH`` picked up any executable named
+    ``layup-<verb>`` anywhere on it, so with two layup installations on one
+    machine the dispatcher could run the other one's verb -- and an executable
+    dropped in any writable directory earlier on ``PATH`` (an empty ``PATH``
+    entry means the working directory) would be run in preference to the real
+    one, with the user's privileges.
+    """
```
mschwamb, line 16: This whole docstring isn't what this code is doing adn is convoluted- it's what the AI thinks it fixed. I think it needs to be rewritten by a human for who is going to edit this code in 6 months or 6 years from now

matthewholman: Could you please take charge of that, @mschwamb ?

src/layup_cmdline/main.py
```diff
@@ -1,23 +1,34 @@
...
+    from importlib.metadata import distribution
+
+    verbs = {}
+    for ep in distribution("layup").entry_points:
+        if ep.group == "console_scripts" and ep.name.startswith("layup-"):
+            verbs[ep.name[len("layup-") :]] = ep
+    return verbs
```
mschwamb, line 21: Do we need this in a separate function? It's a bit of Russian doll to read though. These lines could easily be in find_layup_verbs() and it would say having to dive into a separate function

mschwamb (changes requested):
Some feedback to consider

src/layup_cmdline/main.py
mschwamb: I'd get rid of that function and that would solve the issue. As per my other feedback comment.

matthewholman:
I think this addresses your concerns, @mschwamb

mschwamb (approved):

PR #507, opened 2026-08-29
Keep post-convergence gate verdicts in the fit flag (#499)

### LINKED ISSUE #499: Document that flag under-reports gate rejections; triage on the failed_* columns

matthewholman:
`do_fit` assigned its own markers unconditionally -- 3 where no IOD root
converged, 4 where the incremental build-up stopped -- overwriting whatever
the fitter returned.

A candidate that converged and was then rejected by a gate (2 = chi-square per
degree of freedom above threshold, 6 = degenerate covariance) was therefore
reported as a fit that never converged. Anything triaging on flag 2
under-counts.

Only a plain non-convergence now becomes the driver's marker. The three values
it assigns are named rather than bare integers, since they are an output
format.

Four tests drive `_run_fit` through each path. Against the previous behaviour
the two gate cases fail with "gate verdict 2 was overwritten with 3"; the
plain-non-convergence case passes either way, so the tests pin the distinction
rather than the change.

Closes #499.

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -34,6 +34,14 @@ except ImportError:  # extension not rebuilt yet
 # _MU_SUN (= heliocentric GM = k^2) is used by the BK-native fit for the
 # bound-orbit energy prior on gdot; SPEED_OF_LIGHT (au/day) by the radar ingest.
 from layup.constants import MU_SUN as _MU_SUN, SPEED_OF_LIGHT
+
+# Flag values the driver assigns. The fitter's own verdicts (0 converged,
+# 1 did not converge, 2 chi-square per degree of freedom above threshold,
+# 6 degenerate covariance) come from the C++ side; these are the driver's.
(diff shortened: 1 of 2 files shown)


matthewholman:
Superseded by #513: the gate verdict now has its own column, so there is nothing left for this guard to protect.

Issue #509, opened 2026-08-30
gauss_iod picks the first, middle and last observation, which is too wide on long arcs


matthewholman:
`gauss_iod` takes `seq[0][0]`, the midpoint and `seq[0][-1]` (`iod.py:104-106`).
On a 67-day arc that hands a short-arc approximation three points a month apart.

Fitting 500 Rubin-only arcs (MPC observatory code X05) with the #465 close-Earth
root deferred leaves five `flag=3` failures. They are not the information-starved
cases — for the discoveries they have **longer** arcs than the converged ones,
40 d median against 18 d. Choosing a triplet by time span instead:

| object | arc | first/mid/last | ~2 d | ~5 d | ~10 d | ~20 d | MPCORB a | \|da/a\| |
|---|---|---|---|---|---|---|---|---|
| K25OI3O | 40.1 d | flag 3 | ok | ok | ok | ok | 2.1913 | 6.0e-4 |
| K25MU7E | 38.1 d | flag 3 | ok | ok | ok | flag 3 | 2.7095 | 3.9e-3 |
| K25MY9V | 67.1 d | flag 3 | flag 3 | ok | ok | ok | 44.9821 | 4.6e-3 |
| K21RF8T | 11.0 d | flag 3 | flag 3 | ok | ok | ok | 2.6692 | 4.1e-3 |
| K14E02J | 11.3 d | flag 3 | ok | ok | ok | ok | 2.2394 | 4.7e-3 |

A ~5-day triplet converges all five on the catalogue orbit, including a TNO at
45 au from a Rubin-only arc. With #465 and a 5-day triplet the sample is 500/500.

The usable span is two-sided and depends on the object: a = 2.71 fails at 20 d
(too long for the series), a = 45 fails at 2 d (too little curvature). So a fixed
constant is not the answer either.

Cheapest fix that needs no theory: try triplets at several spans and pool the
candidates. `_pick_best_root` already selects among multiple candidates on
chi-square, so the machinery exists.

matthewholman:
Measured, on a 2000-object cold start.

The premise holds. Replacing first/middle/last with a triplet targeted at a
fixed **60-day** span takes convergence from 95.55% to 98.65% at 1.37x runtime;
a cost-ordered ladder on top of it reaches 99.50%. Across three populations
(2243 objects) that is 92.78% -> 98.31%, NEOs gaining most at 65% -> 87%.

Two natural fixes both fail, in opposite directions:

[...]

PR #510, opened 2026-08-30
Defer close-approach IOD candidates instead of fitting them in polynomial order

### LINKED ISSUE #465: Unusually long orbitfit execution time for certain MBAs

matthewholman:
Closes #465.

A Gauss root that grazes the Earth is expensive rather than obviously wrong:
the integrator resolves the encounter step by step and one candidate can
consume the whole fit budget. On short main-belt arcs it is the degenerate
branch that collapses onto Earth's own orbit, and it is rejected at
1e4-1e6 sigma once it is actually evaluated (traces in the issue).

This partitions the candidates on the existing `_CLOSE_EARTH_AU`, fits the
rest first, and returns to the close ones only if none of the others
converged. **Deferred, not discarded** -- for a genuine near-Earth object with
a real close approach the close root is the correct orbit, which is why
(308635) 2005 YU55 and (367943) Duende must keep working.

No new tunable, no change to the residual prefilter.

**Effect.** 500 Rubin-only arcs (MPC observatory code X05), 420 s deadline:

| | discoveries | known |
|---|---|---|
| converged, before | 229 (91.6%) | 201 (80.4%) |
| converged, after | **247 (98.8%)** | **248 (99.2%)** |
| exceeded the deadline, before | 18 | 37 |
| exceeded the deadline, after | **0** | **0** |

The quality statistics are unchanged -- median reduced chi-square identical at
0.0061 and 0.0023, formal sigma_a/a and |da/a| moving in the third digit -- so
the newly-converging objects are ordinary fits that were previously never
reached, not marginal ones squeaking in. Six arcs that returned nothing after
420 s now finish in ~0.04 s.

Also 96.8% of a 400-object MPC sample now fits from a cold start with no
initial guess at all.

**Tests.** Six unit tests on the partition: a far candidate is safe, a close one
is deferred and never dropped, ordering and completeness are preserved, the
threshold is the shared constant, motion toward the observer during the arc is
detected, and an unreadable candidate degrades to safe. Full suite 491 passed,
1 skipped. (`test_comet_output` fails locally on a stale `layup` on PATH -- the
problem #506 fixes -- and is unrelated.)

The five arcs that still return flag 3 are a separate matter: #509.

### DIFF
--- a/src/layup/iod.py
+++ b/src/layup/iod.py
@@ -181,6 +181,47 @@ def _inertial_min_geocentric_AU(state, state_epoch, observations) -> float:
     return math.sqrt(min_d2)
 
 
+def partition_close_approach(observations, candidates, close_earth_AU: float = _CLOSE_EARTH_AU):
+    """Split IOD candidates into those safe to integrate and those to defer.
+
+    A Gauss root whose trajectory passes very close to Earth is expensive
+    rather than wrong-looking: the integrator resolves the encounter step by
+    step and a single candidate can consume the whole fit budget. On short
+    main-belt arcs the offender is usually the degenerate branch that
+    collapses onto Earth's own orbit -- near-zero topocentric distance,
+    co-moving with the observer -- which is spurious for the objects that
+    produce it and is rejected at 10^4-10^6 sigma when it is actually
+    evaluated (Smithsonian/layup#465).
+
+    These candidates cannot simply be dropped: for a genuine near-Earth
+    object with a real close approach, the close root is the *correct*
+    orbit. So they are deferred, not discarded. The caller fits the safe
+    candidates first and only falls back to the deferred ones if none of
+    them converged.
+
+    The same ``close_earth_AU`` threshold and the same inertial
+    approximation are used as in :func:`filter_candidates_by_residual`,
+    which passes these candidates through its residual test unfiltered for
+    the same reason.
+
+    Returns
+    -------
+    (safe, deferred) : tuple[list, list]
+        Both preserve the input ordering. ``deferred`` is empty in the
+        common case.
+    """
+    safe, deferred = [], []
+    for c in candidates:
+        try:
+            min_geo = _inertial_min_geocentric_AU(c.state, c.epoch, observations)
+        except Exception:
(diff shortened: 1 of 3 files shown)


kjnapier (commented):
Awesome!

kjnapier (approved):

PR #513, opened 2026-08-31
Report fit outcome as independent columns instead of one flag (#499)

matthewholman:
Adds five columns reporting the fit outcome as independent facts: `converged`, `stage`, `failed_csq`, `failed_cov`, `failed_physical`. `do_fit` takes an optional record to fill, so its return type and every existing caller are unchanged.

`flag` is untouched and stays the summary: 0 if and only if the fit converged and every check passed. No existing filter or column changes meaning.

Supersedes #507. #497 becomes one line setting `failed_physical`.

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -31,7 +31,7 @@ try:
 except ImportError:  # extension not rebuilt yet
     get_ias15_adaptive_mode = lambda: -1
     set_ias15_adaptive_mode = lambda m: None
-# _MU_SUN (= heliocentric GM = k^2) is used by the BK-native fit for the
+# MU_SUN (= heliocentric GM = k^2) is used by the BK-native fit for the
 # bound-orbit energy prior on gdot; SPEED_OF_LIGHT (au/day) by the radar ingest.
 from layup.constants import (
     CONVERGED_FLAGS,
@@ -45,7 +45,7 @@ from layup.constants import (
     FLAG_NO_SOLUTION,
     FLAG_NOT_ATTEMPTED,
     FLAG_PRIOR_NOT_POSITIVE_DEFINITE,
-    MU_SUN as _MU_SUN,
+    MU_SUN,
     OUTCOME_COLUMNS,
     SPEED_OF_LIGHT,
     STAGE_BUILDUP,
@@ -152,7 +152,7 @@ def _run_fit(assist_ephem, initial_guess, observations, engine, iter_max=100):
     if engine == "cartesian":
         return run_from_vector_with_initial_guess(assist_ephem, initial_guess, observations, iter_max)
     if engine == "bk_native":
-        return run_bk_native_fit(assist_ephem, initial_guess, observations, _MU_SUN)
+        return run_bk_native_fit(assist_ephem, initial_guess, observations, MU_SUN)
     raise ValueError(f"Unknown engine {engine!r}; expected one of 'cartesian', 'bk_native'.")
 
 
@@ -1077,7 +1077,7 @@ def do_fit(
         # diagnostic scan Gauss+BK covers ~90% of cases vs ~84% for Gauss alone.
         # Epoch convention matches do_gauss_iod's middle observation.
         logger.debug(f"All {len(solns)} Gauss roots failed; trying BK-IOD fallback")
-        bk_seed = run_bk_iod(obs, float(obs[len(obs) // 2].epoch), _MU_SUN)
+        bk_seed = run_bk_iod(obs, float(obs[len(obs) // 2].epoch), MU_SUN)
         if bk_seed.flag == 0:
             cand = _run_fit(assist_ephem, bk_seed, obs, engine, full_iter_max)
             candidates.append(cand)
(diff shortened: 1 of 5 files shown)


src/layup/orbitfit.py
```diff
@@ -33,7 +33,28 @@
...
+    FLAG_BUILDUP_FAILED,
+    FLAG_CONVERGED,
+    FLAG_CSQ_TOO_LARGE,
+    FLAG_DEGENERATE_COV,
+    FLAG_INCREMENTAL_NO_FULL_OBS,
+    FLAG_NO_ROOT_CONVERGED,
+    FLAG_NO_SOLUTION,
+    FLAG_NOT_ATTEMPTED,
+    FLAG_PRIOR_NOT_POSITIVE_DEFINITE,
+    MU_SUN as _MU_SUN,
```
mschwamb: is there a reason to do this?

tests/layup/test_fit_outcome.py
```diff
@@ -0,0 +1,191 @@
...
+    """A fit that never converged must not look like a gated one."""
+    result, outcome = _drive(monkeypatch, [FLAG_DID_NOT_CONVERGE] * 3)
+    assert result.flag == FLAG_NO_ROOT_CONVERGED
+    assert outcome.converged is False
+    assert outcome.failed_csq is False and outcome.failed_cov is False
+    assert outcome.stage == STAGE_PRIMARY
+
+
+def test_clean_fit_is_complete_and_ungated(monkeypatch):
+    result, outcome = _drive(monkeypatch, [0, 0, 0])
```
mschwamb: What's this test doing?

src/layup/orbitfit.py
matthewholman: No — it predated this PR; expanding the import just made it visible. Dropped in `cab3ed4`, and the two tests that reached for it now take `MU_SUN` from `constants`.

tests/layup/test_fit_outcome.py
matthewholman: It's the all-succeeds case: the screen and the full-set refit both converge, so no build-up runs and nothing rejects the fit — `flag` is 0 and every column agrees. It had no docstring and bare `0` literals; both fixed in `51dd979`.

tests/layup/test_fit_outcome.py
```diff
@@ -0,0 +1,191 @@
...
+    assert result.flag == FLAG_CONVERGED
+    assert outcome.converged is True
+    assert outcome.stage == STAGE_COMPLETE
+    assert not any((outcome.failed_csq, outcome.failed_cov, outcome.failed_physical))
+
+
+def test_no_iod_candidates_reports_that_stage(monkeypatch):
+    monkeypatch.setattr(orbitfit, "get_iod", lambda name: (lambda obs, s: []))
```
mschwamb, line 125: what's this test doing? I can sorta understand from the name and code, but it would be helpful to have a docstring.

tests/layup/test_fit_outcome.py
```diff
@@ -0,0 +1,191 @@
...
+        (FLAG_DEGENERATE_COV, True, STAGE_COMPLETE),
+        (FLAG_DID_NOT_CONVERGE, False, STAGE_NOT_ATTEMPTED),
+        (FLAG_NO_ROOT_CONVERGED, False, STAGE_PRIMARY),
+        (FLAG_BUILDUP_FAILED, False, STAGE_BUILDUP),
+        (FLAG_NO_SOLUTION, False, STAGE_NO_CANDIDATES),
+        (FLAG_INCREMENTAL_NO_FULL_OBS, False, STAGE_INCREMENTAL),
+    ],
+)
+def test_from_flag_reconstructs_what_the_summary_still_permits(flag, converged, stage):
+    outcome = FitOutcome.from_flag(flag)
```
mschwamb, line 169: what's this test doing? I can sorta understand from the name and code, but it would be helpful to have a docstring.

tests/layup/test_fit_outcome.py
```diff
@@ -0,0 +1,191 @@
...
+    assert outcome.converged is converged
+    assert outcome.stage == stage
+
+
+def test_as_row_is_ints_in_column_order():
```
mschwamb, line 174: what's this test doing? I can sorta understand from the name and code, but it would be helpful to have a docstring.

tests/layup/test_fit_outcome.py
```diff
@@ -0,0 +1,191 @@
...
+    assert dt.names[-2:] == ("obs_hash", "nobs_fit")
+    assert set(OUTCOME_COLUMNS) <= set(dt.names)
+    assert dt.names.index("flag") < dt.names.index("converged")
+
+    dt_ng = orbitfit._get_result_dtypes("ObjID", ("A2",))
+    assert dt_ng.names[-4:] == ("a2", "a2_unc", "obs_hash", "nobs_fit")
+    assert dt_ng.names.index("failed_physical") < dt_ng.names.index("a2")
+
+
+def test_empty_result_reports_not_attempted():
```
mschwamb, line 205: what's this test doing? I can sorta understand from the name and code, but it would be helpful to have a docstring.

mschwamb (commented):
A few suggestions for consideration

mschwamb (approved):

PR #514, opened 2026-08-31
remove unused AI generated notes

mschwamb:
There are AI generated notes in the JOSS paper draft. It splits the document. This PR removes them as the paper is complete.


matthewholman (approved):
Build-stripped comments, so the rendered paper is unchanged. Content kept in a working note.

PR #519, opened 2026-09-01
Report the chi-square on fits that were rejected rather than accepted

### LINKED ISSUE #518: csq is NaN for fits rejected on chi-square
`csq` is written only when the fit succeeded — `orbitfit.py:1556`,
`(res.csq if success else np.nan)` with `success = res.flag == 0`. So a fit that
reached a solution and was then rejected reports `csq = NaN`: the one number
that explains the rejection is discarded exactly when it is wanted.

`ndof` on the next line is *not* gated, so the output currently carries a
degrees-of-freedom count alongside a NaN chi-square.

Found diagnosing the objects a cold start cannot fit. Across two populations 36
of 37 failures carry `failed_csq = 1` — they converged and were rejected — and
for none of them can we see whether the reduced chi-square missed by a factor of
two or by 1e9, which are different diagnoses with different fixes.

**Suggested condition.** Write `csq` whenever the fitter produced one, i.e.
everything except flag -1 (never attempted) and flag 5 (no candidates). That
includes flags 3 and 4: `do_fit` returns `min(candidates, key=csq)`, so the value
is the least-bad chi-square among the candidates, which is the diagnostic number.

Worth keeping to `csq` alone. `success` also gates the state, covariance, epoch
and FORMAT, and writing states for non-converged rows would be a real behaviour
change for anything treating `flag == 0` as "state present".

matthewholman:
`csq` is now written for every flag except -1 (never attempted) and 5 (no
candidate found to score). The state, covariance and epoch stay gated on
`flag == 0`, so a finite chi-square on a rejected row cannot be read as an
accepted orbit.

This makes the stage markers diagnosable. On ten main-belt objects a cold start
cannot fit -- all flag 3 or 4, all previously NaN -- the chi-squares separate
into two groups the flags could not distinguish: four between 22 and 2652, and
six between 5e7 and 2e10. The first group is a weighting or outlier problem on a
good orbit; the second is a wrong orbit.

Closes #518.

### DIFF
--- a/src/layup/constants.py
+++ b/src/layup/constants.py
@@ -109,3 +109,9 @@ CXX_GATE_FLAGS = {FLAG_CSQ_TOO_LARGE: "failed_csq", FLAG_DEGENERATE_COV: "failed
 
 # The flags that mean the differential correction reached a solution.
 CONVERGED_FLAGS = (FLAG_CONVERGED, FLAG_CSQ_TOO_LARGE, FLAG_DEGENERATE_COV)
+
+# The flags for which no chi-square exists to report: one where no fit was ever
+# run, and one where no initial-orbit candidate was found to score. Every other
+# flag carries a chi-square -- including the stage markers, since ``do_fit``
+# returns the lowest-chi-square candidate it tried.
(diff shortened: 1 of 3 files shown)


kjnapier (approved):
Looks good

PR #524, opened 2026-09-02
Propagate fitted non-gravitational parameters in predict

### LINKED ISSUE #522: predict and residuals_at_state ignore fitted non-gravitational parameters
`predict` and `residuals_at_state` propagate gravity-only, even when the fit
solved for non-gravitational parameters. `FitResult` carries `a1`, `a2`, `a3`,
their uncertainties and `nongrav_mask`, and neither path reads any of them --
`predict.cpp` and `residuals_at_state` (`orbit_fit.cpp:648`) contain no reference
to non-gravs, while the fitting path threads `nongrav_mask` and `a123` through
(`:478`, `:589`). There is no warning.

The predicted position, and the uncertainty ellipse computed about it, are
therefore both wrong for any object fitted with A1/A2/A3.

Measured on (152563) 1992 BF, 644 optical observations over a 71-year arc, fitted
with `fit_nongrav="A2"`:

```
gravity-only fit   csq 1133.9   chi2 from residuals_at_state 1133.9   ratio 1.00
A2 fit             csq  665.5   chi2 from residuals_at_state 4779.9   ratio 7.18
```

The A2 fit's own residuals disagree with its reported chi-square by a factor of
7, because `residuals_at_state` recomputes them without A2. On the four 1953
precovery plates the discrepancy is ~15 arcsec -- which is the expected size,
since Yarkovsky is what links those observations at all (Vokrouhlicky, Chesley &
Matson 2008).

Found while testing whether Layup recovers a published Yarkovsky detection; it
does, at 31.5 sigma.

matthewholman:
Fixes #522.

`predict` re-integrated a fitted orbit under gravity alone, discarding any A1/A2/A3 the fit had solved for. For a comet, or an asteroid with a detected Yarkovsky acceleration, the prediction drifted away from the orbit that produced it.

Two halves were missing. `predict.cpp` never enabled `ASSIST_FORCE_NON_GRAVITATIONAL` or set `particle_params`, and `parse_fit_result` dropped the `a1`/`a2`/`a3` columns on the way from the fit table back into a `FitResult` — so fixing either one alone leaves the bug in place. `residuals_at_state` now passes the parameters through as well.

Verified on 152563 (1992 BF), whose A2 is well determined over a 1953–2016 arc: residuals at the fitted state drop from 15" to 2" in 1953, and the predicted-to-fitted agreement is 1.000 for both gravity-only and A2 fits. Two regression tests, each confirmed to fail when either half of the fix is reverted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,7 +53,9 @@ residual rows they contribute, not in the code path they take.
 - Ephemeris-quality integrations through ASSIST and REBOUND, using the IAS15
   integrator.
 - Predictions that integrate once and interpolate to many epochs and observatory
-  locations.
+  locations, propagating any non-gravitational acceleration the fit solved for, so
+  a predicted position and its uncertainty ellipse are consistent with the orbit
+  that produced them.
 
 ### Interfaces
 
--- a/src/layup/utilities/data_processing_utilities.py
+++ b/src/layup/utilities/data_processing_utilities.py
@@ -362,6 +362,23 @@ def parse_fit_result(
             cov[i] = 0.0
     res.cov = cov
 
+    # Carry the fitted non-gravitational parameters back onto the FitResult
+    # (issue #522). The a1/a2/a3 columns exist only when a non-gravitational fit
+    # ran, so their presence is what identifies which were active. Without this
+    # the parameters are silently dropped at the Python boundary and every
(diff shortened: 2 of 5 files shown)


kjnapier (approved):
Good catch

PR #525, opened 2026-09-02
Add the 55 X05 short arcs as an initial-orbit fixture

matthewholman:
The 55 Rubin-only arcs (MPC obscode X05, 6-20 observations each) whose cold-start Gauss fit ran past a 420 s deadline and returned nothing. These are the cases an alternative initial-orbit method exists to cover, and there was no fixture for them.

Reference elements are MPCORB's, fitted to each object's full observation history rather than to the short arc, so they are an external check and not a restatement of layup's own answer. Published semi-major axes run 2.10 to 3.93 au: 54 main belt and one Hilda-region object. That is the regime the Herget prototype's hardcoded 40 au starting range does not reach, which is what makes the choice of that range measurable.

The tests here run unconditionally and need no initial-orbit method to exist: the arcs and the manifest have to agree, the sample has to stay in the regime that makes it interesting, and the arcs have to stay short. The test that fits these arcs and compares against the published orbits skips wherever `herget` is not registered, so it has been held back to travel with the method on `feat/herget-x05-fit-test` rather than sit here skipped.

Measured over the sample, with the method from #502: 42 of 53 converge to a median |da/a| of 2.6e-3 against MPCORB; 8 raise a Kepler convergence error, 3 find no root, and 2 exceed a 180 s deadline.

Relevant to #311 and #502.

The arcs are one obs80 file rather than 55, each line carrying its own designation.


mschwamb:
Why do we need all of these in the repo as separate files?

mschwamb:
Is there a strong reason to add in ~50 files for a unit test?

mschwamb (changes requested):
See comments

mschwamb:
I'm confused how this tests Herget IOD, if Herget is not implemented into the main branch.

matthewholman:
These are the 55 arcs from your #502 question, so the starting range there can be checked against something external.

Fair point on the files -- consolidated to a single obs80, now 3 files instead of 57. Every line carries its own designation, so nothing is lost.

Three of the four tests run on `main` and check the fixture itself. Only the fitting test needs Herget, and both skips now say why they skipped.

mschwamb:
I don't think we should have tests in for something not implemented. That's dangerous if we forget to change it.

mschwamb:
Skipped tests should go into a separate PR or be designed to not require something to be skipped.

matthewholman:
Dropped the fitting test. What's left runs unconditionally and needs no IOD method to exist: arcs match the manifest, the sample stays in regime, the arcs stay short.

The fitting test is on `feat/herget-x05-fit-test` and will go in with #502.

mschwamb:
Do these files need to be labeled herget? Will that be confusing for maintenance?

matthewholman:
Agreed -- renamed to `x05_short_arcs.*`. With the fitting test gone nothing here is Herget-specific, so the label would have sent someone looking for code that isn't in these files.

mschwamb (approved):

PR #526, opened 2026-09-02
Read the obs80 star catalogue under its ADES name

### LINKED ISSUE #521: Debiasing is silently skipped when the data has no astCat column
`_orbitfit` takes the star catalog from an `astCat` column
(`orbitfit.py:1304`, used at `:1349` and `:1423`). `Obs80DataReader` produces
`cat`, so fitting obs80 data with `bias_dict=generate_bias_dict(...)` applies no
debiasing at all, and says nothing.

The per-observation fallback is deliberate and right -- historical astrometry
often has no catalog, and `debias()` documents that. The gap is the whole-file
case, where debiasing was explicitly requested and is entirely inert.

Suggestion: warn once when `bias_dict` is supplied but no observation resolves to
a bias model.

Found while comparing Rubin-only fits against JPL. A debiased and an undebiased
run agreed to 1e-6, which read as evidence that debiasing does not matter here.
After renaming `cat` to `astCat` it moves individual semimajor axes by up to
5e-3.

matthewholman:
Closes #521.

The MPC 80-column format carries the star catalogue in column 72, and the reader emitted it as `cat`. Everything downstream reads it as `astCat` -- the Farnocchia/Chesley debiasing and the Veres (2017) weighting both key off that name -- so `astcat_column_present` was False, `catalog=None` went to the bias lookup, and the correction quietly did nothing. Nothing raised and nothing was logged, so a caller asking for debiased astrometry got undebiased astrometry back with no way to tell.

The reader now emits `astCat` and `program`, matching every other reader. Nothing outside the reader referenced the old names, and no fixture carries them.

Also logs a warning when debiasing is requested and no `astCat` column is present, so any other input path fails loudly instead of silently.

Worth knowing: the catalogue and Rubin results are unaffected. Every harness that enabled debiasing did this rename by hand first, and the runs that did not enable it were undebiased by choice. This is a defect for anyone using the CLI on 80-column input, which is the most likely way to reach the feature.

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -1339,6 +1339,14 @@ def _orbitfit(
 
         # bias_dict will be a dictionary when the debias flag is set to True.
         if bias_dict is not None:
+            if not astcat_column_present:
+                # Without a star catalog per observation there is no bias to look
+                # up, so debiasing would quietly do nothing. Say so rather than
+                # return astrometry the caller believes was corrected.
+                logger.warning(
+                    "Debiasing was requested but the data has no astCat column, so no "
+                    "star-catalog bias can be applied. The astrometry is unchanged."
+                )
             for d in data:
                 if radar_columns_present and _is_radar(d, column_names):
                     continue  # debiasing is an astrometric (ra/dec) correction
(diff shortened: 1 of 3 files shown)


kjnapier (approved):
Looks good to me

PR #527, opened 2026-09-02
Add an accepted column so a usable fit reads as a predicate

### LINKED ISSUE #498: Decide what the orbit fit status column is called before v1.0
Raised by @hannorein and @mschwamb on #495: rename the `flag` column to
`error` or `success` so code reads `if (!error)`.

Two things to settle before anyone implements it.

**They are not interchangeable.** 0 means converged today, so `if (!error)`
works with the values unchanged, but `if (success)` is false for every good fit
unless the values also invert. "error or success style" covers both.

**Renaming is a breaking change.** `"flag"` appears 58 times across
`orbitfit.py`, `unpack.py`, `convert.py` and 16 test files, and 7 fixture CSVs
in `tests/data/` carry it as a column header. Output already written stops
loading. Cheap before v1.0, not after.

**Alternative:** keep `flag`, add a derived boolean column (`accepted`). Reads
the same, breaks nothing, inverts nothing, and gives `-sf` (#490) a named
predicate. Also avoids labelling -1 (not attempted) and 8 (bookkeeping) as
errors, which they are not.

Flag values are documented in #495.

matthewholman:
Closes #498.

Reviewers on #495 asked for a status column that reads as a predicate -- `if (!error)` or `if (success)` -- rather than an integer compared against a documented list.

The two are not interchangeable: `0` means converged today, so `!error` works with the values unchanged while `success` is false for every good fit unless the values also invert. And renaming `flag` is breaking -- seven fixture CSVs carry it as a column header, so output already written stops loading.

This is the alternative from #498 instead: keep `flag`, add a derived boolean. `accepted` is 1 exactly when `flag == 0`. It reads the same, inverts nothing, and does not label -1 (never attempted) or 8 (incremental bookkeeping) as errors, which they are not.

It is derived from the flag the row carries rather than tracked beside the other outcome facts, so the two cannot disagree, and it is taken after any stage marker has overwritten the fitter's own verdict -- that overwritten value is what a reader sees.

Two things came along with it:

- `create_empty_result` hand-copied the outcome tuple, a second copy of the column layout that would have silently desynced. It now calls `as_row`.
- Corrects the `failed_cov` entry in `docs/fit_flags.rst`. Both of its assignments sit inside the non-gravitational weak-constraint guard, so a gravity-only fit is never marked there. The page implied the covariance was examined for every fit, which would let a reader take a column of zeros as reassurance.

### DIFF
--- a/docs/fit_flags.rst
+++ b/docs/fit_flags.rst
@@ -1,9 +1,9 @@
 Orbit fit status
 ========================================================================================
 
-Every fit that ``layup orbitfit`` produces reports its outcome in six columns: a
-single-value summary, ``flag``, and five columns that report the individual facts
-behind it.
+Every fit that ``layup orbitfit`` produces reports its outcome in seven columns: a
+single-value summary, ``flag``, a plain ``accepted`` predicate, and five columns
+that report the individual facts behind them.
 
 .. list-table::
    :header-rows: 1
@@ -11,6 +11,10 @@ behind it.
 
    * - Column
      - Meaning
+   * - ``accepted``
+     - ``1`` if the fit is usable: it converged and passed every check. The same
+       thing ``flag == 0`` says, as a predicate you can filter on without knowing
+       the flag values.
    * - ``flag``
      - Summary. ``0`` if and only if the fit converged and passed every check.
    * - ``converged``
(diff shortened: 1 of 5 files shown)


kjnapier (approved):
Okay

Issue #528, opened 2026-09-02
Radar model is monostatic; bistatic observations are silently mis-modelled


matthewholman:
`Observation` carries one station, used for both the transmit and the receive leg. That is monostatic by design — `ISSUE_146_RADAR_DESIGN.md` scopes the first implementation to TX == RX and lists bistatic as a refinement — but a bistatic measurement passed in anyway is not rejected. The receive leg is simply evaluated at the transmitting antenna.

Measured on JPL radar for (6489) Golevka: observed/model Doppler is 1.039–1.044 on bistatic rows against 0.999934 on monostatic ones. Against a 0.11 Hz uncertainty that is roughly a hundred sigma per observation. Golevka has 18 such rows out of 40; including them gives flag 2 with a reduced chi-square of 1.5e7, excluding them gives flag 0 at 0.80.

Bistatic pairs are common in the archive — Goldstone DSS-14 transmitting with DSS-13 or DSS-38 receiving. Supporting them needs a second station on the observation; until then the caller has to filter, which #529 documents in `_radar_observation`.

matthewholman:
A simpler fix than the one this issue was opened with, and it subsumes a second problem.

The up leg currently derives the transmit position by extrapolating the *receive* station backwards:

```
rtx = xe - vo*tau - 0.5*ao*tau^2
```

That is wrong for bistatic in the obvious way -- the transmitting antenna is not the receiving one moved back in time -- and it also carries a truncation error even when TX == RX. Over Golevka's ~100 s round trip the Earth's rotation takes the station about 170 m past the linear term, roughly 1.1 us, against delay uncertainties of 0.3 us.

Both go away if we simply evaluate the station position at each of the two epochs instead of extrapolating between them.

The transmit epoch is not something we have to solve for: **the round-trip delay is the observable**, so the transmit time is `t_receive - delay_observed`. The only error in that is the delay *residual*, of order microseconds, which moves a station by about half a millimetre.

`_append_observer_acceleration` already shifts `et` and re-evaluates `obscodes_to_barycentric`, so evaluating a station at an arbitrary epoch is machinery layup already has -- it is currently used to build a Taylor series where a direct evaluation would do:

```python
shifted = data.copy()
shifted["et"] = data["et"] - delay_seconds   # transmit epoch, from the observable
shifted["stn"] = data["stnTx"]               # the transmitting antenna
tx = observatory.obscodes_to_barycentric(shifted)
```

[...]

PR #529, opened 2026-09-03
Model the Shapiro delay in radar residuals

matthewholman:
Radar signals pass through the Sun's gravitational potential on both legs, and that delay was not modelled. For a leg between A and B,

    dt = (2 GM / c^3) ln[(r_A + r_B + rho) / (r_A + r_B - rho)]

with r_A and r_B heliocentric distances. It is one to two microseconds at the geometries these observations are made, which sounds negligible until you notice that radar delay uncertainties in the JPL database run 0.2 to 4 microseconds, and are 0.30 for Golevka at Arecibo. Leaving it out biases every radar observation by several sigma.

**Validated against real Apophis radar**, on the epochs near the reference orbit where propagation error is negligible: the delay residual falls from **2.16 sigma to 0.312 sigma**, and Doppler is unchanged at 0.28 sigma. The comparison is against an independent Python implementation of the same formula (`ISSUE_146_shapiro_validation.py`), so it checks the physics rather than this code. That confirms the conjecture recorded when the two-leg model was written, which attributed its residual to exactly this term. It also bears on the Apophis covariance work, where the unmodelled term was running at a median 8.5 sigma.

Two things came with it:

- The ephemeris is passed into `compute_radar_residuals` rather than read from `r->extras`. The simulation used for residuals has no `assist_extras` attached, so that pointer is NULL there.
- The radar test fixtures carry the term now. `radar_synthetic.json` was generated with `delay = 2 rho/c` and no Shapiro, so once the fitter modelled it the truth was inconsistent by 19.2 us per observation. There were **two** stale generators: the fixture script, and a second copy inlined in `test_radar_end_to_end`. Fixing both made the end-to-end tests pass with no tolerance touched.

`test_radar_validation`'s chi-square bound did move, and the reason is in the test: the fixture now comes from an independent implementation, which agrees with the C++ to about 0.3 ns per observation rather than exactly. That is 3e-4 of the fixture's own stated uncertainty. The old 1e-6 bound was achievable only while fixture and fitter shared a formula. State tolerances are **tightened** rather than relaxed, 10x on position and 100x on velocity, since those carry the physics.

The partials are left alone: the Jacobian is already a single-leg approximation by design, and a microsecond-scale slowly varying term does not change the step direction.

Suite: 536 passed, 1 skipped on all four CI legs. Related: #528.

**Reviewing this branch requires rebuilding the C++ extension.** `orbit_fit.cpp` and `radar_synthetic.json` change together, so an older binary gets scored against a fixture that now carries the Shapiro term. The fit still converges and absorbs most of the 19.2 us into the orbit, which makes the failure look like a physics error rather than a stale build: `test_radar_validation` reports chi-square 0.0104 against its 1e-4 bound, and the end-to-end test a 3.7e-6 velocity drift. After `pip install -e .` all five radar tests pass.


kjnapier (approved):
Very impressive

PR #530, opened 2026-09-03
Ship the Marsden and asteroidal g(r) laws as named constants

matthewholman:
A non-gravitational fit takes its sublimation law as five bare numbers, `[alpha, nm, nn, nk, r0]`, and `orbitfit`'s docstring already told callers to pass a cometary law without saying what one is. So fitting a comet meant knowing five magic numbers, and getting one wrong gives a fit that converges and answers a different question.

`MARSDEN_1973_GOFR` is the standard water-ice law of Marsden, Sekanina & Yeomans (1973), AJ 78, 211, in the form ASSIST uses (Holman et al. 2023, section III.4), with the values tabulated by Attree et al. (2019), A&A 625, A19, Table 2.

`ASTEROIDAL_GOFR` spells out the inverse-square law a Yarkovsky A2 fit wants. It is already the default when `nongrav_gr` is None, but naming it lets a caller state the choice and compare the two. Its `nn` is set to 0.0 rather than left at ASSIST's default of 5.093 — that default is Marsden's cometary *n* sitting inert, harmless only because `nk` is zero. Set `nk` without also setting `nn` and you silently get your own *k* with Marsden's *n*, which is reachable from the API since all five are exposed as a raw list.

The tests check the laws against the properties that define them rather than against the numbers: normalization at 1 au, the asteroidal case reducing to (1 au/r)^2, the cometary case falling away faster beyond r0, and r0 being where the bracket contributes 2^-nk. A transcription error in any of the five would be caught by the physics.

One detail worth knowing: the Marsden law normalizes to 1 within 2.4e-9 rather than exactly, which is where alpha's ten published significant figures run out. The tolerance is set from that and the reason is stated in the test.

### DIFF
--- a/src/layup/constants.py
+++ b/src/layup/constants.py
@@ -95,6 +95,35 @@ STAGE_BUILDUP = 3  # reached the incremental build-up to all observations
 STAGE_COMPLETE = 4  # fit the full observation set
 STAGE_INCREMENTAL = 5  # sequential-update bookkeeping rather than a fresh fit
 
+# ---------------------------------------------------------------------------
+# Non-gravitational g(r) laws
+#
+# ASSIST parameterizes the sublimation law as
+#
+#     g(r) = alpha * (r/r0)^-nm * (1 + (r/r0)^nn)^-nk
+#
+# which is the form of Marsden, Sekanina & Yeomans (1973), AJ 78, 211, and is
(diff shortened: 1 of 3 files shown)


kjnapier (approved):
Looks good

PR #533, opened 2026-09-03
Choose the Gauss triplet by time span rather than first/middle/last

### LINKED ISSUE #509: gauss_iod picks the first, middle and last observation, which is too wide on long arcs

matthewholman:
Gauss truncates the Lagrange f and g series, so the interval between the outer two observations has to be short against the orbital period -- classically well under 60 degrees of mean anomaly. We were taking the first, middle and last observation of `seq[0]`, and `seq[0]` is by construction the longest-span chunk, so on a long arc the triplet was as wide as it could possibly be. Objects that fail to converge cold span a median **53 degrees** of mean anomaly against **29** for those that fit.

This picks the triplet whose outer span is nearest a target instead.

The target is in mean anomaly, which needs a period, which needs the orbit we are trying to find. It is converted to days using an **assumed a = 2.5 au** rather than the object's own -- about **60 days** -- which keeps it a prior rather than an oracle. Using each object's published a is slightly *worse*, so there is no chicken-and-egg problem here.

A balance guard requires each sub-interval to be at least 10% of the outer span. Without it a repeated epoch can leave the middle observation sitting on an endpoint, giving a zero-length interval and no usable root. Where no triplet qualifies, the old first/middle/last choice is used, so behaviour is never worse than before.

Measured cold on objects from the MPC catalog:

| | old | this PR |
|---|---|---|
| objects the old selection already fit | 400/400 | **398/400** (99.50%) |
| objects it did not (flag 3) | 0/46 | **39/46** (84.8%) |

Net 39 gained against 2 lost. **The orbits are the same ones** -- median \|da/a\| against the old selection is 1.7e-4 -- so this is a convergence improvement, not a different answer. Median runtime 2.8 s on the regression sample.

Worth noting for the issue: #509 proposed pooling several spans and trying each. We measured that at **-13.9 points**, nearly 5x the cost. Choosing the interval beats trying every interval.

Tests: 87 pass (36 IOD/Gauss, 51 wider fit suite).

Closes #509.

### DIFF
--- a/src/layup/iod.py
+++ b/src/layup/iod.py
@@ -192,7 +192,9 @@ def gauss_iod(observations, seq):
         logger.debug("gauss_iod: span selection found no triplet; using first/middle/last")
     idx0_, idx1, idx2 = trip
     logger.debug(f"gauss_iod: indices {idx0_}, {idx1}, {idx2}")
-    solns = gauss(GMtotal, observations[idx0_], observations[idx1], observations[idx2], 0.0001, SPEED_OF_LIGHT)
+    solns = gauss(
+        GMtotal, observations[idx0_], observations[idx1], observations[idx2], 0.0001, SPEED_OF_LIGHT
+    )
     return solns
 
 
--- a/src/layup/iod.py
+++ b/src/layup/iod.py
@@ -31,6 +31,7 @@ implementation through a class hierarchy.
 
 from __future__ import annotations
 
+import bisect
 import logging
 import math
 from typing import Callable, Optional, Sequence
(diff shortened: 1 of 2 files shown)


kjnapier (approved):
Interesting

PR #534, opened 2026-09-03
Use the transmitting antenna for the radar up leg (bistatic)

### LINKED ISSUE #528: Radar model is monostatic; bistatic observations are silently mis-modelled

matthewholman:
`Observation` carried one station and used it for both legs, so a bistatic measurement had its up leg evaluated at the antenna that *received* rather than the one that transmitted. On (6489) Golevka that is roughly 125 sigma per Doppler point -- observed/model 1.039-1.044 on bistatic rows against 0.999934 on monostatic ones -- and 20 of its 40 radar rows are bistatic.

`Observation` now carries the transmitting antenna's state at the transmit epoch and the up leg uses it directly. Where it is absent the receive station is extrapolated exactly as before, so monostatic behaviour is unchanged.

The transmit epoch does not have to be solved for: the round-trip delay is the observable, so it is `t_receive - tau`. Golevka's bistatic rows are all Doppler-only and carry no delay of their own, so tau is interpolated from the object's delay rows, which run smoothly from 46.5 to 47.2 s across its 1995 apparition. Evaluating the station rather than extrapolating to it also removes the Taylor truncation, measured at 2.2e-3 m/s in velocity at that round trip.

The transmitting antenna is named by **`trx`**, the ADES field, so ADES-derived radar columns need no renaming (`stnTx` is accepted as an alias).

Measured on Golevka, all 40 radar observations:

| | flag | reduced chi-square |
|---|---|---|
| bistatic, receive station extrapolated | 2 | **1.487e7** |
| bistatic, transmitter supplied | 0 | **6.68** |
| monostatic rows only (the old workaround) | 0 | 0.80 |

The fit converges on the full set for the first time; the previous workaround required discarding 18 observations. Aten and Nyx are unchanged controls -- neither has bistatic rows, and both move by under 10%.

**One consequence a reviewer should weigh.** The C++ gate is `chi2/ndof > 10`, so this moves Golevka from flag 2 to flag 0: a loudly-rejected fit becomes an accepted one that still carries the residual below. That is worth stating plainly, but I do not think it argues against merging -- the old behaviour only failed loudly by luck. An object with two bistatic rows instead of twenty would have produced a chi-square under 10 and been *quietly* corrupted before this change, which is exactly what this issue is titled for.

**A residual remains**, all of it in the 20 bistatic Doppler rows, at about 5 Hz -- a round-trip range-rate error near 0.1 m/s. Eliminated so far, each by measurement: the bounce point (every Golevka delay row is already `bp=C`, Chesley et al. 2003's re-reduced values), the peak-power reference (dropping those rows improves only 6.68 to 5.02), station identity and coordinates (DSS-13 to DSS-14 checks out at 21.6 km against a 1.05 m/s velocity difference), the velocity truncation (0.12 Hz, ~1 sigma), and uniform weight inflation (the two tight-sigma rows and the eighteen loose ones agree to ~5 Hz, so it is not the weights).

Tests: 5 radar tests pass, on top of the Shapiro work merged in #529 -- the two compose without interference.

Closes #528.

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -476,6 +476,13 @@ def _radar_observation(objID, d, epoch_jd, column_names):
     else:
         observer_acc = [0.0, 0.0, 0.0]
 
+    # Transmitting antenna at the transmit epoch (issue #528). When absent the
+    # C++ falls back to extrapolating the receive station, as it did before.
+    tx_cols = ("txx", "txy", "txz", "txvx", "txvy", "txvz")
+    has_tx = all(c in column_names for c in tx_cols) and not np.isnan(d["txx"])
+    tx_pos = [float(d["txx"]), float(d["txy"]), float(d["txz"])] if has_tx else [0.0, 0.0, 0.0]
+    tx_vel = [float(d["txvx"]), float(d["txvy"]), float(d["txvz"])] if has_tx else [0.0, 0.0, 0.0]
+
     return Observation.from_radar_with_id(
         str(objID),
         delay_days,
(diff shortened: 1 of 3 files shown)


kjnapier (approved):
Looks good

PR #536, opened 2026-09-03
Take the Levenberg-Marquardt step from the Jacobian, not the normal matrix

### LINKED ISSUE #535: Joint state + non-gravitational fits return a rank-truncated step
The fit forms the normal matrix BᵀWB and takes the Levenberg-Marquardt step from it. Squaring the condition number is survivable for a six-parameter orbit but not once a non-gravitational amplitude is added: A2 is of order 1e-14 au/day² alongside positions of order 1 au, so on (6489) Golevka cond(√W B) = 2.6e10 becomes cond(BᵀWB) = 9.2e19 — four orders past what double precision resolves.

The solve then rank-truncates without complaining: the three position components of the step come back as exactly zero, and A2 never leaves its seed. Measured against the six Yarkovsky objects of Farnocchia et al. (2013), A2 was coming out several times too small, or as NaN.

Solving the damped least-squares problem by QR on the weighted Jacobian instead leaves the step unchanged but the condition number unsquared. That recovers 2062 Aten to 0.15σ of the published A2 from optical astrometry alone, against 3.97σ now.

matthewholman:
Closes #535.

Replaces the normal-equation solve with a Householder QR of the augmented weighted Jacobian,

    min || [ √W B ; √λ I ] dX − [ −√W r ; 0 ] ||

which is the same damped step with the condition number no longer squared. A sequential-update prior (#419) enters exactly, as Cholesky rows. BᵀWB is still formed for the covariance and the conditioning check.

Full suite green (549 passed). Cross-check: solving the normal equations after Jacobi equilibration gives bit-identical A2 on all seven test fits, so both routes reach the same least-squares solution.

Two follow-ups this exposes are filed separately: the weak-constraint guard tests the whole correlation matrix rather than the non-grav direction, and the 1e-15 A2 placeholder is not dynamically negligible.

### DIFF
--- a/src/lib/orbit_fit/orbit_fit.cpp
+++ b/src/lib/orbit_fit/orbit_fit.cpp
@@ -833,7 +833,10 @@ namespace orbit_fit
         }
 
         Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic> Bt = B.transpose();
-        C = Bt * W * B + lambda * eye; // This is where the extra term for LM is.
+        // Undamped normal matrix. It is still wanted for the covariance and for
+        // the non-grav conditioning check, but the Levenberg-Marquardt step below
+        // is taken from the Jacobian directly rather than from C.
+        C = Bt * W * B;
 
         grad = Bt * W * resid_v;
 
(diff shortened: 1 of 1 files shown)


kjnapier (approved):
Okay

PR #538, opened 2026-09-03
Carry the retardation denominators in the radar Doppler model

### LINKED ISSUE #537: Radar Doppler model omits the light-time retardation term
The Doppler observable is c·dτ/dt_receive. Differentiating the two implicit light-time equations leaves a retardation denominator on each leg; to first order that reduces to the instantaneous sum of the two one-way range rates, which is what the model computes. The omitted term is a fractional error of order ρ̂·v/c.

Negligible for most targets, not for a fast one. On (6489) Golevka's 1995 apparition the line-of-sight rate reaches 6 km/s, making it 5–7 Hz against stated uncertainties of 0.09–0.40 Hz. Every Doppler residual is negative, −0.11 to −0.24 m/s, and the size grows with the Doppler shift as the term predicts (correlation 0.94).

The same denominator is already applied to the range partials, as `ltdenom`, so the Jacobian and the residual currently disagree. With it in the model Golevka's reduced chi-square goes from 6.53 to 0.79.

matthewholman:
Closes #537.

On (6489) Golevka, with all 40 radar observations: reduced chi-square **6.53 → 0.79**, the delay rows' rms(residual/sigma) 3.39 → 0.54, and the two centre-of-mass Doppler rows go from **63σ to 0.02σ**. Aten improves slightly; Nyx is unchanged — as expected for a term that scales with range rate.

Both fixture generators encoded the old model. One lived outside the repository, which is why it went stale; the first commit brings it in under `tools/` unchanged, the second corrects it with everything else. Only the Doppler values in `radar_synthetic.json` move, by 6e-7 to 1.7e-5 fractionally; the delays are untouched. Full suite green (549 passed).

The A2 figures quoted in #536 need that PR too — the non-grav solve cannot move A2 without it.


kjnapier (approved):
Looks good

PR #540, opened 2026-09-03
Make the non-grav placeholder negligible and the discarded refinement visible

### LINKED ISSUE #539: A failed non-grav refinement is invisible, and its placeholder amplitude is not negligible
Two small things in the same path, both of which hide a failed non-gravitational fit.

**The placeholder.** ASSIST skips the non-grav block when A1=A2=A3=0, so a zero seed leaves the parameter's column zero. The fit substitutes 1e-15 au/day² to avoid that. But that is the order of a real Yarkovsky amplitude: on (6489) Golevka it is 6% of the true A2 and raises the starting chi-square from 12,888 to 138,416 before the first iteration, enough to trip the chi-square gate. The partial does not depend on the seed's magnitude, so a far smaller value works — 1e-20 restores the gravity-only chi-square exactly.

**The reporting.** When the joint fit returns a non-zero flag the driver keeps the six-parameter solution and reports the parameters as NaN, but says so only at DEBUG. The caller sees flag 0, a chi-square belonging to the gravity fit, and NaN parameters, with nothing to explain them.

matthewholman:
Closes #539.

Placeholder 1e-15 → 1e-20 au/day², and the discarded-refinement message moves from DEBUG to WARNING and now names the flag the joint fit actually returned — 6 is a weakly-constrained solution, not a failure to converge, and 2 is a fit rejected on chi-square.

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -1568,7 +1568,17 @@ def _orbitfit(
             if res_ng.flag == 0:
                 res = res_ng
             else:
-                logger.debug("Non-grav refinement did not converge; reporting non-grav params as NaN.")
+                # Not silent: the caller asked for a non-grav fit and is getting a
+                # gravity-only one, with the parameters reported as NaN and a flag
+                # of 0 that refers to the gravity fit. Name the flag the joint fit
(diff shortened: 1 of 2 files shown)


kjnapier (approved):
Looks good

PR #542, opened 2026-09-03
Reject a non-grav fit only when the covariance cannot support an amplitude

### LINKED ISSUE #541: The non-grav weak-constraint guard measures the orbit's conditioning, not the amplitude's
`WEAK_NONGRAV_RCOND` rejects a joint state + non-grav fit when the reciprocal condition number of the whole npar×npar correlation matrix falls below 1e-8. That is not a statement about the non-grav column: a six-parameter orbit fit is itself strongly correlated, so the test is dominated by the orbit underneath. On (6489) Golevka the **state block alone** has rcond 2.6e-10 — twenty-five times below the threshold — while A2 correlates with the state at only 0.91–0.93 and its formal uncertainty, 1.6e-15 au/day², matches the published one. The fit is rejected for the conditioning of its own orbit, so A2 is reported as NaN even after #536.

I tried replacing it with the amplitude's variance inflation factor and **measured that this does not work either**. On a synthetic A2 arc (truth −5.0e-14):

| arc | VIF | A2 ± σ |
|---|---|---|
| 10 d | 5.7e4 | −7.5e−12 ± 2.6e−7 |
| 30 d | 1.2e4 | −3.0e−13 ± 1.8e−8 |
| 90 d | 23 | −5.2e−14 ± 1.3e−10 |
| 4 yr | 19 | −5.00e−14 ± 3.6e−13 |
| 6489 Golevka | 3.2e5 | −10.2e−15 ± 1.6e−15 |

Golevka is a 6σ detection with a *higher* VIF than the useless 10-day arc. Correlation with the state is not what separates them; whether the amplitude is determined is, and the covariance already reports that faithfully.

So the question is what the guard is for. With the solve fixed the state is no longer contaminated, and a large σ is an honest answer rather than a failure. Worth a decision before picking a new criterion — `test_a2_weak_constraint_guard_on_short_arc` encodes the current one.

matthewholman:
Closes #541.

**This is a behaviour change, and the two rewritten tests are the crux — please push back if you disagree with the policy.**

The guard rejected on rcond of the whole npar×npar correlation matrix, which is dominated by the orbit rather than the non-grav column. Radar makes it worse, since it tightens the state: adding seven radar rows to 2062 Aten's 980 optical ones turned a reported A2 = −14.96 ± 1.70 into NaN. All three radar-bearing Yarkovsky objects were suppressed at reduced chi-square 0.28–1.55.

I first tried the amplitude's variance inflation factor and measured that it does not separate the cases either (#541 has the table). What the old test stood in for — a contaminated, over-confident amplitude — was a symptom of taking the step from the squared-condition normal matrix, fixed in #536. A ten-day arc now returns A2 with a formal error five million times its own value, which is the honest answer and, unlike a significance test, does not suppress a well-constrained null. So reject only on a non-positive or non-finite variance.

With this, on merged main: Aten −15.08 ± 1.34 (0.12σ from Farnocchia+2013), Nyx +16.72 ± 2.98 (1.50σ from JPL), Golevka −10.21 ± 1.61 (0.91σ from JPL). Full suite green, 550 passed.

### DIFF
--- a/src/lib/orbit_fit/orbit_fit.cpp
+++ b/src/lib/orbit_fit/orbit_fit.cpp
@@ -87,12 +87,6 @@ namespace py = pybind11;
 
 namespace orbit_fit
 {
-    // Minimum reciprocal condition number of the normal-matrix CORRELATION matrix
-    // for a non-grav (A2) fit to be considered well constrained (issue #351).
-    // Below this the A2 column is effectively collinear with the state (typical on
-    // short arcs) and the fit is reported as weakly constrained (flag = 6).
-    static constexpr double WEAK_NONGRAV_RCOND = 1e-8;
-
     // Arcseconds per radian (180*3600/pi). Converts astrometric/rate
     // uncertainties from arcseconds to radians and scales residuals for display.
     static constexpr double ARCSEC_PER_RAD = 206265.0;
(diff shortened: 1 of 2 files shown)


kjnapier (approved):
Okay

PR #553, opened 2026-09-05
Merging herget with main to check for problems before PR #502

awilson110:
@matthewholman recommended merging with main before merging the other way around


mschwamb:
But you can [rebase](https://www.atlassian.com/git/tutorials/merging-vs-rebasing) to bring main into the PR - so I don't see why we would do the dangerous thing of merging without knowing this is working

mschwamb (changes requested):
There shouldn't be changes to the paper from implemented herget. If all this is changing, this is screwing up the commit history and merging in things already merged. This should be done with a proper PR that merges in the specific changes. Because 57 files should not have been changed with this PR

mschwamb:
I would recommend closing this PR and starting from a clean branch of main and manually coping in the bits you've edited for Herget. Because this is editing things that have already been updated and runs the risk of reverting changes we're not aware aren't fixed in this version

awilson110:
> I would recommend closing this PR and starting from a clean branch of main and manually coping in the bits you've edited for Herget. Because this is editing things that have already been updated and runs the risk of reverting changes we're not aware aren't fixed in this version

Good point. I was just worried that this is the second time I've made a new branch for this reason and I didn't want to make loads of unnecessary branches, but I do agree this is probably the best way to handle being so many commits behind main

PR #558, opened 2026-09-08
Make the layup verbs callable from the package namespace

matthewholman:
Appendix B of the methods paper documents `import layup` followed by `layup.orbitfit(...)`, but `__init__.py` exported only `__version__`, so that raises `AttributeError`. The six functions exist only as `layup.<module>.<function>`.

The obvious fix does not work. `layup.orbitfit` already means the **module**: `test_nongrav_auto`, `test_fit_outcome`, `test_iod_picker` and `test_iod_auto` reach `do_fit`, `create_empty_result` and `_select_nongrav_auto` through it, and 92 places import siblings as `from layup.convert import convert`. Rebinding the name to the function fails 11 tests in `test_nongrav_auto` alone.

A lazy PEP 562 `__getattr__` does not work either: importing any verb imports its siblings — `layup.orbitfit` pulls in `layup.convert` — which binds the submodule under the same name, so the hook is never consulted for it.

So the verb modules are made callable, and answer to both:

```python
layup.orbitfit(...)      # runs the fit
layup.orbitfit.do_fit    # still the module, unchanged
```

Nothing is imported until a verb is asked for, so `import layup` stays at about 0.05 s against the 2 s that binding all six costs. That is what keeps `layup --help` and `--version` at 0.06 s, and the new test pins it.

Full suite: **558 passed, 1 skipped**. `black` 25.1.0 clean.

The alternative is one sentence in Appendix B instead — `from layup.orbitfit import orbitfit`, which is what the code already does in 92 places. Happy to go that way if reviewers prefer it to a callable module.

### DIFF
--- a/src/layup/__init__.py
+++ b/src/layup/__init__.py
@@ -1,3 +1,5 @@
+import sys as _sys
+import types as _types
 import warnings as _warnings
 
 # Silence a cosmetic AstropyDeprecationWarning emitted at import time by sbpy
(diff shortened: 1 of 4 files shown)


kjnapier (approved):
Looks good to me

PR #560, opened 2026-09-09
Stop the comet fallback guard permitting out-of-bounds ASSIST calls

matthewholman:
`ASSIST_TIMEFRAME_MIN_MJD` was `-163545`, which is **1411-02-09**. layup ships `linux_p1550p2650.440`, which begins in **1550**. So a backward integration kept calling ASSIST for 139 years outside the ephemeris it was given, instead of falling back to the pure REBOUND path.

**That range is reached routinely.** `comet.py` integrates backwards to a barycentric distance of 250 au, and a near-parabolic comet takes of order a thousand years to get there — so an integration starting from a modern epoch lands in the 1400s as a matter of course.

**And it fails silently.** ASSIST's out-of-bounds interpolation leaves `current_state` stale rather than raising, so the integration continues against frozen perturber positions and returns plausible numbers.

| | before | after | ephemeris |
|---|---:|---:|---:|
| lower | −163545 (1411) | **−112815 (1550)** | 1550 |
| upper | 236455 (2506) | 236455 (2506) | 2650 |

The upper bound is left alone: it is conservative against the file's 288952 and therefore safe, though equally not derived from the file.

Both should come from `assist_ephem_time_bounds()` rather than being hardcoded. That entry point is in ASSIST `main` but **not in any release** — the pinned `assist 1.2.3` exports neither the Python attribute nor the C symbol — so the constants stand until a release carries it. The comment now says so, and this PR is the interim fix rather than the intended one.

The test asserts the guard lies inside the shipped ephemeris in both directions and needs no ephemeris file to run. `tests/layup/test_comet.py`: 17 passed. `black` clean.

Found while checking a sentence in the methods paper that claimed the code detects the ephemeris file's range — it doesn't, and after this it at least agrees with the file it ships.

### DIFF
--- a/src/layup/comet.py
+++ b/src/layup/comet.py
@@ -16,9 +16,24 @@ from layup.utilities.data_processing_utilities import (
     resolve_num_workers,
 )
 
-# These are the maximum and minimum dates that the ASSIST ephemeris file allows for
+# Bounds of the ASSIST planetary ephemeris, outside which the integration falls
+# back to a pure REBOUND simulation. layup ships linux_p1550p2650.440, which
+# covers 1550-01-01 to 2650-01-01, i.e. MJD -112815 to 288952.
+#
+# The lower bound previously read -163545 (1411-02-09), 139 years EARLIER than the
+# file begins, so a backward integration kept calling ASSIST for 139 years outside
+# its coverage. That matters here: a near-parabolic comet takes of order a
+# thousand years to reach the 250 au reference distance, so backward integrations
+# reach the 1400s routinely. ASSIST's out-of-bounds interpolation leaves
+# current_state stale rather than raising, so the result is plausible numbers
+# rather than a failure.
+#
+# These should come from assist_ephem_time_bounds() rather than being hardcoded.
+# That entry point is in ASSIST main but not in any release -- the pinned
+# assist 1.2.3 exports neither the Python attribute nor the C symbol -- so the
+# constants stand until a release carries it.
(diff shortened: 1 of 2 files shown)


matthewholman:
Closing this: **the premise is wrong and the change makes the guard worse.**

`ASSIST_TIMEFRAME_MIN_MJD` is not an MJD. `generate_simulations` sets `sim.t = epoch - ephem.jd_ref`, and `ephem.jd_ref` is 2451545.0, so `sim.t` is days from J2000. The constants are compared against `sim.t`, so they are J2000 days too — the `_MJD` suffix is a misnomer, and I read the value as an MJD without checking the operand.

| constant | read as MJD | actual, as J2000 days | ephemeris |
|---|---|---|---|
| `MIN` = −163545 | 1411 | **1552.2** | 1550 |
| `MIN` = −112815 (this PR) | 1550 | **1691.1** | 1550 |
| `MAX` = 236455 | 2506 | **2647.4** | 2650 |

So the original values were already correct, and deliberately just inside the ephemeris at both ends. This PR moved the lower guard from 1552 to 1691 and discarded 139 years of legitimate ASSIST coverage.

CI caught it: `test_comet_output` failed because two demo comets crossed to the pure-REBOUND path ~140 years early, shifting `inv_ao` by 1.66 and 5.00 in units of 1e-6/au — where the CODE catalogue's median uncertainty is 2.2. I initially read that failure as evidence the original bug was real. It was the opposite.

No change is needed to the guard. What is worth doing is renaming the constants so the next reader is not misled the same way; filing that separately.

PR #561, opened 2026-09-09
Orphan process bug fix

### LINKED ISSUE #523: concurrent.futures.ProcessPoolExecutor doesnt close all subprocesses when ctrl+C (KeyboardInterrupt) is used on Linux
**Bug report**

The functions that use concurrent.futures.ProcessPoolExecutor have a bug when keyboard interrupt is used.

When a keyboard interrupt is used, the code (main process) errors out; however, subprocesses created from ProcessPoolExecutor do not error out and continue to run (called orphan processes, https://en.wikipedia.org/wiki/Orphan_process):
https://github.com/Smithsonian/layup/blob/6e8d1978daa5727df11f2152272ccd01f55d3af4/src/layup/utilities/data_processing_utilities.py#L239-L247

In edge cases where orbitfits conversion hangs and doesnt converge to an orbital solution, a user using ctrl+C (KeyboardInterrupt),might not realise that subprocesses are still running taking up computational resources.

A possible solution might be to use multiprocessing.Pool instead of ProcessPoolExecutor for parallel code so that we can terminate subprocesses on a keyboard interrupt, e.g https://noswap.com/blog/python-multiprocessing-keyboardinterrupt :
```
except KeyboardInterrupt:
        print "Caught KeyboardInterrupt, terminating workers"
        pool.terminate()
        pool.join()
```

https://github.com/Smithsonian/layup/blob/6e8d1978daa5727df11f2152272ccd01f55d3af4/src/layup/utilities/data_processing_utilities.py#L167-L203

https://github.com/Smithsonian/layup/blob/6e8d1978daa5727df11f2152272ccd01f55d3af4/src/layup/utilities/data_processing_utilities.py#L206-L247

**Before submitting**
Please check the following:

- [x] I have described the situation in which the bug arose, including what code was executed, information about my environment, and any applicable data others will need to reproduce the problem.
- [x] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a description of what I expected instead.
- [x] If I have a solution in mind, I have provided an explanation and/or pseudocode and/or task list.

Little-Ryugu:
Fixes #523  .

Replaced `concurrent.futures.ProcessPoolExecutor` with `multiprocessing.Pool`

The logic these packages use is still the same. The only difference is that `multiprocessing.Pool` allows running processes to be terminated if specified `pool.terminate()`

`_init_worker()` prevents the `KeyboardInterrupt` from propagating to the subprocesses. A `try` and `except` statement is implemented to catch `KeyboardInterrupt` to terminate the subprocesses from the main process to avoid bugs. 

### DIFF
--- a/src/layup/utilities/data_processing_utilities.py
+++ b/src/layup/utilities/data_processing_utilities.py
@@ -182,7 +182,7 @@ def _apply_with_kwargs(func, data, kwargs):
     return func(data, **kwargs)
 
 
-def _run_pool(run_function, n_workers):
+def _run_pool(tuple_task_list, n_workers):
     """
     General function to run multi_processing.Pool .
     This function spawns (_MP_CONTEXT) a pool of n_workers and
@@ -192,7 +192,7 @@ def _run_pool(run_function, n_workers):
     Parameters
     -----------
 
-    run_function : list of (func, chunked_data, kwargs) tuples
+    tuple_task_list : list of (func, chunked_data, kwargs) tuples
         Arguments used for the _apply_with_kwargs function. Each tuple in list is for a
         parallel core/worker
     n_workers : int
@@ -205,8 +205,9 @@ def _run_pool(run_function, n_workers):
     """
     with _MP_CONTEXT.Pool(processes=n_workers, initializer=_init_worker) as pool:  # parallel across n_workers
         try:
-            # run the function with supplied kwargs
-            results = pool.starmap(_apply_with_kwargs, run_function)
+            # starmaps takes a function and an iterable parameter (in this casue the list)
+            # and iterates through all the chunked data. (each chunk is given a core)
+            results = pool.starmap(_apply_with_kwargs, tuple_task_list)
         except KeyboardInterrupt:
             # if keyboard interupt stop all processes
             pool.terminate()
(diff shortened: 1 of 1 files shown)


matthewholman:
Tested this and it does close #523: with a normal SIGINT handler the parent catches the interrupt promptly, `pool.terminate()` runs, and no children are left behind. Two things I'd change before merging.

**`chunksize=1` on the starmap.** `ProcessPoolExecutor.submit` created one work item per task; `Pool.starmap` batches, and the default is large:

| tasks | workers | default chunksize |
|---:|---:|---:|
| 1,000 | 8 | 32 |
| 100,000 | 8 | 3,125 |

`process_data_by_id` makes one task per object, and per-object cost is very uneven — median about 0.4 s, with individual objects running to hours. Batched that way, one worker can draw several slow objects and block while the rest idle. `pool.starmap(_apply_func_with_kwargs, tuple_task_list, chunksize=1)` restores the old granularity.

**`raise` rather than `sys.exit` in `_run_pool`.** Since #558 the verbs are callable as an API (`layup.orbitfit(...)`), so a library calling `sys.exit` on Ctrl-C takes the decision away from the caller. Re-raising after `terminate()` leaves the CLI free to catch `KeyboardInterrupt` and print a clean message.

Neither affects the mechanism, which is right.

matthewholman (approved):
Look good!

PR #567, opened 2026-09-10
--separate-flagged: keep the full row, so a converged fit keeps its orbit

matthewholman:
## What changes

With `-sf/--separate-flagged`, the flagged output carries the same columns as the
main output. It previously carried three: the identifier, `method` and `flag`.

## Why

Flags 2, 6 and 9 all mark fits that **converged** — reduced chi-square above
threshold, degenerate covariance, implausible excess speed. Three of the four
converged flag values therefore had their orbit discarded, and recovering it
meant re-running the fit without `-sf`. A reduced chi-square of 2.5 is a
judgement for the caller, not a failure.

The reduction was right when it was written: for flag 1 the state columns hold
the *initial value*, so a full-width file would have emitted a seed that looked
like a fit. #527 removed that objection — every row now carries `accepted` and
`converged`, so the flagged output says what each row is.

## Tests

The split path had no coverage; nothing set `separate_flagged = True`. The
partition is factored into `split_accepted_flagged()` and tested on a synthetic
results array, so it runs without a fit and cannot quietly stop being exercised.
Reverting the change fails it on the dtype comparison. Help text updated.

Answers the question raised in #493.

*Drafted with Claude; tests run locally.*


tests/layup/test_orbit_fit.py
```diff
@@ -378,3 +378,107 @@ def __init__(self, g=None):
             cli_args=FakeCliArgs(),
         )
         assert "Unknown engine" in str(e.value)
+
+
+# --------------------------------------------------------------------------
+# -sf / --separate-flagged
+#
+# The split path had no coverage. It matters because a flagged row is not
```
mschwamb: why doesn't it have coverage? It seems like it does with the unit test.

tests/layup/test_orbit_fit.py
```diff
@@ -378,3 +378,107 @@ def __init__(self, g=None):
...
+# necessarily a failed one: flags 2, 6 and 9 all mark fits that converged
```
mschwamb: If flags ever changed in the future, someone would need to remember to change this. Should there be one place referenced that defines the flag so we don't have to worry about maintaining docstrings if the ordering ever changes?

tests/layup/test_orbit_fit.py
```diff
@@ -378,3 +378,107 @@ def __init__(self, g=None):
...
+        arr[i]["accepted"] = 1 if f == 0 else 0
+        arr[i]["converged"] = 1 if f in (0, 2, 6, 9) else 0
+    return arr
+
+
+def test_split_accepted_flagged_keeps_every_column():
+    """The flagged half keeps the full row, orbit included.
+
+    Flags 2, 6 and 9 converged; emitting only the identifier would force a
+    second run without --separate-flagged to recover their orbits.
```
mschwamb: This reads more like it's telling why the change is made to include the other columns. Maybe this doesn't go in the docstrings for a unit test.

src/layup/orbitfit.py
```diff
@@ -2354,23 +2354,36 @@ def _emit(arr, path):
...
 
 
+def split_accepted_flagged(fit_orbits):
+    """Partition fit results into the accepted rows and the flagged ones.
+
+    Both halves keep every column. A flagged row is not necessarily a failed
+    one: flags 2, 6 and 9 all mark a fit that converged -- reduced chi-square
+    above threshold, a degenerate covariance, an implausible excess speed --
+    and each is a judgement the caller may want to make differently. Returning
+    the identifier alone would force a second run without ``--separate-flagged``
```
mschwamb: I would remove "Returning  the identifier alone would force a second run without ``--separate-flagged``
to recover the orbit. " - It's explaining what changed not what the function is doing. Useful for maintaining this later/JOSS code review.

mschwamb: Would it be worth a reference here that all of the flags are defined in constants.py?

mschwamb (changes requested):
It looks fine codewise. A few suggestions to consider in the docstrings for long term maintainability.

tests/layup/test_orbit_fit.py
matthewholman: It didn't before this PR -- nothing in the suite set `separate_flagged = True`. The comment was describing the state I found, which reads wrong now that the PR adds the test. Removed.

*Drafted with Claude Opus 5.*

tests/layup/test_orbit_fit.py
matthewholman: There is one: `CONVERGED_FLAGS` in `constants.py`. Pointing at it turned up that it was wrong -- #497 added `FLAG_IMPLAUSIBLE_ORBIT` as a converged outcome and never added it there, so the constant contradicted the flag's own comment in the same file.

That had a consequence: through `FitOutcome.from_flag`, a flag-9 row came back byte-identical to "did not converge" -- `converged 0, stage 0, failed_physical 0` -- losing both that it had converged and why it was rejected. The contract test in `test_fit_outcome.py` covered eight flags and not that one.

So: the constant now includes it, with the rule written down (every flag whose comment begins "converged" belongs in it), `from_flag` maps it to `STAGE_COMPLETE` and sets `failed_physical`, the contract test gains its missing row, and this test derives its flags from `CONVERGED_FLAGS` rather than writing them out. Your comment found a live defect.

*Drafted with Claude Opus 5.*

tests/layup/test_orbit_fit.py
matthewholman: Agreed -- that was PR rationale in a test docstring. Cut.

*Drafted with Claude Opus 5.*

src/layup/orbitfit.py
matthewholman: Removed. Same point as above: it explained what changed rather than what the function does.

*Drafted with Claude Opus 5.*

matthewholman: Added -- the docstring now points at `constants.py` and `CONVERGED_FLAGS` instead of listing the values, and the mask uses `FLAG_CONVERGED` rather than a bare 0.

*Drafted with Claude Opus 5.*

mschwamb (approved):

PR #570, opened 2026-09-10
Orphan process bug fixes: adding SIGTERM handler

### LINKED ISSUE #562: Worker pools are orphaned when layup is stopped in the background
#523 and PR #561 cover the interactive case: Ctrl-C reaches the parent, which terminates the pool. A layup run started in the background is not covered, and this is the case that leaves processes running for days.

A shell backgrounding a job sets `SIGINT` to `SIG_IGN`, and the parent inherits it. The interrupt never arrives, so the `except KeyboardInterrupt` teardown never runs. Such a job is stopped with `SIGTERM` instead, whose default action kills the parent outright without unwinding, leaving the workers with `PPID = 1`.

Measured on a backgrounded parent with four workers, stopped by `SIGTERM`:

| | orphans left |
|---|---:|
| as in #561 | 4 |
| with a SIGTERM handler | 0 |

Six such processes were found on one machine recently, `PPID = 1` and five days old, in an environment nobody was still using.

Suggested fix, in `_run_pool`: install a SIGTERM handler that raises, so the existing teardown runs, and restore the previous handler in a `finally` so it does not leak into a caller's process.

```python
def _terminate(signum, frame):
    raise KeyboardInterrupt

_prev_term = signal.signal(signal.SIGTERM, _terminate)
try:
    ...  # existing pool block
finally:
    signal.signal(signal.SIGTERM, _prev_term)
```

Two caveats: `signal.signal` only works on the main thread, so this should be skipped when layup is driven from a worker thread; and `SIGKILL` cannot be handled by anything.

Little-Ryugu:
Fixes #562  .

Added a SIGTERM (siganl terminate) handler so _run_pool properly terminates subprocesses from keyboardInterrupt and SIGTERM cases now.

I did this using the suggestions from #562. I also added guards for if a user is not using the main thread. 

If a user is not using the main thread, then a logger.warning will warn a user that orphan processes can occur and also includes a command to check for said processes.

Also cleaned up my previous comments and docstrings with better descriptions and spelling fixes. 

### DIFF
--- a/src/layup/utilities/data_processing_utilities.py
+++ b/src/layup/utilities/data_processing_utilities.py
@@ -180,8 +180,8 @@ def _init_worker():
 
 def _terminate(signum, frame):
     """
-    Handler for SIGTERM while using multiprocessing.
-    Any SIGTERM (signal terminate) will be treated as a
+    Handler for SIGTERM and SIGHUP while using multiprocessing.
+    Any SIGTERM (signal terminate) or SIGHUP will be treated as a
     KeyboardInterrupt in _run_pool()
     """
     raise KeyboardInterrupt
@@ -225,8 +225,9 @@ def _run_pool(tuple_task_list, n_workers):
     on_main_thread = threading.current_thread() is threading.main_thread()
     _prev_term = None
     if on_main_thread:
-        # _prev_term stores the previous way SIGTERM was handled.
-        _prev_term = signal.signal(signal.SIGTERM, _terminate)
+        # _prev_term stores the previous way SIGTERM and SIGHUP was handled.
+        _prev_term_SIGTERM = signal.signal(signal.SIGTERM, _terminate)
+        _prev_term_SIGHUP = signal.signal(signal.SIGHUP, _terminate)
     else:
         logger.warning(
             "layup is being called not on the main thread, SIGTERM will not be handled. Therefore orbitfitting may leave orphaned processes if terminated. Use: \n`ps -eo pid,ppid,args -ww | awk '$2==1' | grep -i \"multiprocessing.spawn\\|layup\"`\n to check for layup orphan processes, and `kill <pid>` to remove them."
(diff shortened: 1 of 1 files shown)


matthewholman:
Nice approach — raising `KeyboardInterrupt` from the handler reuses the cleanup path that already works, and the non-main-thread guard is worth having, since `signal.signal` raises off the main thread.

Three things:

**1. `_prev_term` can be `None`.** `signal.signal` returns `None` when the previous handler wasn't installed from Python, and `signal.signal(SIGTERM, None)` raises `TypeError`. It's in the `finally`, so that turns a clean exit into a crash. `if _prev_term is not None:` covers it.

**2. No test**, though the checklist ticks the box. Worth one, and harder than it looks: a test that backgrounds the parent inherits `SIGINT = SIG_IGN` and passes for the wrong reason. Set `signal.default_int_handler` in the child.

**3. Is SIGHUP in scope?** A closing terminal sends SIGHUP, not SIGTERM, and #562 is about background stops.

*Drafted with Claude Opus 5.*

Little-Ryugu:
I've added 1 and 3 there. Working on making a unit test now. I'll see what i can do

matthewholman:
The approach looks right, and the `finally` restore is the correct shape. Four things:

1. **The two `signal.signal` calls sit outside the `try`.** If the second raises, the first is never undone. That happens on Windows, where `signal.SIGHUP` does not exist: SIGTERM is left pointing at `_terminate` for the life of the process. Moving both inside the `try` fixes it.

2. **`pyproject.toml` claims `Operating System :: OS Independent`, but CI runs only macOS and Ubuntu**, so the above cannot be caught. Either guard with `getattr(signal, "SIGHUP", None)` or drop the claim.

3. **`if _prev_term_SIGTERM is not None`** skips the restore when `signal.signal` returns `None`, which means the handler was not installed from Python. The ordinary case returns `SIG_DFL`, so this only bites the inherited-handler case -- but that is the one where restoring matters.

4. `_prev_term = None` is dead, and the `KeyboardInterrupt` block ignores further SIGTERM during teardown but not SIGHUP.

Could `_terminate` and the restore path get a test? Both are exercisable without starting a pool.

*Drafted with Claude Opus 5.*

PR #571, opened 2026-09-11
Zero-initialise FitResult::cov so a flagged fit is deterministic

### LINKED ISSUE #547: FitResult.cov is left uninitialised when a Cartesian fit returns flag != 0
### Summary

`orbit_fit` populates `FitResult.cov` only when the fit converges cleanly, but `cov` has no
initialiser, so **every fit that returns `flag != 0` hands back uninitialised stack memory as
its covariance**. `bk_fit` already does the right thing (`result.cov.fill(0.0)`), so the
Cartesian path is the outlier.

This is easy to miss because `flag = 2` is not an error — it is a converged fit whose reduced
χ² exceeded `thresh = 10.0` — so callers reasonably treat the result as usable and read `cov`.

### Where

- `src/lib/orbit_fit/fit_result.cpp:21` — `std::array<double, 36> cov;`, no default member
  initialiser (same for `state` and `root`).
- `src/lib/orbit_fit/orbit_fit.cpp:1180` — `FitResult result;` is default-initialised, so
  `cov` is indeterminate.
- `src/lib/orbit_fit/orbit_fit.cpp:1545` — `if (flag == 0) { ... result.cov[...] = cov(i, j); }`
  is the only place it is written.

Reachable with `flag = 1` (no convergence, `orbit_fit.cpp:1180`), `flag = 2`
(`orbit_fit.cpp:1315`, reduced χ² > 10) and `flag = 6` (`orbit_fit.cpp:1348`, weakly
constrained / non-positive variance).

Contrast `src/lib/orbit_fit/bk_fit.cpp:42`, which zeroes it up front:

```cpp
result.cov.fill(0.0);
```

### Observed

Same object, same input, same build, three separate processes — a `flag = 2` fit on
(20) Massalia:

```
### numbered__00020: flag=2 csq=7.73354e+07 ndof=19978 red=3871.03
  cov finite: False   any NaN: True
  diagonal: [0.0e+000 4.9407e-324 6.9527e-310 0.0e+000 0.0e+000  nan]   run 1
  diagonal: [0.0e+000 4.9407e-324 6.9526e-310 0.0e+000 0.0e+000  nan]   run 2
  diagonal: [0.0e+000 4.9410e-324 6.9530e-310 0.0e+000 0.0e+000  nan]   run 3
  symmetric: True / False / True
```

`6.95e-310` is a pointer value reinterpreted as a double and it **moves with ASLR between
runs**; `4.94e-324` is the smallest subnormal. The array is not merely wrong, it differs every
execution, so anything downstream of it is irreproducible.

`csq`, `ndof` and `state` are identical across the three runs — the fit itself is
deterministic. Only `cov` varies.

### Why it bites

Any consumer that reads `fit.cov` after a `flag != 0` result. Concretely, an OrbFit-style
outlier-rejection loop (Carpino, Milani & Chesley 2003) computes a per-observation χ² on the
residual covariance `Γ_res = Γ_obs ∓ B Γ_x Bᵀ`, taking `Γ_x` from `fit.cov`. With `cov`
indeterminate, every reject/recover decision taken while the fit sits at `flag = 2` is made on
garbage — and a NaN silently defeats the usual `det <= 0` sanity check, because every
comparison against NaN is false.

Measured over a full 1.51 M-object MPC catalogue fit (455 run logs, 1,506,746 object blocks):
**5,562 objects (0.37%) took at least one rejection iteration at `flag != 0`**, and 5,311 of
them moved their active observation set on it. Guarding the covariance changes the outcome for
836 of the 1,366 objects that ended at `flag = 2` — they converge at median reduced χ² 0.25
instead, and agree with MPCORB at median rel|da| 1.5e-7.

### Suggested fix

One line, matching `bk_fit`:

```cpp
FitResult result;
result.cov.fill(0.0);   // and result.state, result.root
```

or give the members default initialisers in `fit_result.cpp` so every `FitResult` is
zero-initialised regardless of construction site.

Zeroing is safe for the rejection use above (`Γ_res` degrades to `Γ_obs`, i.e. no projection)
and is unambiguous to test for, whereas indeterminate values are not. Worth documenting on the
pybind11 binding too: `cov` is only meaningful when `flag == 0`.

### Version

Reproduced on `main` @ `b6d7edd`, and on the 2026-07 run build (`eea7d3b` + #411/#416/#412).
Unaffected by #536.

<sub>Surfaced while re-fitting the objects that did not reach `flag=0` in the full-MPC
catalogue run.</sub>

matthewholman:
Closes #547.

## The defect

`orbit_fit` fills `FitResult::cov` only inside `if (flag == 0)`, and `run_sequential_update` returns early at the non-positive-definite prior check without reaching its fill either. `cov` had no initialiser, so those paths returned whatever was on the stack.

Measured on the flag-7 path before the fix, over repeated identical calls in one process:

| | |
|---|---|
| non-zero entries | **31 to 32 of 36** |
| values seen | `1.9e-313` denormals on one run; `6.0e-154` and `1.78e+21` on another |
| between identical calls | **the count differed** |

So the same object could score differently on two machines, and a consumer could not tell an absent covariance from a real one.

## The fix

A default member initialiser on the declaration:

```diff
-        std::array<double, 36> cov;    // Covariance matrix
+        std::array<double, 36> cov{};  // Covariance matrix
```

On the declaration rather than after each `FitResult result;`. `bk_fit.cpp:42` already does the per-site fill correctly, and that is precisely why the other sites were missed — a per-site fill only covers the sites someone remembered. On the declaration it is a property of the type, so every construction site is covered, including ones added later (`orbit_fit.cpp` has two, and `gauss.h` returns a `vector<FitResult>`).

**Zero rather than NaN.** NaN is more informative in principle — it propagates and cannot be mistaken for a real covariance — but zero matches the convention `bk_fit` established, and the requirement here is determinism. What says whether a covariance is meaningful is the `flag`, or the `accepted` column. NaN would also change behaviour for consumers that currently treat a zeroed `cov` as "absent".

**Converged fits are untouched**: both fill sites are downstream of this.

## Tests

`tests/layup/test_fit_result_cov_init.py` — three cases against `run_sequential_update` with a non-positive-definite prior, which takes the flag-7 early return at `orbit_fit.cpp:1643`. They assert the covariance is zero, is *the same* across repeated identical calls, and is finite. The middle one is the point, since the symptom was non-determinism rather than merely non-zero values.

⚠️ **The tests exercise the C++ path deliberately, and the file says why.** A Python-constructed `FitResult()` is value-initialised by pybind11's `py::init<>()` and reads as all-zero **with or without** the fix — my first version of this test did that and passed against the unfixed build, proving nothing.

Verified by reverting the one-line change, rebuilding, and re-running: 2 of the 3 fail, with the garbage values above. (The third, the finiteness check, is a weaker assertion and can pass by luck.)

Full suite locally: **582 passed, 1 skipped** in 13m36s, the skip pre-existing.

## Why it matters beyond hygiene

The MPC full-catalogue harness works around this by zeroing the covariance whenever the fit is flagged. That guard is what makes its Carpino et al. (2003) outlier statistic — which normalises by the post-fit residual covariance `Γ_ν = W⁻¹ − BΓBᵀ` — silently collapse to plain `W⁻¹` on flagged fits, so the rejection criterion changes definition mid-loop as the flag flips. Once this lands, that guard can be removed.

### DIFF
--- a/src/lib/orbit_fit/fit_result.cpp
+++ b/src/lib/orbit_fit/fit_result.cpp
@@ -18,7 +18,13 @@ namespace orbit_fit
         double epoch;                  // Epoch
         double root;                   // Root value (gauss)
         std::array<double, 6> state;   // State vector
-        std::array<double, 36> cov;    // Covariance matrix
+        // Zero-initialised: `orbit_fit` fills `cov` only when the fit converges
+        // cleanly, so without this every `flag != 0` result would hand back
+        // uninitialised stack memory as its covariance -- differing between
+        // processes, so the same object could score differently on two machines.
+        // Done at the declaration rather than after each `FitResult result;`
+        // (the `bk_fit.cpp` pattern) so every construction site is covered.
+        std::array<double, 36> cov{};  // Covariance matrix
         int niter;                     // Number of iterations
         std::string method;            // Method used for fitting
         int flag;                      // Flag indicating the success of the fit
--- a/tests/layup/test_fit_result_cov_init.py
+++ b/tests/layup/test_fit_result_cov_init.py
@@ -1,0 +1,72 @@
+"""A flagged fit must return a deterministic covariance (#547).
+
+``orbit_fit`` fills ``FitResult::cov`` only inside ``if (flag == 0)``, and
+``run_sequential_update`` returns early at the non-positive-definite prior check
+without reaching its fill either. ``FitResult::cov`` had no initialiser, so
+those paths handed back whatever was on the stack: measured on the flag-7 path
+before the fix, 31 to 32 of the 36 entries were non-zero denormals around
+1.9e-313, and **the count differed between identical calls** -- so the same
+object could score differently on two machines, and a consumer could not tell an
+absent covariance from a real one.
+
+``bk_fit`` already did the right thing (``result.cov.fill(0.0)``), but a
+per-site fill only covers the sites someone remembered. The fix is a default
+member initialiser on the declaration in ``fit_result.cpp``, which covers every
+construction site including ones added later.
+
+Zero rather than NaN, matching the convention ``bk_fit`` established. The flag,
+not the covariance, is what says whether a fit is usable.
+
+⚠️ These must run against the C++ path. A Python-constructed ``FitResult()`` is
+value-initialised by pybind11's ``py::init<>()`` and reads as all-zero **with or
+without** the fix, so a test built on one passes either way and proves nothing.
+"""
+
+import numpy as np
+import pytest
+
+from layup.routines import FitResult, get_ephem, run_sequential_update
+
+from _bk_guards import EPHEM_CACHE, requires_ephem
+
+pytestmark = requires_ephem
+
+CACHE = str(EPHEM_CACHE)
+STATE = [40.0, 10.0, 5.0, -8e-4, 9e-4, 1e-4]
+EPOCH = 2460000.5
+
+# Prior covariance not positive-definite, so the LLT check fails and
+# run_sequential_update returns before it would populate `cov`. The docstring on
+# the binding states this contract: "Fails (flag != 0) if the prior covariance is
+# not positive-definite."
+FLAG_PRIOR_NOT_POSITIVE_DEFINITE = 7
+
+
+def _degenerate_prior():
+    prior = FitResult()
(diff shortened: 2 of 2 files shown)


kjnapier (approved):
Looks okay to me

Issue #543, opened 2026-09-04
Observations before 1849 fail with a raw SPICE error (de440s.bsp coverage)


matthewholman:
`LayupObservatory` loads **`de440s.bsp`**, the short variant of DE440, which covers only **1849 Dec 25 to 2150 Jan 21**. Any observation earlier than that fails when the observer position is computed:

```
SPICE(SPKINSUFFDATA) -- Insufficient ephemeris data has been loaded to compute the state
of 399 (EARTH) relative to 0 (SOLAR SYSTEM BARYCENTER) at the ephemeris epoch
1845 DEC 15 00:42:02.591.
```

Hit on (5) Astraea, discovered 1845 — its earliest observations fall ten days short of four years before the kernel begins.

The mismatch is the awkward part: ASSIST integrates against `linux_p1550p2650.440`, which spans 1550–2650, so the *dynamics* reach three centuries further back than the observer positions that feed them. A user fitting a 19th-century arc gets a raw SPICE exception rather than anything naming the cause.

Two things would help: an error that names the kernel and its span, and an option to load the full `de440.bsp` (1550–2650) for historical arcs. Presumably `de440s` was chosen for download size — roughly 32 MB against 114 MB — which is a reasonable default, just not a silent one.

mschwamb:
Since we have the option to run with a config file, the  user can swap to planet ephemeris variable to de440.bsp, and I think everything would work as is. It's not an option used very often (on the CLI you can optional specify the config file and the API is setup to pass a config variable). So if the formats of the files are the same, we have the capability to this already I think. 

layup init creates the config file that can be edited. So I think the user could just edit these lines in the config file


planet_ephemeris = de440s.bsp
planet_ephemeris_url = https://naif.jpl.nasa.gov/pub/naif/generic_kernels/spk/planets/de440s.bsp 

Then as the user ran when it checks for the files it will automatically download the bigger file. So maybe the solution here is just a check on timespan of observations on readin and whether the user should rerun with the config file to use the larger file?

matthewholman:
Confirmed -- same format, both are SPK kernels, so the swap works as is. Two keys rather than one: `planet_ephemeris` **and** `planet_ephemeris_url`, or the fetch pulls the short file under the long file's name. The docstring at `layup_configs.py:13` already spells this out, including that Earth orientation before 1962 falls back to the analytical IAU model.

[...]

PR #572, opened 2026-09-11
Name the ASSIST coverage guards for the units they use

### LINKED ISSUE #563: comet.py's ASSIST_TIMEFRAME_*_MJD constants are J2000 days, not MJD
`comet.py` defines

```python
ASSIST_TIMEFRAME_MAX_MJD = 236455
ASSIST_TIMEFRAME_MIN_MJD = -163545
```

but these are not MJDs. They are compared against `sim.t`, and `generate_simulations` sets `sim.t = epoch - ephem.jd_ref` with `ephem.jd_ref = 2451545.0`, so both are **days from J2000**.

Read as MJD they look like 1411 and 2506, which is wrong at both ends and alarming at the lower one. Read correctly they are **1552.2 and 2647.4**, sitting just inside the shipped `linux_p1550p2650.440`, which is what they should be.

The values are right; only the names are wrong. Suggested:

```python
# Bounds of the shipped ASSIST planetary ephemeris (linux_p1550p2650.440,
# covering 1550-2650), expressed in the simulation clock: sim.t = JD - 2451545.0.
# NOT MJD. 1552.2 and 2647.4, deliberately just inside the file at both ends.
ASSIST_TIMEFRAME_MAX_SIMT = 236455
ASSIST_TIMEFRAME_MIN_SIMT = -163545
```

This cost a wrong PR (#560, closed) that moved the lower guard to 1691 and discarded 139 years of valid coverage before CI caught it. Worth renaming so it does not cost anyone else one.

matthewholman:
Closes #563.

## The defect

`comet.py` named these `ASSIST_TIMEFRAME_MAX_MJD` / `_MIN_MJD`, but they are not MJDs. They are compared against `sim.t`, which `generate_simulations` sets to `epoch - ephem.jd_ref` with `jd_ref = 2451545.0` — so they are **TDB days from J2000**:

| constant | `sim.t` | JD | date |
|---|---:|---:|---|
| MAX | +236455 | 2688000.0 | 2647-05-24 |
| MIN | −163545 | 2288000.0 | 1552-03-25 |

which is `linux_p1550p2650.440` (1550–2650) with a margin at each end.

**The values were always right; only the names said otherwise.** A negative "MJD" is the tell. The risk was that someone would eventually reconcile the name with the number in the wrong direction and narrow the guard by three millennia.

## The change

Renamed to `ASSIST_TIMEFRAME_MAX_J2000_DAYS` / `_MIN_J2000_DAYS`, with the derivation recorded beside them. The constants are used only in `comet.py`, so nothing else moves.

Three tests pin the units, in `test_comet.py`:

- the constants convert through `jd_ref` to 1552 and 2647;
- reading the same integers **as MJD** does *not* reproduce the kernel coverage — so a future "fix" toward MJD fails instead of silently narrowing the guard;
- dates just inside and just outside coverage fall on the right side of the comparison the code actually makes.

## What is NOT fixed, and why

🔴 These stay **hardcoded to the default kernel**. A run configured with a different planetary ephemeris has different coverage and these guards would then be wrong — which is not hypothetical: #543 just closed on exactly that swap, pointing `planet_ephemeris` at `de440.bsp` so that pre-1849 observations can be reduced.

The real remedy is upstream and exists: ASSIST has `assist_ephem_time_bounds()` (`assist.c:524`), which would let the bounds be queried from the loaded ephemeris. **It is not in a released `assist`** — 1.2.3 does not export the symbol (`hasattr(clibassist, "assist_ephem_time_bounds")` is `False`), and 1.2.3 is the latest on PyPI. So the swap is blocked on an assist release, not on this repo, and the warning is recorded next to the constants until then.

Full `test_comet.py`: 19 passed. `black --check` clean.

### DIFF
--- a/src/layup/comet.py
+++ b/src/layup/comet.py
@@ -16,9 +16,26 @@ from layup.utilities.data_processing_utilities import (
     resolve_num_workers,
 )
 
-# These are the maximum and minimum dates that the ASSIST ephemeris file allows for
-ASSIST_TIMEFRAME_MAX_MJD = 236455
-ASSIST_TIMEFRAME_MIN_MJD = -163545
+# Coverage limits of the ASSIST planetary ephemeris, in the units the guards
+# below actually compare against: **TDB days from J2000**, not MJD.
+# ``generate_simulations`` sets ``sim.t = epoch - ephem.jd_ref`` with
+# ``ephem.jd_ref = 2451545.0``, so ``sim.t`` is days from J2000 and these
+# constants have to be in the same units to be meaningful.
+#
+#   +236455 -> JD 2688000.0 -> 2647-05-24
+#   -163545 -> JD 2288000.0 -> 1552-03-25
+#
+# which is ``linux_p1550p2650.440`` (1550-2650) with a margin at each end.
+#
+# WARNING: these are hardcoded to the DEFAULT kernel. A run configured with a
+# different planetary ephemeris -- the ``de440.bsp`` swap that lets observations
+# before 1849 be reduced, for instance -- has different coverage, and these
+# guards would then be wrong in whichever direction the kernel is narrower.
+# ASSIST gained ``assist_ephem_time_bounds()`` upstream, which would let these
+# be queried from the loaded ephemeris instead; it is not in a released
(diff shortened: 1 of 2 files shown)


kjnapier (approved):
Looks good to me

PR #573, opened 2026-09-11
Decode the obs80 star-catalog code before the Veres sigma lookup

### LINKED ISSUE #548: astrometric_uncertainty_Veres2017 selects on catalogue names, but obs80 supplies single-char codes
### Summary

`astrometric_uncertainty_Veres2017` selects its per-station σ by comparing `catalog` against
catalogue **names** (`"Gaia1"`, `"UCAC4"`, `"PPMXL"`, …), but `Obs80Reader` supplies the
single-character MPC **code** (`"U"`, `"q"`, `"t"`, …). Every name-keyed branch is therefore
unreachable for obs80 input, and the affected stations silently fall through to their `else`
value.

`debias()` already handles both spellings — that was the #409 fix, and its comment says so
explicitly:

> ``catalog`` may be a catalog NAME (an MPC_CATALOGS key, e.g. "UCAC4", as ADES provides) or
> the single-char CODE that obs80 supplies (an MPC_CATALOGS value, e.g. "q"). bias_dict is
> keyed by the code, so map a name to its code and let a code pass through -- otherwise
> debiasing silently no-ops on all obs80 input (issue #409).

`astrometric_uncertainty_Veres2017` never got the same treatment.

### Where

- `src/layup/utilities/astrometric_uncertainty.py:87,107,117` and the `568` / `Y28` / `G83` /
  `309` branches — all compare against `MPC_CATALOGS` **keys**.
- `src/layup/utilities/file_io/Obs80Reader.py` — `"cat": slice(71, 72)`, a single character.
- `src/layup/utilities/debiasing.py:79-88` — the correct handling, for comparison.

### Reproduction

```python
from layup.utilities.astrometric_uncertainty import astrometric_uncertainty_Veres2017 as V
jd = 2458000.0
for stn, code, name in [("568", "U", "Gaia1"), ("T09", "W", "Gaia3"),
                        ("Y28", "t", "PPMXL"), ("309", "q", "UCAC4")]:
    prg = "&" if stn == "309" else None
    print(stn, V(obsCode=stn, jd_tdb=jd, catalog=code, program=prg),
               V(obsCode=stn, jd_tdb=jd, catalog=name, program=prg))
```

```
568 1.5 0.1
T09 1.5 0.1
Y28 1.5 0.3
309 1.5 0.3
```

So a Gaia-referenced observation from 568 is weighted at 1.5″ instead of 0.1″ — 15× too loose,
225× in the weight.

### Impact

Small in practice, and I would rather report that than overstate it. Measured over a 2.5%
shard sample of the MPC archive as ingested for a full-catalogue fit:

| | observations sampled | σ changes if the code is decoded | mean σ as-is | mean σ decoded |
|---|---:|---:|---:|---:|
| numbered | 13,272,316 | 3,345 (**0.025%**) | 0.7887″ | 0.7884″ |
| unnumbered | 623,714 | 1,297 (**0.208%**) | 0.5946″ | 0.5922″ |

Station 568 dominates the affected set. The reason the exposure is so low is that the two
largest catalogue classes in the archive — Gaia DR2 (`V`) and Gaia EDR3 (`X`), together 43% of
numbered and 39% of unnumbered observations — are not in `MPC_CATALOGS` at all, so they take
the generic `elif catalog: 1.0` branch either way.

It did not move any result we depend on. It is still wrong, and it will matter more as the
Gaia-referenced fraction from those specific stations grows.

### Suggested fix

Mirror `debias()` — normalise the input before the comparisons:

```python
from layup.utilities.debiasing import MPC_CATALOGS
_CODE_TO_NAME = {v: k for k, v in MPC_CATALOGS.items()}
catalog = _CODE_TO_NAME.get(catalog, catalog)   # code -> name; a name passes through
```

Two things worth deciding at the same time:

1. `MPC_CATALOGS` has no entry for Gaia DR2 (`V`), Gaia EDR3 (`X`), UCAC-5 (`Y`) or ATLAS-2
   (`Z`). For debiasing that is correct — the Farnocchia tables have no column for them and
   Gaia-referenced astrometry needs no correction — but the Veres model *does* want to
   distinguish them, so the σ path needs its own code→name map rather than borrowing the
   debiasing one.
2. `"Gaia3E"` appears in the σ branches but not in `MPC_CATALOGS`, so the two modules disagree
   on the naming already.

### Version

`main` @ `b6d7edd`.

<sub>Surfaced while reconstructing how a full-MPC catalogue fit was weighted.</sub>

matthewholman:
Closes #548.

## The defect

The branches in `astrometric_uncertainty_Veres2017` are keyed on catalog **names** (`"Gaia1"`, `"UCAC4"`, …), which is what ADES supplies. obs80 supplies the single-character MPC **code** in column 72, so for obs80 input every name-keyed test was unreachable and the observation fell through to the station's `else` value — silently, and by up to a factor of **15 in sigma, 225 in the weight**.

`debias()` already normalised both spellings; that was #409, and its comment says so explicitly. The sigma path never got the same treatment.

## Measured impact

1-in-7 sample of the MPC archive, **3,942,528 CCD observations**:

| | |
|---|---:|
| sigma changes | **1,103 (0.028%)** |

| station | n | | transition | n |
|---|---:|---|---|---:|
| 568 | 1,013 | | 1.50 → 0.50 | 817 |
| T12 | 56 | | 1.50 → 0.10 | 182 |
| T14 | 13 | | 1.50 → 0.20 | 104 |
| T09 | 11 | | | |
| 309 | 10 | | | |

Small, and worth stating as small. But every one of them is wrong today, and only seven observatory codes branch on catalog at all, so this is the whole exposure rather than a sample of it.

## The two decisions the issue left open

**1. The map is its own, not `debiasing.MPC_CATALOGS`.** That map keys the Farnocchia bias tables, which have no column for Gaia DR2, Gaia EDR3, UCAC-5 or ATLAS-2 — correctly, since Gaia-referenced astrometry needs no bias correction. But `V` and `X` are the two commonest catalogs in the modern archive, so borrowing it would leave exactly those undecodable. A test asserts they are absent from the debiasing map, so a later "simplification" that borrows it **fails** rather than silently reverting this fix.

**2. `USNOB2` is correct** — MPC code `s` is USNO-B2.0, so the existing branch was right and merely unreachable. The `# TODO Unsure if "USNOB2" is correct abbreviation!!!` is resolved rather than left standing. It is also the **largest contributor**: 817 of the 1,103 changes are 568 via codes `o` and `s`, not the Gaia cases.

## Tests

New `tests/layup/test_astrometric_uncertainty.py`, 66 cases. The central one is parameterised over every mapped catalog × every catalog-sensitive station and asserts the **two spellings agree** — the property that actually matters, since the same observation must weigh the same whether it arrived as obs80 or as ADES.

Also pinned: the case from the issue (568 + Gaia goes 1.5 → 0.1); that codes the model does *not* branch on are unperturbed; that a missing catalog still takes the unknown branch; and a guard that every catalog name appearing in the branches has a code to decode from, so the map cannot drift away from the code it serves.

**35 of the 66 fail on revert.** Full suite: 645 passed, 1 skipped (pre-existing). `black` clean.

### DIFF
--- a/src/layup/utilities/astrometric_uncertainty.py
+++ b/src/layup/utilities/astrometric_uncertainty.py
@@ -1,3 +1,30 @@
+# obs80 supplies the single-character MPC star-catalog CODE (column 72); ADES
+# supplies the NAME. The branches below are keyed on names, so a code has to be
+# decoded first or every name-keyed test is unreachable and the observation
+# falls through to the station's `else` -- silently, and by up to a factor of 15
+# (issue #548). `debias()` already normalises both spellings; this is the same
+# treatment for the sigma path.
+#
+# This map is deliberately SEPARATE from `debiasing.MPC_CATALOGS` rather than
+# derived from it. That one exists to key the Farnocchia bias tables, which have
+# no column for Gaia DR2, Gaia EDR3, UCAC-5 or ATLAS-2 -- correctly, since
+# Gaia-referenced astrometry needs no bias correction. The Veres model does
+# distinguish them, so borrowing the debiasing map would leave `V` and `X`, the
+# two most common catalogs in the modern archive, undecodable.
+#
+# Only the names the model actually branches on need an entry. Anything else
+# passes through unchanged and reaches the same generic branch it does today.
+_CODE_TO_CATALOG_NAME = {
+    "o": "USNOB1",  # USNO-B1.0
+    "s": "USNOB2",  # USNO-B2.0 -- resolves the "unsure of the abbreviation" note below
+    "q": "UCAC4",  # UCAC-4
+    "t": "PPMXL",  # PPM-XL
+    "U": "Gaia1",  # Gaia DR1
+    "V": "Gaia2",  # Gaia DR2
+    "W": "Gaia3",  # Gaia DR3
+    "X": "Gaia3E",  # Gaia EDR3
+}
+
 EARLY_703_JD = 2456658.5
 EARLY_691_JD = 2452640.5
 EARLY_644_JD = 2452883.5
@@ -27,6 +54,9 @@ def astrometric_uncertainty_Veres2017(obsCode, jd_tdb, catalog=None, program=Non
     float
         Astrometric uncertainty in arcseconds
     """
+
+    # Accept either spelling: a code becomes its name, a name passes through.
+    catalog = _CODE_TO_CATALOG_NAME.get(catalog, catalog)
 
     sigma_arcsec = 1.5
     if obsCode == "703":
@@ -102,7 +132,7 @@ def astrometric_uncertainty_Veres2017(obsCode, jd_tdb, catalog=None, program=Non
             sigma_arcsec = 1.5
     elif obsCode == "568":
         if catalog:
-            if catalog in ["USNOB1", "USNOB2"]:  # TODO Unsure if "USNOB2" is correct abbreviation!!!
+            if catalog in ["USNOB1", "USNOB2"]:  # MPC codes o and s
                 sigma_arcsec = 0.5
             elif catalog in ["Gaia1", "Gaia2", "Gaia3", "Gaia3E"]:
                 sigma_arcsec = 0.1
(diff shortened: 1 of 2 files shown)


kjnapier (approved):
Looks okay to me

PR #574, opened 2026-09-11
Thread primary_id_column_name through the sequential path

### LINKED ISSUE #515: incremental_orbitfit fails unless the id column is named provID
`_observations_for_update` does not take `primary_id_column_name` and reads
`d["provID"]` directly (`orbitfit.py:1735` and `:1747`), so the sequential-update
path only works when the id column happens to have the default name:

```python
incremental_orbitfit(data, cache_dir, prior, primary_id_column_name="ObjID")
# ValueError: no field of name provID
```

via `sequential_update` -> `_observations_for_update`. `orbitfit()` handles
`ObjID` fine, so only the incremental path is affected. The existing incremental
tests all use `provID`, which is why it does not show up.

Fix is to thread the column name through and use it.

matthewholman:
Closes #515.

## The defect

`incremental_orbitfit` takes `primary_id_column_name`, but it was never passed to `sequential_update`, which did not accept it, and `_observations_for_update` read `d["provID"]` directly. So the sequential-update path worked only when the id column happened to carry the default name and raised otherwise — on data that `orbitfit()` itself reads without complaint.

`_orbitfit` (line 1283) has threaded the name correctly all along. The sequential path was the sole exception, which is why it went unnoticed.

## The change

Nine lines: `_observations_for_update` and `sequential_update` gain the parameter, the two call sites forward it, and `incremental_orbitfit` passes its own value through. Callers that never pass the argument are unaffected — the default is still `"provID"`.

## Tests

`tests/layup/test_incremental_id_column.py` pins something stronger than "does not raise": the same observations under a different id column must give the **same state and covariance, bit for bit** (`atol=0, rtol=0`). A column name is not allowed to change the answer.

Two of the three fail on revert; the third guards that the default path is unchanged.

⚠️ Note for anyone writing a similar test: `recfunctions.rename_fields` raises *"Cannot change data-type for array of references"* on the object-dtype columns these readers produce, so the helper builds the renamed array field by field. My first version of the test failed for that reason rather than for the reason under test.

Full suite: 582 passed, 1 skipped (pre-existing). `black` clean.

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -1829,7 +1829,9 @@ def orbitfit(
     return fitted
 
 
-def _observations_for_update(data, cache_dir, weight_data=False, bias_dict=None):
+def _observations_for_update(
+    data, cache_dir, weight_data=False, bias_dict=None, primary_id_column_name="provID"
+):
     """Augment one object's observations with the observer barycentric state and
     build the C++ ``Observation`` list, mirroring ``orbitfit()``'s preprocessing.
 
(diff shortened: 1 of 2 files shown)


kjnapier (approved):
Looks fine to me

PR #575, opened 2026-09-11
Score IOD candidates in one pass instead of one per observation

### LINKED ISSUE #555: IOD residual prefilter re-integrates from scratch for every observation
`filter_candidates_by_residual` calls `_predict_rho_hat` once per (candidate, observation), and each call builds a fresh `rebound.Simulation` plus `assist.Extras` and integrates from the seed epoch -- 180 independent integrations for 3 candidates over 60 observations.

It dominates Gauss IOD. On two objects where Gauss and Herget both converge, the prefilter took 9.72 s of 9.80 s and 8.37 s of 8.45 s, against 0.07-0.08 s for the fit itself. Herget skips the prefilter entirely (`len(solns) > 1`, `iod.py:1104`), which is most of why it looks faster.

We already have the fast version in C++: `residuals_at_state(ephem, fit, detections, cov)` (`orbit_fit.cpp:741`) does the whole set in one forward+backward pass over the sorted times, and IOD candidates are already `FitResult` objects, so it can be called directly.

One related note: `residuals_at_state` sets IAS15 `adaptive_mode = 2` after `assist_attach` (`orbit_fit.cpp:580`), which `_predict_rho_hat` never does. Since that is the setting that fixed the close-Earth grind in #324, the prefilter's close-approach pass-through guard may be working around the missing mode rather than an intrinsic limit.

matthewholman:
Closes #555.

## The change

`filter_candidates_by_residual` called `_predict_rho_hat` once per (candidate, observation), and each call built a fresh `rebound.Simulation` plus `assist.Extras` and integrated from the seed epoch. `residuals_at_state` does the whole set in a single forward+backward pass over the sorted times, reusing one simulation — and IOD candidates are already `FitResult` objects, so it can be called directly.

## Measured

| object | obs | before | after | factor |
|---|---:|---:|---:|---:|
| 119839 | 587 | 29.73 s | 0.260 s | **114x** |
| 134367 | 627 | 17.44 s | 0.129 s | **135x** |
| 137749 | 848 | 19.37 s | 0.110 s | **176x** |
| 144605 | 686 | 44.59 s | 0.287 s | **155x** |
| 146058 | 1669 | 23.70 s | 0.108 s | **219x** |

**Survivor counts identical in every case.** The full test suite drops from **13m37s to 7m17s**.

## Two differences that are improvements, not side effects

- `residuals_at_state` sets IAS15 `adaptive_mode = 2` after attaching ASSIST — the setting that fixed the close-Earth grind in #324, which the per-observation path never applied.
- Its residuals carry the **light-time correction**, where the old ones were explicitly uncorrected ("coarse filter only"). The correct root therefore scores slightly better than it used to, which is the safe direction for a filter whose job is not to discard it.

The caller now passes the C ephemeris, which leaves `_get_python_ephem` and its module-level cache dead. Both removed.

## The close-Earth guard is KEPT, and the issue was right about it

Disabling the pass-through now costs about **0.1 s** and **correctly rejects** the candidate it waves through — so it was indeed working around the missing `adaptive_mode` rather than an intrinsic limit.

🔴 I did not remove it. The five cases above are main-belt objects whose **phantom** roots approach the Earth. The case the guard exists for is a genuine near-Earth object, where the **correct** root is the close one, and that is untested. Removing it changes which candidates survive rather than how fast they are scored, so it wants its own measurement on NEOs. The docstring now records the measurement and that reasoning rather than the stale "tens of seconds per propagation".

## Tests

The stub in `test_iod_picker.py` moves from the per-observation seam to the new whole-arc one, so the 609631 regression guard — a valid Gauss root at ~9500σ that was once silently dropped — still exercises the invariant it was written for. 11 passed.

Full suite: **579 passed, 1 skipped** (pre-existing). `black` clean.

⚠️ One unrelated observation while testing: an earlier full-suite run hung at 0.9% CPU with only a multiprocessing resource-tracker alive, and did not reproduce on re-run or when the multiprocessing suites were run in isolation (44 s, 15 passed). Flagging it rather than attributing it — it has the shape of #562.

### DIFF
--- a/src/layup/iod.py
+++ b/src/layup/iod.py
@@ -37,7 +37,7 @@ import math
 from typing import Callable, Optional, Sequence
 
 from layup.constants import GMtotal, SPEED_OF_LIGHT
-from layup.routines import FitResult, Observation, gauss
+from layup.routines import FitResult, Observation, gauss, residuals_at_state
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


kjnapier (approved):
Woah

PR #576, opened 2026-09-11
Try every single non-grav parameter, and take the best in a tier

### LINKED ISSUE #544: fit_nongrav="auto" cannot find a radial or out-of-plane non-gravitational acceleration
Two properties of the `auto` ladder mean it returns a gravity-only fit for objects with strong, published non-gravitational detections.

**The ladder never tries A1 alone.** It runs `A2 -> A1A2 -> A1A2A3` (`_AUTO_NONGRAV_LADDER`), so a radial-dominated object has to reach the second rung, where A2 is already absorbing signal.

**And the gate is on absolute fit quality, not on evidence for the parameter.** A gravity-only fit at or below `accept_reduced_chi2 = 1.5` is kept and no non-grav model is tried at all.

1I/'Oumuamua shows both. Its gravity-only fit gives reduced chi-square **0.637**, so `auto` stops there — yet fitting A1 alone gives **4.4 sigma**, within 1.7 sigma of the published 2.790e-7 +/- 3.57e-8 au/d^2. Fitting the full triplet is worse than useless: reduced chi-square is identical to three decimals across A1, A1A2 and A1A2A3 (0.588/0.589/0.589) while sigma(A1) inflates elevenfold and the sign flips.

The same shape of problem applies to the dark comets, where A3 is the significant term (Seligman et al. 2023, PSJ 4, 35) and A2 is not.

The ladder was designed for asteroids, where A2/Yarkovsky is expected and a failing gravity fit is the trigger, and it does that job. But an A1 rung, and a gate on parameter significance rather than reduced chi-square, would let it find the cases it currently cannot.

matthewholman:
Closes #544.

## Three defects, not two — and the third is why the first two are not enough

The issue names the missing A1 rung and the absolute-quality gate. Working through it turned up a third, and **fixing only the first two does not fix 1I** — I checked.

**3. Within a tier, the FIRST warranted model won, not the best.** "Most parsimonious" cannot discriminate between equally-complex models, so the tie was broken by the order the tuple happened to be written in.

## Measured on 1I/'Oumuamua

222 observations, Veres weights, every fit through the same entry point so the chi-squares are comparable:

| model | Δχ² | significance | warranted |
|---|---:|---|---|
| gravity only | — | reduced χ² **0.762** | gate fires |
| A2 | 13.877 | A2 **3.7σ** | ✅ |
| **A1** | **118.969** | A1 **10.9σ** | ✅ |
| A3 | 2.246 | A3 1.5σ | — |
| A1+A2 | 119.182 | A1 10.3σ, A2 **0.5σ** | — |
| A1+A2+A3 | 119.324 | A1 **0.9σ** | — |

Both A2 and A1 are warranted, so the old ladder took **A2 — the 3.7σ one — over A1 at 10.9σ and 8.6× the χ² drop**, purely on position. The triplet does not fit better (χ² unchanged to three decimals); it divides one detection between three parameters and inflates the uncertainty tenfold, exactly as the issue describes.

## The change

The ladder is tiered by parameter count. Within a tier, the warranted model with the lowest χ² wins; parsimony still decides *between* tiers, so a warranted single is preferred to a better-fitting pair.

1I now selects **A1 alone at 1.965e-07 ± 1.80e-08**, about 2σ from the published 2.790e-07 ± 3.57e-08.

A2 is still first in the tier, so an asteroid with a Yarkovsky signal and no competing model selects what it selected before.

## The gate is documented, not changed

🔴 Its failure mode is now precise. It assumes that omitting a real non-grav pushes reduced χ² above 1.5 — which holds only if the astrometric uncertainties are calibrated. Where they are conservative, which is most of the archive since the majority of modern astrometry takes the Veres catch-all rather than a per-station value, reduced χ² sits well below 1 whether or not the acceleration is there. 1I fits gravity-only at **0.762** and carries a 10.9σ A1.

`accept_reduced_chi2=0.0` already disables it. I did not change the default: doing so costs up to five extra fits per object, which is a survey-cost decision rather than a bug fix.

## Tests

Four new, all failing on revert: A1 alone reachable, A3 alone reachable (the dark-comet case), best-in-tier rather than first-in-tier, and parsimony still winning across tiers.

⚠️ One existing test needed correcting rather than accommodating. `test_select_auto_escalates_when_a2_alone_insufficient` stubbed the same significant result for every mask but A2, so under best-in-tier **A1 alone legitimately won** — the stub said it fit equally well with one parameter. It now makes no single parameter warranted, so it still tests what it was written to test: escalation past the whole one-parameter tier.

Full suite: **583 passed, 1 skipped**. `black` clean.

### DIFF
--- a/src/layup/orbitfit.py
+++ b/src/layup/orbitfit.py
@@ -214,7 +214,29 @@ def _parse_nongrav(fit_nongrav):
 # increasing complexity, and the first that converges, is well-conditioned
 # (flag 0), and is statistically warranted (see NongravAutoThresholds) is adopted;
 # otherwise the gravity-only fit is kept.
-_AUTO_NONGRAV_LADDER = (("A2",), ("A1", "A2"), ("A1", "A2", "A3"))
+#
+# Every SINGLE parameter is tried before any pair (issue #544). The ladder began
+# as A2 -> A1A2 -> A1A2A3, which can only find a radial or out-of-plane
+# acceleration by way of a rung that is already fitting A2 alongside it -- and
+# that is not a detour, it is fatal, because the parameters are strongly
+# correlated over a single apparition. Measured on 1I/'Oumuamua (222
+# observations, Veres weights):
+#
+#     A1 alone      reduced chi2 0.217   A1 = 2.06e-07   8.7 sigma
+#     A2 alone                   0.372   A2 = -2.36e-08  2.8 sigma
+#     A3 alone                   0.389   A3 = -6.06e-09  0.8 sigma
+#     A1 + A2                    0.218   A1 8.2 sigma, A2 0.1 sigma
+#     A1 + A2 + A3               0.218   A1 0.9 sigma  <-- from 8.7
+#
+# The triplet does not fit better -- reduced chi2 is unchanged to three
+# decimals -- it just divides one detection between three parameters and
+# inflates the uncertainty tenfold. A2 is kept first so that an asteroid with a
+# Yarkovsky signal still selects the model it selected before.
+_AUTO_NONGRAV_LADDER = (
+    (("A2",), ("A1",), ("A3",)),
+    (("A1", "A2"),),
+    (("A1", "A2", "A3"),),
+)
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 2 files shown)


matthewholman:
**A measured distribution for the acceptance gate**, from the MPC full-catalogue work — worth recording here because it turns the argument above from qualitative into a number.

20,000 numbered objects, **gravity-only fits, rejection disabled** so nothing is conditioned:

| weights | median χ²/dof | p90 | p99 | **fraction > 1.5** |
|---|---:|---:|---:|---:|
| stock Veres et al. (2017) | 0.332 | 0.493 | 1.347 | **0.81%** |
| per-station measured widths | 0.934 | 1.416 | 2.379 | **7.36%** |

🔴 **Under stock Veres the gate is nearly unreachable.** Fewer than one per cent of ordinary gravity-only fits exceed 1.5 *at all*, so a threshold there is not discriminating between "this object is missing a non-gravitational term" and "it is not" — it is discriminating against being weighted by a catch-all that is roughly 5x too loose. Under measured widths the same threshold sits at about the p90 of the distribution instead of past its p99.

That is consistent with 1I/'Oumuamua fitting gravity-only at 0.762 while carrying a 10.9σ A1: the reduced chi-square never had the chance to rise.

⇒ It does not change this PR, which deliberately documents the gate rather than altering it. But it sharpens what a future change should do: **a gate expressed in reduced chi-square inherits the calibration of the weights**, so if the default is ever revisited it should be set against a measured distribution rather than a round number — and the right value differs before and after an extended weighting table.

Measured by the USDF MPC-catalogue session; reported here so the number is attached to the code it bears on.

(AI-assisted analysis, reported by me.)

kjnapier (approved):
Okay

Issue #492, opened 2026-08-26
Per-object timeout in the batch path


matthewholman:
One object that never converges currently stalls the whole chunk. Fitting 108 short arcs in a single `orbitfit` call ran over an hour with no output; the same 108 fitted individually took under a minute of compute, because two of them grind indefinitely and nothing interrupts them.

At catalogue scale that is the difference between a run finishing and not. The pathological rate on short MPC arcs measured 1.9% at 7 days and 7.2% at 14 (#465). On 1.5M objects that is tens of thousands of fits that never return.

Suggested: a per-object wall-clock budget in the chunk loop, defaulting to something generous. On expiry, record the object with a distinct flag and move on. The result should say which objects were abandoned rather than silently omitting them.

This also makes a run resumable in a useful way — a chunk that hits the budget still produces output for everything else in it.

PR #577, opened 2026-09-11
Let a run finish when one object never converges

matthewholman:
Addresses #492 (mechanism; see "What this does not do" for the half that needs a decision).

## The bug

`_run_pool` used `starmap`, which returns only once the slowest task has finished — so one object that never converges withheld **every other result in the run**. 108 short arcs in a single call took over an hour with no output; the same 108 fitted individually took under a minute, because two of them ground indefinitely.

## The change

`_run_pool` gains an optional `per_task_budget_s`.

- **Unset** — the default, and every current caller — the path is the old `starmap`, unchanged.
- **Set** — tasks are submitted with `apply_async` and collected as they finish. At a deadline of `budget × ceil(n_tasks / n_workers)`, whatever is still running is abandoned, the pool is terminated, and the results of everything that finished are returned.

`process_data_by_id` exposes it as `per_object_budget_s`. That is exactly a per-object budget there, because the function already builds **one task per object** (`pid_list` splits on the primary id, `chunksize=1`) — which is also why this turned out much smaller than the issue implies.

Abandoned objects are **named** in the warning through `task_labels`, rather than silently omitted.

## 🔴 Nothing here interrupts a running task

A fit inside the C integrator cannot be preempted from Python: a signal handler runs only at a bytecode boundary. That is why a `signal.alarm` timeout does not bound these fits **at any value**. The budget abandons the task and kills its worker; it does not unwind the call. Said in the docstring so the next reader does not expect otherwise.

## Why the budget is per-batch rather than per-task

A worker cannot report when its task *started* — the parent sees submission and completion only. A strict per-task deadline needs a fresh pool per wave of `n_workers`, which is exact but respawns a worker every wave.

🔬 Measured, that respawn is **80 ms, of which `get_ephem` is 1 ms** — the ephemeris is memory-mapped, not re-read, which I had assumed would make per-object processes unaffordable and it does not. So the strict variant costs ~13% on a 0.6 s/object run and is available if it is ever wanted.

## ⚠️ What this does NOT do

**No caller sets a budget.** The mechanism is here and tested; wiring it to a CLI flag with a default — the part that would actually save a catalogue run — is deliberately not in this PR, because it changes behaviour for `orbitfit`, `convert` and `predict` and the default value is a judgement call.

For whoever picks that up: the measured pathological rate on short MPC arcs is 1.9% at seven days and 7.2% at fourteen (#465), and with `adaptive_mode = 2` the slow tail completes in about a second. **300 s per object** would therefore be generous by more than two orders of magnitude while still bounding a grind — but it is a number worth choosing rather than inheriting from me.

## Tests

Six, of which **four fail on revert**; the two that do not are the guards asserting the default path is unchanged. The central one checks both halves of the bug: the quick tasks come back **and** the call returns in under 25 s rather than waiting out a 30 s hang.

⚠️ Helpers are module-level because spawn pickles callables by reference, so a function defined inside a test cannot reach a worker.

Full suite: **585 passed, 1 skipped**. `black` clean.

### DIFF
--- a/src/layup/utilities/data_processing_utilities.py
+++ b/src/layup/utilities/data_processing_utilities.py
@@ -186,7 +186,7 @@ def _apply_func_with_kwargs(func, data, kwargs):
     return func(data, **kwargs)
 
 
-def _run_pool(tuple_task_list, n_workers):
+def _run_pool(tuple_task_list, n_workers, per_task_budget_s=None, task_labels=None):
     """
     General function to run multi_processing.Pool.
     This function spawns (_MP_CONTEXT) a pool of n_workers and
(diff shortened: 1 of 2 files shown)


kjnapier (approved):
Okay

Issue #568, opened 2026-09-10
layup demo prepare and how to doesn't have an example for layup unpack


mschwamb:
unpack doesn't have an example or demo files or a demo command

Issue #566, opened 2026-09-10
demo notebook that shows how to use incremental orbitfit from the API/function calls


mschwamb:
incremental orbitfit is not getting a CLI verb as this is an advanced feature where you're adding an additional data likely during a linking/detection process. Last call, it was agreed that there would be a demo notebook instead that shows how to call the incremental orbitfit.

PR #579, opened 2026-09-11
Propagate a fitted state and its covariance to a new reference epoch

### LINKED ISSUE #578: Propagate a fitted state and its covariance to a new reference epoch
## Summary

There is no way to move a fitted orbit to a different reference epoch. `convert` changes **parameterization** at a fixed epoch (BCART / KEP / BK); it does not change epochs. The integration machinery exists in C++ — `predict.cpp` calls `assist_integrate_or_interpolate` — but every entry point returns **sky** quantities: `PredictResult` is `rho`, `obs_cov`, `epoch`. Nothing returns a state.

Proposed: a C++ binding and a Python function that integrate a state, and its covariance, from one epoch to another using the same ASSIST integration the fitter uses.

**No CLI verb.** Every consumer identified below is programmatic, and a verb would add to the demo/example backlog (#439, #512, #566, #568) for no user need that exists yet. It can follow later if one appears.

## Why

**It is being written repeatedly, badly, outside the package.** Four separate throwaway implementations of the same fifteen lines in one working session (a trajectory comparison, a backward integration across a close encounter, a covariance cloud, an element check). The MPC full-catalogue benchmark does it too, and had to correct its own scoring after a 69-second epoch mismatch meant it was differencing two states referenced to different times — the kind of error that is invisible in the output.

**Layup already emits two epoch conventions with no way to reconcile them.** `gauss.cpp:215` sets the cold-path epoch to the middle observation of the IOD triplet, so every cold-started fit is referenced inside its own arc. The warm path inherits the initial guess's epoch (`orbit_fit.cpp:1489`, carried to `result.epoch` at :1534), so every warm-started fit sits at MPCORB's. Both are reasonable; the catalogue contains both; nothing can bring them together.

**It blocks #550.** The moved-seed retry is precisely "integrate MPCORB to an epoch inside the arc and fit from there". Without this, that issue has to grow its own propagator inline, and choose an epoch convention while it is at it.

## Scope

**In:** state propagation between epochs, forward and backward, for a single object and for an array of them; covariance propagated with it.

The covariance is the part worth arguing about, so the argument is here rather than implied. `C_new = Phi C Phi^T` needs the state-transition matrix, which ASSIST can supply through variational particles — the same mechanism the fitter already uses for its partials. It roughly doubles the work. It is included because a state without its covariance at the new epoch is half a tool: scoring a fit against a reference needs both, and so does any comparison of two catalogues at a common epoch. It is also the obvious thing to cut if a smaller first version is wanted.

**Out:** non-gravitational parameters carried through the propagation, and any automatic re-referencing inside `orbitfit`. Both are follow-ons, and the second needs a decision this issue does not make.

## Two properties it must have

🔴 **No light-time correction.** Propagation to an epoch is a dynamical operation. `predict` applies light time because it computes an observable; conflating the two would be subtle and expensive.

🔴 **Fail cleanly outside ephemeris coverage.** The planetary ephemeris spans 1550-2650 (see #563); a request outside it must raise rather than return silently wrong numbers. `de440s.bsp` narrows this further for observer positions (#543), though that limit does not apply to a pure state propagation.

## Note

Consumers should be able to ask for a *specific* target epoch rather than only "the catalogue epoch" — one caller wants MPCORB's, another wants the middle of an object's own arc, a third wants whatever a reference solution is published at.

<sub>AI-assisted analysis, reported by me.</sub>

matthewholman:
Closes #578.

## The gap

`convert` changes the **parameterization** of an orbit at a fixed epoch; nothing changed the epoch. The integration machinery was already here — `predict.cpp` drives `assist_integrate_or_interpolate` — but every entry point returns **sky** quantities (`PredictResult` is `rho`, `obs_cov`, `epoch`). So a state at a different epoch had to be obtained by writing the integration again outside the package, which had happened four times in one working session.

Layup also emits two epoch conventions with no way to reconcile them: a cold-started fit is referenced to the middle observation of its IOD triplet (`gauss.cpp:215`), a warm-started fit inherits its initial guess's epoch (`orbit_fit.cpp:1489`).

## What this adds

**C++** — `propagate_state(ephem, fit, target_epoch)` returns the re-referenced `FitResult` **and** the 6x6 state-transition matrix. The covariance is carried as `C' = Phi C Phi^T`, with Phi from six variational particles seeded along the state axes — the same mechanism the fitter uses for its partials.

Phi is returned rather than hidden: it is independently useful, since it is what says which directions of a solution are soft.

**Python** — `layup.propagate.propagate_row` and `.propagate`, over `orbitfit` output rows, handling the MJD/JD offset in one place.

**No CLI verb.** Every consumer identified is programmatic, and a verb would add to the demo/example backlog (#439, #512, #566, #568) for no user need that exists yet. It can follow if one appears.

## Deliberate properties

- 🔴 **No light-time correction.** Propagation is dynamical; `predict` applies light time because it computes an observable. Conflating the two would be subtle and expensive.
- 🔴 **On failure the orbit comes back UNCHANGED** with `flag != 0` — including a target outside ephemeris coverage (1550–2650, see #563) — so a caller who ignores the flag holds the original rather than a fabricated state.
- **A zero interval is a free identity**; no integration is burned.
- **A zeroed covariance stays zeroed**, which is the right answer for an unconverged fit (#547) rather than an error.

## Tests

The central one is **composition**: A→B→C must agree with A→C, in the state (5e-14 au), the covariance (7e-13 relative) and Phi itself (6e-10). That is a self-check requiring **no reference orbit**, so it catches a wrong sign, a dropped perturber or a mis-indexed Phi with nothing to compare against. Also pinned: round trip, symmetry and positive semi-definiteness of the propagated covariance, and that an out-of-coverage request leaves the orbit untouched at both ends of the range.

⚠️ **One test exists only to catch a silent failure this repository has hit before.** `reb_simulation_add_variation_1st_order` defaults to `testparticle = -1`, and ASSIST fills variational accelerations only where `testparticle == j` — so with the default, **Phi comes back as exactly the identity with no error raised**, and a propagated covariance is silently unchanged. `add_variational_particles` passes `np` explicitly; the test asserts `max|Phi - I| > 1` over a real interval so a future refactor cannot reintroduce it quietly.

Full suite: **593 passed, 1 skipped**. `black` clean.

## Not done, on purpose

**Not parallelised.** The catalogue use is 1.5M rows and `process_data_by_id` exists for exactly that, but wiring it means choosing a chunking and inheriting worker-pool behaviour that is being changed in #492. Follow-on rather than smuggled in here.

Also out of scope: non-gravitational parameters carried through the propagation, and any automatic re-referencing inside `orbitfit` — the second needs a decision this does not make.

### DIFF
--- a/src/layup/propagate.py
+++ b/src/layup/propagate.py
@@ -1,0 +1,128 @@
+"""Move fitted orbits to a different reference epoch (issue #578).
+
+``convert`` changes the *parameterization* of an orbit at a fixed epoch;
+this changes the epoch. The two are complementary and neither substitutes for
+the other.
+
+The state is integrated with ASSIST, the same dynamics the fitter uses, and the
+covariance is carried with it as ``C' = Phi C Phi^T`` where ``Phi`` is the
+state-transition matrix over the interval.
+
+No light-time correction is applied. Propagation to an epoch is a dynamical
+operation; :mod:`layup.predict` applies light time because it computes an
+observable, and conflating the two would be a subtle and expensive error.
+
+Why this exists: Layup already emits two epoch conventions and could not
+reconcile them. A cold-started fit is referenced to the middle observation of
+its IOD triplet, so it sits inside its own arc; a warm-started fit inherits the
+epoch of its initial guess, so it sits whereever the catalogue that seeded it
+does. Comparing the two, or comparing either against a reference solution
+published at a third epoch, requires moving one of them.
+"""
+
+import logging
+
+import numpy as np
+
+from layup.routines import get_ephem, propagate_state
+from layup.utilities.data_processing_utilities import parse_fit_result
+from layup.utilities.cache_location import default_cache_dir
+
+logger = logging.getLogger(__name__)
+
+__all__ = ["propagate", "propagate_row"]
+
+MJD_TO_JD = 2400000.5
+
+
+def _cache(cache_dir):
+    return str(default_cache_dir()) if cache_dir is None else str(cache_dir)
+
+
+def propagate_row(fit_row, target_epoch_mjd_tdb, cache_dir=None, return_stm=False):
+    """Propagate one fit result to ``target_epoch_mjd_tdb``.
+
+    Parameters
+    ----------
+    fit_row : numpy structured array
+        One row of ``orbitfit`` output. Its covariance columns are carried if
+        present; a row without them propagates the state alone.
+    target_epoch_mjd_tdb : float
+        Target epoch, MJD TDB -- the same units as the ``epochMJD_TDB`` column.
+    cache_dir : str, optional
+        Kernel cache. Defaults to Layup's.
+    return_stm : bool
+        Also return the 6x6 state-transition matrix.
+
+    Returns
+    -------
+    numpy structured array
+        A copy of ``fit_row`` at the new epoch. ⚠️ **Check the flag.** If the
+        propagation fails -- most often a target outside the ephemeris
+        coverage, 1550 to 2650 -- the row comes back with ``flag != 0`` at its
+        ORIGINAL epoch and state, so a caller that ignores the flag gets an
+        unchanged orbit rather than a plausible wrong one.
+    """
+    row = np.atleast_1d(fit_row).copy()
+    if len(row) != 1:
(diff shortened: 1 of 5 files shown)


kjnapier (approved):
Looks okay to me

PR #580, opened 2026-09-11
Compare comet 1/a against the catalogue's own uncertainty

### LINKED ISSUE #508: test_apply_comet's tolerance is far looser than the agreement it checks
`test_apply_comet` compares the original 1/a against the CODE catalogue with
`atol=1000` (units of 1e-6/au). Running the full 369-comet fixture: 325 of 334
reproduce the catalogue to a median of **0.009**, against a median catalogue
uncertainty of 2.2. The test would pass with the integration badly wrong.

Tightening it first needs the nine that disagree (|delta| 1.6-29). Not step
size -- 10, 4 and 1 day steps agree to six significant figures. Non-gravitational
forces explain two, both the smallest-perihelion cases: C/2020 S4 needs
A1 = 7.6e-8 au/d^2, and JPL fits 6.34e-8 under the same distant-activity
scaling. The other seven carry osculating elements inconsistent with the
catalogue value stored beside them; substituting current JPL elements takes
C/2003 WT42 from 29.0 to 1.0.

Doing that substitution for the *whole* sample makes agreement much worse
(median 0.009 -> 3.9), so this is nine rows, not a stale fixture.

Suggest fixing or excluding those nine, then setting `atol` near 0.1.

matthewholman:
Closes #508. cc @carrholt @awilson110 @hannorein @mschwamb — **this edits a test fixture**, see the fixture section.

## The defect

`test_apply_comet` compared against the CODE Catalogue with `atol=1000`, in units of 1e-6/au, where the median agreement is **0.0095** and the median catalogue uncertainty is **2.12**. That is roughly a hundred thousand times looser than the quantity it checks: the integration could have been badly wrong and the test would still have passed.

An absolute tolerance is also the wrong instrument, because the catalogue's quoted uncertainty spans three orders of magnitude across the sample:

| comet | delta | quoted sigma | |
|---|---:|---:|---|
| C/1980 E1 | 2.28 | 5.16 | **0.4 sigma — not a failure** |
| C/2010 U3 | 1.58 | 0.29 | **5.5 sigma — a failure** |

An `atol` cannot tell those apart. `dinv_ao` is published beside `inv_ao` precisely so that it can, and the test now uses it.

⚠️ **The issue's own suggestion would not have worked.** It proposed fixing nine rows and setting `atol` near 0.1 — but **38 of 334 exceed 0.1**, so 29 would still have failed. That only became visible on measuring the whole sample.

## 🔴 The fixture edit — please review this as data, not code

Seven rows of `tests/data/code_LPCs.csv` carried osculating elements inconsistent with the catalogue value stored beside them. They are replaced with JPL Horizons elements **at each row's own epoch**. Effect, in units of the catalogue's own sigma:

| comet | before | after | | comet | before | after |
|---|---:|---:|---|---|---:|---:|
| C/2003 WT42 | 111.7 | **4.0** | | C/2020 U4 | 9.8 | **<1** |
| C/2010 U3 | 15.9 | 5.5 | | C/2012 K8 | 11.7 | **<1** |
| C/2015 XY1 | 13.5 | **1.8** | | C/2009 F4 | 7.1 | 5.5 |
| C/2020 F2 | 5.2 | **1.1** | | | | |

That the whole-sample substitution makes agreement *worse* (median 0.009 → 3.9) is what shows these are seven bad rows rather than a stale fixture.

## Nothing is excluded

@mschwamb asked whether excluding nine comets was right. It is not, and nothing is excluded.

Over the full catalogue the agreement is median **0.0042 sigma**, p90 **0.079**, and the bound is **6 sigma**. Two comets exceed it: **C/2020 S4** and **C/2007 Q3**, the two smallest perihelia, where a gravity-only integration cannot reproduce a comet with a real non-gravitational acceleration. C/2020 S4 needs `A1 = 7.6e-8 au/d^2`, and JPL independently fits `6.34e-8` under the same distant-activity scaling — so the explanation is corroborated rather than assumed.

🎯 They are outside the bound but **not outside the comparison**: a second test asserts they **remain discrepant**, and its failure message says what to do if they ever agree (either non-gravitational forces were added to this path, in which case tighten; or the disagreement went away for a reason worth finding). A test that quietly stops covering its awkward cases is exactly the failure that concern was about.

Also asserted: **the median agreement across the sample**. A per-comet bound is loose by construction and would not catch a systematic drift that leaves every individual comet inside it.

## Verification

Reverting the fixture alone makes the new test fail with:

```
C/2003 WT42: original 1/a differs from the CODE Catalogue by -29.035 = 111.7 sigma of its quoted 0.260
```

Full suite **678 passed, 2 skipped**. Full catalogue (`LAYUP_SLOW_TESTS=1`) **20 passed**, 8m12s.


kjnapier (commented):
Looks good to me

kjnapier (approved):

Issue #582, opened 2026-09-12
Covariance is formed by inverting the normal matrix, so it comes back indefinite on ill-conditioned short arcs


matthewholman:
A converged fit can return a covariance whose position block has a negative trace, with `flag = 0` and no indication that anything is wrong. In a re-fit of the MPC catalog this happens on about 5 per cent of the converged fits in one short-arc population (40 of 768), at `N` = 4-60 and a median reduced chi-square of 1.40 — they look like ordinary good fits.

The cause is that the step and the covariance take different routes through the same matrix.

The step avoids squaring the condition number, which is what #536 was for. `orbit_fit.cpp:908-916` records why, on (6489) Golevka: `cond(sqrt(W) B) = 2.6e10` becomes `cond(B^T W B) = 9.2e19`, past what double precision resolves, and the solve silently rank-truncates. So the step is solved as a damped least-squares problem by Householder QR:

```
orbit_fit.cpp:938    dX = A.householderQr().solve(rhs);      // A = [sqrt(W) B ; sqrt(lambda) I]
```

The covariance is not:

```
orbit_fit.cpp:887    C = Bt * W * B;
orbit_fit.cpp:1318   cov = C.inverse();
```

`C.inverse()` is the squared-condition-number path the QR was introduced to avoid. Where `cond(sqrt(W) B)` reaches ~1e10 the inverse is dominated by rounding and is no longer positive definite, so `cov` is emitted with negative position variances.

The marginal cases show it is conditioning rather than a code path that writes bad values. One object fits identically on two runs — same `N`, `n_rej`, `ndof`, `niter`, arc — with chi-square differing by 3e-6 relative, and the sign of the position trace flips between them. Repeated fits on one node are bit-identical; the variation is between nodes at the 1e-6 level.

Two fixes, and they are independent:

1. Form the covariance from the same decomposition. With `sqrt(W) B = QR` (undamped), `B^T W B = R^T R`, so `cov = R^-1 R^-T` by two triangular solves, at unsquared conditioning. The `R` already computed is of the *damped* system, whose `sqrt(lambda) I` rows flatter it precisely when `lambda` is large, so this needs one QR of the undamped `sqrt(W) B` at convergence.

2. Regardless of (1), test the result before emitting it. An LDLT that fails, or any non-positive diagonal entry, should set a flag rather than return silently. A caller cannot detect this: the matrix is full and finite, and only the eigenvalues give it away.

[...]

PR #585, opened 2026-09-13
#477: thread a scaled convergence tolerance through the Cartesian fit (opt-in, default off)

### LINKED ISSUE #477: Convergence test is absolute, so it is not invariant under reparameterization
## Summary

`converged()` in `src/lib/orbit_fit/orbit_fit.cpp:815` decides that a
differential-correction fit has converged when **every** component of the
Levenberg-Marquardt step is smaller than a single absolute number:

```cpp
for (size_t i = 0; i < dX.size(); i++)
{
    if (std::isnan(dX(i)) || abs(dX(i)) > eps)
    {
        return 0;
    }
}
return 1;
```

with `eps = 1e-12` at both call sites (`orbit_fit.cpp:1160` for the Cartesian
fit, `bk_fit.cpp:104` for the Bernstein-Khushalani fit).

The test is absolute, but the parameters it is applied to are not commensurate.
For a Bernstein-Khushalani fit of a trans-Neptunian object the six parameters
span roughly seven orders of magnitude:

| parameter | meaning | typical size at 40 au |
|---|---|---|
| `alpha`, `beta` | gnomonic tangent-plane angles | ~1e-1 |
| `gamma` | 1/(barycentric distance), au^-1 | ~2.5e-2 |
| `adot`, `bdot`, `gdot` | scaled velocity components, au/day | ~3e-5 |

A step of 1e-12 is a fractional change of 1e-11 in `alpha` but 3e-8 in `adot`.
The stopping rule therefore demands about three more significant figures of the
angles than of the rates, for no stated reason. The same `eps` is also applied
to the Cartesian fit, whose parameters (positions in au, velocities in au/day)
differ among themselves by a further ~1e-2.

## Why this matters beyond tidiness

Because the criterion is absolute, **it is not invariant under a change of
parameterization**. Rescaling a parameter — which does not change the orbit
being fitted, or the chi-square surface, or the location of its minimum —
changes where the fit stops.

This surfaced while implementing #445. Under the rescaled convention `gdot`
grows by a factor of 1/gamma (40x for a TNO), and the iteration count on one
noise-free test case moved from 44 to 30 purely as a result. In that instance
the effect on the answer was negligible and the real cause of the residual
difference lay elsewhere (tightening `eps` to 1e-16 changed no result), so
**this is not a bug report about a wrong orbit** — it is a latent fragility
that #445 happened to expose.

The concern for real work is the short-arc regime, where `gamma` and `gdot` are
weakly constrained and the normal matrix is poorly conditioned. There the LM
step in those two directions is large and slow to shrink, so an absolute
threshold is a stricter demand on exactly the parameters the data cannot pin
down, and the loop is more likely to run to `iter_max` than to converge.

## Suggested fix

Use a scaled test, e.g. accept when

```
|dX_i| <= eps_abs + eps_rel * |x_i|
```

for each parameter, or compare the step against the formal uncertainty already
available from the normal matrix, `|dX_i| <= frac * sqrt(C_ii)` — which is the
statistically meaningful statement ("the step is small compared with what the
data can determine") and is automatically invariant under reparameterization.

## Caution for whoever picks this up

`converged()` is shared by the Cartesian and Bernstein-Khushalani fits, so a
change alters convergence behaviour for **every** fit the package performs, not
just the BK path. It needs a before/after comparison over a realistic
population — fitted state, formal covariance, iteration count and non-
convergence rate — rather than a unit test alone. The full-catalog harnesses
would be a reasonable place to measure it.

Deliberately kept out of #476 for that reason: that PR is a reparameterization
whose whole claim is that fitted orbits do not move, and changing the stopping
rule in the same change would have undermined the comparison.

matthewholman:
Closes #477.

The convergence test compares every parameter step against the same absolute `eps`, so it is not invariant under reparameterization and it demands more significant figures of a large parameter than of a small one. This threads a scaled tolerance through the Cartesian fit as an **opt-in parameter**.

A step counts as converged when it is small compared with what the data can determine, `sigma_i = sqrt((N^-1)_ii)`. `eps` stays an absolute **floor**, so a well-determined parameter is never held to a looser bar than today.

**`conv_frac = 0.0` is the default and reproduces the current predicate exactly** — verified as zero differences on `flag`, `niter`, `n_rej`, `N` and `ndof` across **48,000 paired objects**, both sections of the MPC catalogue, `ITER_MAX = 100`.

## Why it ships inert, when the point of the issue is that the current test is wrong

This PR originally defaulted to `conv_frac = 3e-5`, on the reasoning that shipping a correction in a disabled state leaves the defect in place. **That default failed three of layup's own tests and has been reverted** (see the commit history and the comment below).

The failures were **3 of 678**, and they share one mechanism:

- `test_radar_end_to_end` — noise-free synthetic, χ² = 2.587e-6 against a `< 1e-6` bar
- `test_radar_validation` — noise-free synthetic, χ² = 8.445e-4 against `< 1e-4`
- `test_bk_everywhere[scattered_70AU_arc_014.00d]` — a 70 AU object on a 14-day arc

`sigma_i` is the **marginal** uncertainty, and `N^-1` is dominated by the **flattest** direction. So the tolerance is loosest exactly where the geometry is weakest, and the fit can stop after very few iterations. With eigenvalues 1e12 (stiff) and 1e-4 (flat) the marginal sigma is ~100, the bar is `3e-5 × 100 = 3e-3`, and a 1e-3 step with any component along the stiff direction changes χ² by ~1e6 and still passes.

The `have_accepted_step` guard in this patch does not cover it, because on weak geometry **one** accepted step already satisfies a scaled tolerance.

Attribution was checked rather than assumed: `main` at this branch's base commit passes the same workflow, and on a single machine with one environment the unpatched build passes all four tests where the 3e-5 build fails three.

⇒ **A non-zero default needs the premature-stopping behaviour addressed, not merely a smaller `frac`.** That is why the parameter is here and off, rather than tuned to whichever value most recently passed.

## The measurement, for whoever picks a value

A six-point sweep over 48,000 objects per arm, scored on two axes: effect on the **certified catalogue**, and effect on a **29-object cohort of known fit failures** that #477 exists to rescue.

| `conv_frac` | good gains | bad | lost | lost good | net / 1.509M | main belt | Earth-crossers |
|---|---:|---:|---:|---:|---:|---:|---:|
| 1e-5 | 32 | 3 | 4 | 3 | +822 | 2/18 | 0/11 |
| 3e-5 | 33 | 4 | 10 | 9 | +716 | 3/18 | 3/11 |
| 1e-4 | 40 | 2 | 23 | 21 | +504 | 4/18 | 3/11 |
| 1e-3 | 43 | 5 | 107 | 105 | −1,564 | | |
| 3e-3 | 49 | 5 | 173 | 169 | −3,154 | | |
| 1e-2 | 49 | 7 | 269 | 255 | −5,645 | 5/18 | 2/11 |

There is **no plateau** — the net rises monotonically as `frac` falls, driven by losses (across the sweep losses move 67.2x while good gains move 0.62x). But the catalogue-best value abandons the objects the issue is for: at 1e-5 the gap cohort collapses to 2/18 and 0/11.

Three caveats a reader should carry, because the net hides them:

1. **Two-sided churn.** At 3e-5, 33 good certifications are gained and **9 good ones are lost**. The *set* is unstable too: 1e-4 → 3e-5 shares only 5 of 7 gap conversions, and 0.01 → 1e-4 only 2 of 5. A count that holds steady is not evidence the same objects are being helped.
2. **No setting is uniformly safer.** `K19Y03L` is certified at 89x off at 3e-5 but honestly refused at 1e-4; `K17SC0D` is refused at 3e-5 but certified at 153x by both 1e-4 and 0.01. They trade specific objects. The bad-certification counts across the sweep (2, 3, 4, 5, 5, 7) are Poisson noise on single objects and are **not** an ordering.
3. **Bad-certification rate**, measured at 1e-4 on a random catalogue sample: **0.0035%** section-weighted, 95% upper 0.0220%, ~53 per 1.5M. The numbered zero is *irrelevance, not safety* — there is exactly one refusal in 24,000 numbered objects for the predicate to act on.

All of the above was measured against a fixed weight table; #548 changed the star-catalog decode and therefore the weights, so a rerun on a newer build is a different error model.

## An alternative that was tried and does not work

Testing the step in the **N metric** — `dX^T N dX < tol`, the predicted χ² reduction, which is exactly reparameterization-invariant and cannot blow up on degenerate geometry — passes all four failing tests and converges in **fewer** iterations on 49 of 150 catalogue objects, with identical flags.

But it converts **1 of 18** gap objects at its loosest working setting and **0 of 18** one step tighter, against the marginal-sigma test's 3–5 of 18. Converting a refusal requires the predicate to *accept* a state it currently rejects, and that is the same looseness the tests forbid. **On that cohort the two properties are in tension**, which is worth knowing before someone tries the same substitution.

## Scope

- **Cartesian path only.** `bk_fit.cpp:188` calls `converged()` with three arguments, so the BK basis keeps the absolute test.
- **`conv_frac` is threaded through `orbitfit.py`** (`1d1710c`). Before that commit `grep -rn conv_frac src/layup/` was empty and `run_from_vector_with_initial_guess` was called at five sites, none passing it, so the parameter had a default a caller could receive but never change. It is now forwarded at all five and exposed on `orbitfit`, `sequential_update` and `incremental_orbitfit`, defaulting to `0.0` everywhere, so nothing changes for existing callers. There is still no CLI flag, matching the CLI's existing treatment of `fit_nongrav`, `per_arc` and `skip_unchanged`.
- **An unusable covariance falls back to `eps`.** If `(N^-1)_ii` is negative or non-finite the scaled tolerance is not formed and the parameter is held to the absolute test, rather than the fit refusing to converge. Indefinite normal matrices do occur — see #582.
- **No conditioning guard is included.** Pairing the predicate with a `cond(sqrt(W)B)` cutoff was considered and is *not* validated: on a random catalogue sample 0 of 842 objects reach a 1e10 threshold (max 2.29e9), so it never acts there, and the two bad certifications that were found are not ill-conditioned (3.43e5 and 3.02e7, inside the good distribution). It would be insurance for a population a random sample does not contain.

---

https://claude.ai/code/session_01PLuw1wksyUXStpjNFKkT2J


matthewholman:
**Converting this to a draft: CI is red, the failures are caused by the 3e-5 default, and I would rather report that than have a reviewer discover it.**

`testing-and-coverage` fails on all four matrix jobs with **3 failed, 675 passed**:

- `test_radar_end_to_end::test_radar_end_to_end_recovers_orbit` — noise-free synthetic, χ² = 2.587e-6 against a `< 1e-6` bar
- `test_radar_validation::test_radar_fit_recovers_synthetic_orbit` — noise-free synthetic, χ² = 8.445e-4 against `< 1e-4`
- `test_bk_everywhere::test_bk_beats_cartesian_on_short_arc_distant[scattered_70AU_arc_014.00d]` — a 70 AU object on a 14-day arc

## Attribution

This is not pre-existing and not environmental. `main` at `8fbc702` — the exact base of this branch — passed this workflow on 2026-09-11. Running just these three tests on one machine, same environment, two builds differing only by the patch:

| build | result |
|---|---|
| `8fbc702`, unpatched upstream | **4 passed** |
| this branch @ `435d22c`, `DEFAULT_CONV_FRAC = 3e-5` | **3 failed, 1 passed** |

So the default change causes them.

## Mechanism — one cause, not three

The tolerance is `max(eps, frac * sigma_i)`. Where parameters are well determined `sigma_i` is small, `eps` dominates, and the scaled test is inert — which is why the other 675 tests are unaffected. Where the **geometry is weak** the formal sigma is very large (the short-arc along-track/radial degeneracy; the same conditioning that motivates #582), so `frac * sigma_i` swamps `eps` and the fit can stop after very few iterations.

All three failing cases are exactly that regime: two noise-free radar fits that stop before χ² reaches numerical zero, and a 14-day arc at 70 AU. In the BK case the tell is that the *Cartesian* drift drops to 1.06e-5 — on a synthetic whose seed is near the truth, that is what "barely moved" looks like, not a better fit.

This is the same failure mode the `have_accepted_step` guard in the first commit exists for. The guard is not sufficient here, because it only requires **one** accepted step, and on weak geometry one step can already satisfy a scaled tolerance.

It is also visible in the catalogue sweep quoted in the description, as the early-stop column: 13.82% of short-arc fits at `frac = 1e-2`, falling to **0.68% at 3e-5**. Small, but not zero — and these tests are precisely the cases that expose it.

## What this does and does not change

It does not affect the catalogue ledger in the description (+716 net at 3e-5, both axes cleared); those are fits of real objects against JPL, and they are unchanged. What it shows is that the *default* is not free: on weak geometry the scaled test can terminate a fit that the absolute test would have kept refining, and layup's synthetic tests encode a reasonable expectation — noise-free data should fit to ~zero χ² — that it breaks.

Three ways forward, and I do not think this is mine to choose:

1. **Ship the parameter with `conv_frac = 0.0`** (the first commit alone). Reproduces current behaviour exactly, CI green, the scaled test available but opt-in — though note it is not currently reachable from `orbitfit.py`.
2. **Keep the 3e-5 default and treat the three tests as the finding.** Two of them assert something physically reasonable, so this means deciding the scaled test should not apply on degenerate geometry.
3. **Bound the scaled tolerance above** — a ceiling, or requiring more than one accepted step — so it cannot become arbitrarily loose exactly where the normal matrix is near-singular. That is new design and would need its own measurement rather than being asserted.

The two commits are kept separate so the first can be taken on its own if that is the preferred route.

matthewholman:
**Resolved: the default is reverted to `0.0`, CI is green on all seven checks, and this is ready for review.**

`7516d88` reverts the 3e-5 default. I reverted rather than force-pushed so the attempt and its reason stay in the history alongside the comment above. The net diff against `main` is now just the parameter, inert: `conv_frac = 0.0` reproduces the current predicate exactly, verified as zero differences on `flag`, `niter`, `n_rej`, `N` and `ndof` across 48,000 paired objects.

The title and description are rewritten to match. The description keeps the measured sweep, the two-sided churn, `K19Y03L`, and the bad-certification rate, for whoever picks a value later — the point being that **a non-zero default needs the premature-stopping behaviour addressed, not merely a smaller `frac`.**

One addition worth flagging, since it saves repeating an experiment. The obvious alternative — testing the step in the **N metric**, `dX^T N dX`, which is exactly reparameterization-invariant and cannot blow up on degenerate geometry — was built and measured. It passes all four failing tests and converges in *fewer* iterations on 49 of 150 catalogue objects with identical flags. But on the 29-object cohort of known fit failures it converts **1 of 29** at its loosest working setting and **0 of 29** one step tighter, against the marginal-sigma test's 7 of 29. Converting a refusal requires the predicate to *accept* a state it currently rejects, which is the same looseness the tests forbid — so on that cohort the two properties are in direct tension. That is recorded in the description rather than proposed here.

kjnapier (approved):
Okay

PR #589, opened 2026-09-18
Skip a radar obs80 record instead of aborting the file, and count every skip

### LINKED ISSUE #588: obs80 reader: a radar record aborts the whole file instead of being skipped and counted
A matched radar (`R`/`r`) two-line record aborts the whole read, taking every other object in the file with it. On a four-line file -- two ordinary observations plus one radar pair -- `read_rows()` raises and returns nothing:

```
ValueError: Radar (R/r) obs80 two-line records are not supported by
Obs80DataReader (object 00009); ingest radar via ADES delay/doppler columns.
```

The refusal itself is right and the message is clear. The problem is that it is fatal to the file rather than to the record. `Obs80DataReader` has no `try`/`except` and no logging anywhere, so there is no path that skips a record it cannot represent, and no count of what was skipped. A user with one radar observation in an archive extract loses the archive.

Two things would help:

- skip a record the reader deliberately refuses, rather than raising, and keep genuine corruption (a short line, a mismatched observatory code) fatal as it is now;
- report how many records were skipped and of what type, so a partial loss is visible. Today a silent partial loss is indistinguishable from a clean read.

The natural place for both is `_iter_records`, which already drops orphaned and mismatched continuation lines silently. Doing it there also keeps `get_row_count()` consistent with what `read_rows()` actually returns.

---

**Correction, same day.** This issue as first filed was written against a stale checkout and was mostly wrong. For the record, on current `main` these claims of mine do **not** hold:

- *"treats every second line as a satellite position"* -- false. `convert_obs80` dispatches on the first line's note 2 (`ec98eda`).
- *"a roving-observer record aborts the read"* -- false. Roving records read correctly to `sys = WGS84` with geodetic longitude, latitude and altitude.
- *"an East longitude near 2xx degrees is accepted as ICRF_AU"* -- false. That branch is reached only for `S`. My test fixture was missing the `1` that a real roving continuation line carries in column 33, and the failure was in my fixture, not the code.

Apologies for the noise. Only the radar abort and the missing skip-and-count survive, and the issue above is rewritten to those.

matthewholman:
Closes #588.

A single radar (`R`/`r`) two-line record made `read_rows()` raise, so one radar observation anywhere in an 80-column file cost the caller every other observation in it. Refusing radar is right -- the output dtype has no columns for range and Doppler -- but the refusal belonged to the record, not the file. It is now skipped.

The skip happens in `_iter_records` rather than at conversion time, so `get_row_count()` and `read_rows()` still agree. Skipping at conversion time would have had the row count promise rows that never arrive, and block planning reads that count.

The larger half is that nothing counted what a read dropped. Deleted observations, orphan continuation lines and first lines whose continuation is missing were all discarded silently, so a read that lost input was indistinguishable from a clean one. Every skip is now tallied by reason, readable from `skipped_record_counts`, and reported through logging: the first of each reason as it happens, then a total when the walk finishes. A deleted record that carried a continuation line counts once, not twice.

Verified on the archives in `tests/data`: `03666.txt` and `newy6.txt` return 4,312 and 6,585 rows exactly as before, and `03666.txt` now reports the one deleted observation its test comment has always described. Eight new tests; the existing 21 are unchanged.


kjnapier (approved):
Looks good

Issue #282, opened 2025-05-23
Implement Additional ADES Observatory Formats as Needed


wilsonbb:
[observatorypos.pdf](https://github.com/user-attachments/files/19914269/observatorypos.pdf) describes the various formats ADES allows user to provide observatory positions in. https://github.com/Smithsonian/layup/pull/281 provides support for "ICRF_AU" and "ICRF_KM" while assuming that "ctr" is always 399.

This issue is for tracking the remaining coordinate frames as needed (WGS84, ITRF, IAU)

matthewholman:
### Status (verified on `main`, 2026-07-02)

Still only `ICRF_AU` / `ICRF_KM` with `ctr = 399` are supported. WGS84, ITRF and IAU are **explicitly rejected** in `LayupObservatory.populate_observatory` (`utilities/data_processing_utilities.py`):

```python
if data["sys"] not in ["ICRF_KM", "ICRF_AU"]:
    raise ValueError(f"... unsupported reference frame {data['sys']} ... Please use ICRF_KM or ICRF_AU.")
if data["ctr"] != 399:
    raise ValueError(f"... unsupported center {data['ctr']}. Please use the 399 (Earth).")
```

So this is genuinely untouched — the recent observatory work (space-based obscodes via Horizons #377/#55, observer velocities #147, ADES XML reader #378) all route through this same path and don't add a new input frame. (Note: it fails with a *clear* error, so this is an honest "tracked but not implemented" state, not a crash.)

### Suggested priority ordering

- **WGS84 (geodetic lon/lat/height) — do this first.** It's the common case for ground/roving stations, and it **shares its core with #402**: MPC roving-observer records (obs80 `V`/`v` lines) carry geodetic longitude/latitude/altitude, so the geodetic→geocentric conversion needed here is the same one #402 needs. Implementing WGS84 support and roving-observer ingestion together is the natural pairing.
- **ITRF (Earth-fixed Cartesian)** — occasional; medium priority. Mostly a rotation to ICRF via Earth orientation (SPICE already provides this for the fixed-station path).
- **IAU (body-fixed, non-Earth `ctr`)** — observations taken from another body; niche. Truly "as needed".

### Where to implement

`populate_observatory` already branches on `sys`/`ctr` for the non-fixed path — the new frames slot in there: convert the supplied position to the geocentric (then barycentric) observer state layup uses internally, alongside the existing `ICRF_AU`/`ICRF_KM` handling.

[...]

PR #590, opened 2026-09-18
Use a roving observer's WGS84 position instead of refusing it

matthewholman:
Partly addresses #282 -- the WGS84 case, which is the one the MPC archive actually carries (roving observer, code 247). ITRF and IAU are untouched.

A roving-observer record reports geodetic longitude, latitude and height on WGS84 rather than a geocentric vector, and `populate_observatory` accepted only `ICRF_KM`/`ICRF_AU`. The observations parsed and were then thrown away at fit time, which is a strange place to lose data whose position is perfectly well determined.

WGS84 is now converted to the dimensionless Earth-fixed vector a fixed station uses -- the same representation as the MPC parallax constants -- and routed through `barycentricObservatoryRates`.

**The routing is the substance of the change, not a detail.** A roving observer stands on the rotating Earth, so its velocity is dominated by Earth rotation: 393 m/s at the latitude of the test site. The moving-observatory path adds a geocentric vector to Earth's barycentric state and falls back to Earth's velocity alone, so sending a roving observer down it would have placed the observatory correctly and given it the wrong velocity, silently, with aberration carrying the error into the fit. The ground-station path derives position and velocity from the same ITRF-to-J2000 rotation and gets both.

Because the position varies per observation, a roving observer has a per-epoch cache key, which until now meant "moving, do not apply the fixed-station transform". The dispatch is therefore on the frame rather than on the shape of the key.

### Validation

Against layup's own machinery rather than against typed constants:

| check | result |
| --- | --- |
| roving observer placed at station 703's recovered geodetic position | **0 m** in position, **2.6e-11 m/s** in velocity, vs the fixed station |
| geodetic conversion round-tripped over all 2,684 stations with parallax constants in `ObsCodes.json` | median **1.6 nm**, max **1.2 um** |
| end-to-end, an obs80 `V`/`v` record through to a barycentric state | **6,369 km** geocentric at 40.8 deg N (703 gives 6,374 km at 32.4 deg N) |
| WGS84 definitional known-answers (equator, pole, 1 km up) | exact |

Five new tests. Independent of #588/#589 -- disjoint files.

### One thing worth a reviewer's eye

The obs80 continuation line gives height in **metres**, and this assumes ADES `WGS84` uses the same unit. I could not find an ADES fixture in the repo to confirm it. If ADES means kilometres there, a roving observer ingested via ADES would sit a few hundred km off -- worth checking against the `observatorypos.pdf` attached to #282.


kjnapier (approved):
Looks good

